PR #126, opened 2014-10-21
Disabled gamemode votes during pregame.

Wymar:
Just to stop people fucking around with it even though they shouldn't.


Glloyd:
This hasn't been touched since October, and it really isn't an issue anymore, closing.

PR #195, opened 2015-02-10
Nighttime turfs & eventmap ease-of-use.

scrdest:
See commit name - nighttime variants of grass, both static and moving, added for quick mass-varedits to night during train events (not the map proper - but the turf is defined in case an alternate map, defaulting to nighttime is added). 
Image:  (in motion). 
To use them, just add _night to the current icon_state paths.

Moved event map loader to a separate file so that train file could be unticked without removing the eventloader and added an admin button to the Fun pannel to make loading a one-click operation instead of messing with manual proc activation, defined in the same file.


maps/EventMaps/mapevent.dm
```diff
@@ -0,0 +1,56 @@
+proc/LoadEventMap()
+
+	var/list/potentialEventMap = list()
+	world << "\red \b Searching for Event Map..."
+	var/list/Lines = file2list("maps/EventMaps/fileList.txt")
+	if(!Lines.len)	return
+	for (var/t in Lines)
```
Glloyd: mm, that was just a quick hack to have it working for the event. I gotta add it so admins can manually choose the specific map, unless you want to do that while you're in here.

scrdest: Working on it now.

scrdest:
I just made another thing for the train (albeit you may use it elsewhere), want me to roll it in this PR or a separate one?

Glloyd:
Depends, what is it?

scrdest:
Bikes! They are a child of cargo train tugs, although with a bit more code than a simple reskin to override some vars and procs to make them behave properly. Turns out just a reskin of cargo train tugs wouldn't work properly because they have an annoying overlay, so they are a proper object. And pretty sexy in my opinion, but I'll save the writeup for the actual commit.

The idea was to plop a couple of them to the part of the train map with moving grass and spawn in the antags as soldiers instead of assassins, so it becomes a more Nuke-ish variant of the event rather than the Traitor-ish one with the passengers fighting off boarders.

TGameCo:
Maybe a bike vending machine would be a good idea?
Also, could it be possible for a person to fold up a bike, pick it up, and carry it around until they unfold it and use it like an actual bike. Sort of like folding a rollerbed.

scrdest:
Bike as in a motorcycle. A train being stormed by a troop of armed goons on bicycles sounds rather amusing, but it'd be more fitting for a clown invasion.

scrdest:
Included bike sprites, so merge this one before the motorcycle PR.

maps/EventMaps/mapevent.dm
```diff
@@ -0,0 +1,59 @@
...
+		if(isfile(file))
+			maploader.load_map(file)
+
+			world << "\red \b Event Map loaded."
+
+		else
+			if (eventmap == null)
+				world << "\red \b Event Map loading aborted"
+			else
+				src << "\red Event Map couldn't be loaded properly. Yell at the coders."
```
Glloyd: Namely you and me. Honk.

scrdest:
DNM, I gotta remove the motorcycle from the .dmi again since I moved them to their own file, to be used for all urist vehicles.

Glloyd:
Did you fix that runtime? If so, :+1:

scrdest:
The runtime only pops up if you try to jump the gun and load the eventmap before the game can start up properly, so if it pops up it's your fault for rushing it. You should be able to load them even after roundstart without any problems, so no reason to be hasty.

Glloyd:
Cool. I just tested it, it's all good.

PR #197, opened 2015-02-15
Snow Turfs

scrdest:
Adds snow turfs based on (or rather, in the case of the static one, literally ripped from) the one from turfs/snow, now with 100% more animation, for Snowpiercing purposes. Moves at the same rate as the grass one, the gif doesn't because it was whipped up by some web program.


Glloyd:
I've already got the one that we used for the event on my local repo. Just haven't had a chance to push any of the stuff I've done lately because I've been too busy with midterms and whatnot.

scrdest:
Ah, I actually hadn't been around for Snowpiercer, so I didn't know if you had proper animated ones. And I guess you might as well just merge this one, faster for you. And I swear I didn't break everything. Yet. Too much.

Nienhaus:
It would break stuff for him because of the dmi.

Glloyd:
Eh, I can always revert my local copy, not a huge deal.

scrdest:
I realized that I accidentally made it progress in the opposite direction from the grass one; fixing this would be trivial, or you could just make Snowpiercer be oriented southwards instead of northwards.

PR #199, opened 2015-02-15
Vampire rebalance.

scrdest:
An attempt to make Vamp less slooooooooooooow.

Removes job protection unless needed for sanity (borg does not VANT TO ZUCK YOUR BLEH, and square-moustached British comedians shouldn't be vamped to foster counterplay).

Slightly increases scaling; 2 at 15 seems reasonable. I'm afraid of going overboard on it because even with just Rejuv VAMP STRONK, and vamp gang-ups even STRONKer.

Decouples total blood (i.e. the 'level up' blood for all intents and purposes) from usable blood; you gain the former half as fast, but the latter twice as fast as before (this is, from each 'suck' of blood), meaning gaining powers is a slower process, but you should burn through your powers more slowly, hopefully encouraging vamps to do shit instead of chugging monkeymen.

Axes cloneloss from suckage; it made no sense, was a pain in the ass to fix, and essentially turned vamp into a far more irritating husker than lings. And speaking of, causes pain instead; I'll nerf the amount, since I didn't realize at the time it was per-suck, and was in the hurry to fix other shit.

### DIFF
--- a/code/modules/urist/gamemodes/vampire/vampire.dm
+++ b/code/modules/urist/gamemodes/vampire/vampire.dm
@@ -303,9 +303,9 @@ You are weak to holy things and starlight. Don't go into space and avoid the Cha
 			break
 		if(H.stat < 2) //alive
 			blood = min(10, H.vessel.get_reagent_amount("blood"))// if they have less than 10 blood, give them the remnant else they get 10 blood
-			src.mind.vampire.bloodtotal += (blood/2)
-			src.mind.vampire.bloodusable += (blood*2)
-			H.traumatic_shock += 30 // vampire bites suck
+			src.mind.vampire.bloodtotal += (blood)
+			src.mind.vampire.bloodusable += (blood*1.2)
+			H.traumatic_shock += 15 // vampire bites suck, a long suckership will hurt the victim enough to knock them out
 		else
 			blood = min(5, H.vessel.get_reagent_amount("blood"))// The dead only give 5 bloods
 			src.mind.vampire.bloodtotal += blood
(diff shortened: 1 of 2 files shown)


TGameCo:
Suggestions: 

-Holy blood for chaplains. The only defense that chaplains have is their HHG and their Null Rod, and I believe that in the /vg/ version the holy, burning, caustic blood is a thing

-Vampires burn in chapels. Another vg thing I believe, and makes an area for people to rest in. It also makes it an actual challenge to get coffins, making it so that the vamp can't just easily steal it and become nearly invincible

scrdest:
@TGameCo literally 100% of Vamp code is lifted from /vg/, down to the placement to each copypasta'd block of code, with some bits like proper burning (like we had back on tgcode) commented out if they are unavailable. Unless it was added after I ported it, the code is word for word identical.

I agree that some of those make sense, but I'm reluctant to nerf anything for vamps now, as at the moment it's fancy Extended way too often. /vg/ has recently *AXED THE COSTS* for several evasive powers altogether to liven it up; I did it differently, but it hopefully will do the trick.

Glloyd:
So what are you doing with this, did you decide?

scrdest:
Yeah, I did some stuff on my local one already; I'm doing the no-costs things from vg except Screech and Mistform (which is reduced by half; that way, Shadowstep is useful).

scrdest:
OK, I'm happy with those changes for now. I'll see how it plays out in practice.

Glloyd:
Alright, so you're content with this? If so, I'll merge in the morning. Sorry for the delay on this, I've been really sick lately.

scrdest:
's cool, no rush. Yeah, I think this will work for now, I'll see how it plays out with these changes.

PR #200, opened 2015-02-16
Fixes a runtime caused by stationCollision.

scrdest:
Deletes the Hivebot that would be causing the runtime-causing code to execute. He was using an ancient var from the long-removed /code/module/critter, causing the various loaders/controllers to shit themselves, but because it was the map and not the .dm that would cause it, the runtime would go unnoticed.


Glloyd:
Hah, that runtime has been there forever. I kept meaning to remove that, but every time I try to save SC, DM crashes on me. I'm gonna merge this, but do you think you could set up a PR to fix those fucking chairs in the bottom right? They have their icon_state set wrong, so they show up a weird colour ingame.

Glloyd:
#200 is a critical runtime fix. Honk.

scrdest:
Sure, can do, just not sure which ones are those.

Glloyd:
The weirdly coloured ones near the gateway. If you load it up locally, you'll notice them right away. Ingame they're oranger than the normal chairs.

PR #201, opened 2015-02-17
[DNM][WIP] Mixed Mode

scrdest:
A semi-port of Mixed mode from /vg/. Currently not working as intended due to setup error (error setting up mode), but I'm putting the code up in case someone could find the root of the problem.


scrdest:
Augh, I am the worst with adding code. I'm gonna have to remove the nighttime turfs stray code from here.

Glloyd:
Let me know when you've got this working. I'll be able to help test again too.

scrdest:
Made pointless by https://github.com/Baystation12/Baystation12/pull/8360 which does the same thing more elegantly.

Glloyd:
Cool stuff. There's also going to be an update this weekend, so if you're able to finish bandages off, we can get it in in time.

PR #204, opened 2015-02-22
Motorcycles! Now with 100% spontaneous closing.

scrdest:
Motorcycles! Proper, oldtimey motorcycles intended for the train event, sprites (well, a sprite, because I only used one) courtesy of Nienhaus and modified to fit 32x32! A child of cargo train tugs, except modified heavier than a regular spriteswap as it turns out the engine on the tugs is an annoying little overlay.

They are:
* Sexy as fuck (pic slightly outdated, the mob sits a bit higher since I decided it looks better, but it's a matter of a couple pixels and could be easily varedited): ;
* Always oriented north, to use the illusion of motion from the moving grass sprite to always appear in motion;
* So is the passenger if in motion; you can turn around while static, but no riding backwards or sideways, shit's dangerous yo;
* Using a tactically placed overlay to create an illusion of a spessman sitting on the actual seat;
* Beefy health-wise (250), since adding them is a bit of a hassle and it's the riders you'll want to shoot at;
* Emagged by default to allow running people over like a proper outlaw biker;
* Intended to be used to create a more Nuke/Invasion-flavored Train round, with soldiers riding to the train and boarding it and crew fighting them off, rather than the Traitor-ish assassins.

Order now and get 100% less redundant PRs and up to 300% more commit-forcing-to-unfuck-branches!


scrdest:
It's literally the same code, just forced from a clean branch.

Glloyd:
:+1:

Nienhaus:
So when can you make it so it takes plasma as fuel :D

scrdest:
Ugh, I'm afraid it's non-trivial; I'd have to look into how PACMANs (PACMEN?) work; normal vehicles use up electricity from the battery, it would have to increase the charge by each sheet of plasma.

PR #207, opened 2015-02-23
[DNM] Rags: Ghetto Bandages

scrdest:
That's right, something even shittier at keeping people alive than bandages proper, but still better than bleeding to death of a deep papercut. The idea, of course, is after yesterday's train round.

Rags use the damp rag sprite for now and are made by ripping apart suit or jumpsuit layer clothes for two or one rag, respectively, with checks for /space/ /wizrobe/ and /armor/ subtypes of suits to prevent breaking those; it could probably be done with something more elegant like vars or flags, but it would also mean shitting my code all over the branch.

Tearing clothes apart is not immediate; it takes a short while, but you may move during that, and requires the clothing to be held and NOT worn, so you cannot accidentally destroy all your clothes and you have a chance of disarming someone who stole and tries to destroy your clothes.

Unlike bandages, rags can be used multiple times, but they only stop the bleeding, only at a 20% chance (lower than self-applied splints, which are a pain already), slightly raise the wound's germ level and need washing. After being applied five times (so, statistically after stopping one wound) they become soaked with blood, requiring to be washed in a sink to be usable again. The icon stays the same for now, but I'll get some spritework for that later.

The PR is working as of right now, but it's working *TOO* well; I need an adult here, because the bandaging system is so arcane that it involves changing in pentagrams written in a newborn's blood to compile it - somehow, applying the rag bandages the organ even when I didn't tell it to, via some kind of weird cross-checking, so I hope @Zuhayr (because he checks up on us) or someone else who groks it can help me figure it out.

### DIFF
--- a/code/modules/urist/items/uristmed.dm
+++ b/code/modules/urist/items/uristmed.dm
@@ -134,3 +134,41 @@ Space for all Urist-done, non-pill medical items. Please keep it tidy, as usual.
 					return
 			else
 				user << "<span class='notice'>The [affecting.display_name] is cut open, you'll need more than a [src]!</span>"
+
+
+/obj/item/clothing/suit/verb/rip(mob/user as mob)
+	set name = "Rip Apart Suit"
+	set category = "Object"
+
+	if(!usr.canmove || usr.stat || usr.restrained())
+		return 0
+	if(istype(src, /obj/item/clothing/suit/space) || istype(src, /obj/item/clothing/suit/armor) || istype(src, /obj/item/clothing/suit/wizrobe)) //add other protected instances as needed
+		return 0
+	else
+		user.visible_message( 	"<span class='notice'> [user] starts ripping [src] apart into rags!</span>", \
+											"<span class='notice'> You start ripping [src] apart into rags!</span>" )
+		spawn(100)
+			user << "<span class='notice'>You have ripped [src] into pieces, yielding two rags.</span>"
+
+			new /obj/item/weapon/clothesrag(get_turf(src))
+			new /obj/item/weapon/clothesrag(get_turf(src))
+
+			del(src)
+
+/obj/item/clothing/under/verb/rip(mob/user as mob)
+	set name = "Rip Apart Jumpsuit"
+	set category = "Object"
+
+	if(!usr.canmove || usr.stat || usr.restrained())
+		return 0
+/*	if(istype(src, /obj/item/clothing/under/BADSTUFFGOESHERE //I couldn't think of any item types here that shouldn't work, but leaving it here in case it's needed later
+		return 0*/
+	else
+		user.visible_message( 	"<span class='notice'> [user] starts ripping [src] apart into a rag!</span>", \
+											"<span class='notice'> You start ripping [src] apart into a rag!</span>" )
+		spawn(100)
+			user << "<span class='notice'>You have ripped [src] into pieces, yielding a rag.</span>"
+
+			new /obj/item/weapon/clothesrag(get_turf(src))
+
+			del(src)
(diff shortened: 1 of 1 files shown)


Glloyd:
I'm on campus so I don't have time to look at this, but for the love of god, use the existing ripping sound. The sound system's shitty enough as-is, no need to play a sound twice at one.

scrdest:
It's not as one, although I will probably tweak the delay with some other
numericals once I get the bandaging to trigger properly; I simply liked the
sound and triggered it twice with a delay to simulate more ripping than
just once.

2015-02-23 23:33 GMT+01:00 Graham Lloyd <notifications@github.com>:

code/modules/urist/items/uristmed.dm
```diff
@@ -39,3 +39,174 @@ Space for all Urist-done, non-pill medical items. Please keep it tidy, as usual.
...
+				return 1
+
+		if(affecting.status & ORGAN_ROBOT)
+			user << "<span class='warning'> This isn't useful at all on a robotic limb..</span>"
+			return 1
+
+		H.UpdateDamageIcon()
+
+		if(affecting.open == 0)
+			if(!affecting.bandage())
```
comma: Wounds are getting bandaged at this point, so prob check down the line does nothing.

code/modules/urist/items/uristmed.dm
```diff
@@ -39,3 +39,174 @@ Space for all Urist-done, non-pill medical items. Please keep it tidy, as usual.
...
+
+	if(!usr.canmove || usr.stat || usr.restrained())
+		return 0
+	if(!istype(H))
+		return 0
+	if(H.wear_suit == src)
+		user << "<span class='warning'>Deciding not to channel your inner barbarian, you think it's a good idea to take off your [src] first.</span>"
+		return
+
+	if(istype(src, /obj/item/clothing/suit/space) || istype(src, /obj/item/clothing/suit/armor) || istype(src, /obj/item/clothing/suit/wizrobe)) //add other protected instances as needed
```
comma: Should probably have suit/storage (like labcoats) or some handling for those so items stored in coat pockets do not get deleted

code/modules/urist/items/uristmed.dm
```diff
@@ -39,3 +39,174 @@ Space for all Urist-done, non-pill medical items. Please keep it tidy, as usual.
...
+		spawn(100)
+			user << "<span class='notice'>You have ripped [src] into pieces, yielding a rag.</span>"
+
+			playsound(src.loc, 'sound/weapons/cablecuff.ogg', 100, 1)
+			spawn(5)
+				playsound(src.loc, 'sound/weapons/cablecuff.ogg', 100, 1)
+
+			new /obj/item/weapon/clothesrag(get_turf(src))
+
+			del(src)
```
comma: Needs to check hastie to see if there's an accessory to leave behind first, so medals are not ripped in pieces, and more importantly, holsters with guns in them.

code/modules/urist/items/uristmed.dm
Glloyd: Or if you really want suit/storage to be rippable, you could always add a check for them to drop their contents before calling del() on the the suit.

scrdest: I do, yeah, since coats are now a subtype of storage and the primary application for those were intended to be Train rounds. I'll look into this, but since I'm back at the university, I'll have less free time to spare.

Glloyd:
@scrdest, what's the word on this?

scrdest:
Didn't have time to work on it, I've had two 8 AM to 8 PM days at the uni this week. But now that I know what was the issue, I'll work on it when I can.

scrdest:
I think I need to redo this in a more sane manner, closing for now.

PR #211, opened 2015-03-03
Adds icon_overrides to coats.

scrdest:
Also updates body_parts_covered, since I forgot that lower_torso is a thing when adding them initially and it just makes sense.


Glloyd:
Perfect, thanks :+1:

PR #215, opened 2015-04-03
Disables job protection.

scrdest:
Job protection, particularly that of Security, creates an atmosphere where Sec becomes a Good Guys Club that are morally justified in clubbing and perma-ing just about anyone, because there is very little chance any of them is an antag.

### DIFF
--- a/config/config.txt
+++ b/config/config.txt
@@ -112,7 +112,7 @@ TRAITOR_SCALING
 
 
 ## If security is prohibited from being most antagonists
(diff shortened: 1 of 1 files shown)


Glloyd:
We'll toss up a poll for it. There was a poll for it the first time, so we should poll to remove it.

scrdest:
Seems that antag code merrily ignores jobs anyway, it's kinda pointless. Honk.

PR #226, opened 2015-05-10
Aooc changes/fixes

TGameCo:
Adds the icon prefixes to aooc, and was supposed to make it visible to only antags. The only thing is, there is no real way to do so, as nobody put a central antagonist subset for the mind datum, nor gave a general antag setup, nor gave anything that I could do without cannibalizing most of the game's code. Ah well. Also did some color alterations and added a few small comments


code/modules/urist/verbs/antagooc.dm
```diff
@@ -1,10 +1,11 @@
-var/global/normal_aooc_colour = "#FF0000"
+var/global/normal_aooc_color = "#FF3333" //Screw british speling of color. COLOR is correct. Like math. MATH is correct. MATHS is stoopid
 
 /client/verb/aooc(msg as text)
 	set name = "AOOC"
 	set category = "OOC"
+	//set hidden = 1
```
scrdest: Is there any point to keeping this as a comment?

code/modules/urist/verbs/antagooc.dm
```diff
@@ -1,10 +1,11 @@
-var/global/normal_aooc_colour = "#FF0000"
+var/global/normal_aooc_color = "#FF3333" //Screw british speling of color. COLOR is correct. Like math. MATH is correct. MATHS is stoopid
```
scrdest: Probably better handled by a #DEFINE instead.

scrdest:
Also, BS12 is adding something also called AOOC, but one that works differently (admin only - I don't even), so you might want to look into it so there's no confuzzlement.

code/modules/urist/verbs/antagooc.dm
TGameCo: I may be able to find something later, so I kept it. For the moment though, I am stumped

TGameCo:
Why don't they just use msay or asay? Bay is wierd

code/modules/urist/verbs/antagooc.dm
scrdest: I can try to help if you tell me what you're looking for.

scrdest:
Because both are limited to one person at a time, apparently, so it's intended as a mass-message to antags. Some people on their forums even thought it would work like our AOOC.

TGameCo:
Hah, wierd. Where are they talking about it? I'd like to see what they're trying to do

scrdest:
For some reason, I cannot find the commit; it's already included in the latest dev freeze.

TGameCo:
[Found It](https://github.com/volas/Baystation12/commit/c7573a0d11a37dc2dfd14699a30a8bb687fe2eb8). It was so nice of them to tell us about this. I feel so happy that others see my ideas as so beneficial, they made their own version themselves. 

I might as well scrap my code now, next time we pull it'll break otherwise. ¯\\_(ツ)_/¯

scrdest:
No, like I said, theirs is kinda pretty goddamn pointless most of the time, as our AOOC already includes admin->antags messages on top of what theirs has. Honestly, they should have pulled ours, they could benefit from it even more than we do.

TGameCo:
So when the merge comes, we'll keep our version?

scrdest:
Well, I'm not the one in charge of that, but I presume Glloyd wouldn't randomly ditch ours just because the name's the same.

code/modules/urist/verbs/antagooc.dm
scrdest: Oh, I see what you were trying to do here. Hidden won't cut it - all it'd do is hide it from the panel. If you want to remove the verb itself from non-antags/admins, verbs -= /client/verb/aooc should work.

Glloyd:
im deleting the bay one when we get it because it's fucking pointless

Glloyd:
like a lot of things from bay these days

TGameCo:
Why don't we just stop pulling directly from bay? Each merge seems to break more things than it adds, but I'd guess you'd know more about it glloyd

PR #234, opened 2015-06-15
ANTAG Base

scrdest:
On the CentComm Z-level. Minor changes were made to the toilet wall to make it connect to the rest. As a side note, those rwalls are transparent - it's supposed to be a rooftop/penthouse, and that was the closets approximation I could get with SS13 perspective and sprites at hand.

Totally, absolutely not inspired in any shape or form by EXALT HQ, why do you ask?

In the future, if ANTAG becomes an actual official part of the gamemode, there's some Z-level elements that could be used to make an assault from the bottom floor, but the roof and stairway access doors are fakes, as are all the windows currently.


Glloyd:
I'm wondering if this wouldn't be better under missions2.dmm. I'm torn though, because this would be cool to have in an actual round.

Glloyd:
Also, you broke travis.

Glloyd:
Honk, you used step x and y instead of pixel x and y :speak_no_evil:

scrdest:
What? I didn't touch it.

scrdest:
I've been thinking where to put it too, I left it on main map as S-COM is there too, and it could be used in lieu of CentComm for a place to spawn assorted admin-antags or something.

Glloyd:
Yeah. I'm fine with keeping it on z level 2 until I get ANTAG into the actual game, at which point I think I'd move it to the S-COM missions map and expand it into a full base.

Glloyd:
:+1:

PR #235, opened 2015-06-16
Dynamic difficulty and lowpop spawns scaling for SCOM.

scrdest:
Two complementary systems:

**Dynamic Difficulty**: can be disabled by config or by a Fun rights admin (sadistic admins may disable it and set mission difficulty to Harder); uses the current number of living players to determine the next mission's difficulty, so a round with 16 roundstart operatives is not stuck in easymodo when 16 more join mid-round, and conversely so that five-man squad of survivors has a fighting chance and is not forced to fight through a gorillion aliens because they started out at 40 people.

Practically, it moves the difficulty checks for pop, either roundstart if disabled, or current if enabled, to a proc that is triggered on pre-setup, but repeated on mission announcement if the option is on.

**Lowpop spawns**: a quick and dirty way of downscaling the difficulty level - if the pop is below the amount the lowest difficulty spawns were designed for, it uses RNGesus' mercy to cancel out some of the spawns proportionately to the number of players below the Easy threshold (ex. 5 players get 1/3 of the alien spawn chance, 10 get 2/3, 8 get around 50%).

*Disclaimer*: RNGesus's mercy is limited - the spawn chance will never fall below 33% for each spawn point, so 1 player gets at least as much spawns, statistically, as a 5-man team. Also, if you angered the RNGesus by blaspheming against His name, you may find yourself fighting more or less aliens than direct scaling would have it, because statistics. Above 15, spawning is not affected by this system.

*Disclaimer 2*: RNGesus's mercy is universal and does not discriminate between virtuous defenders of mankind and evil collaborators or other scum of the OneClickAntag currently, so keep that in mind - a five man ANTAG team is not only a threat to SCOM players, but also increases the mission difficulty just by existing. If you're deploying them, you're probably delaying a mission anyway, but still.


code/modules/urist/gamemodes/scom/scom.dm
```diff
@@ -9,20 +9,25 @@ var/global/sploded = 0
...
 
+var/global/SCOMplayercount = 0 //count is player number at the moment, C is at roundstart
+var/global/SCOMplayerC = 0 //ugly rename, but AFAIK playerC is a local var of different gamemodes. Honestly, this should be a helper var.
+
+
 /datum/game_mode/scom //the joke is that 90% of this stuff is handled by other procs
 	name = "scom"
 	config_tag = "scom"
-	required_players = 16
+	required_players = 1 //lowpop mode ahoy
```
Glloyd: IMO it shouldn't be lower than 10, 8 at the very lowest.

code/modules/urist/gamemodes/scom/scom.dm
```diff
@@ -39,23 +44,33 @@ var/list/scomspawn3 = list()
...
 	config.allow_random_events = 0 //nooooope
 
-	var/playerC = 0
 	for(var/mob/new_player/player in player_list)
 		if((player.client)&&(player.ready))
-			playerC++
-		world << "\red \b [playerC] players counted."
-	if(playerC >= 15)
+			SCOMplayerC++
+	world << "\red \b [SCOMplayerC] players counted."
```
Glloyd: can you comment this out for me?

code/modules/urist/gamemodes/scom/scom.dm
```diff
@@ -9,20 +9,25 @@ var/global/sploded = 0
 
 var/global/onmission = 0
 
+var/global/scom_lowpop_scale = 0
+
 var/list/scomspawn1 = list()
 var/list/scomspawn2 = list()
 var/list/scomspawn3 = list()
 
+var/global/SCOMplayercount = 0 //count is player number at the moment, C is at roundstart
```
Glloyd: Why does this need to be a global? It's only used in process()

Glloyd:
Why did you touch retiredmissions.dmm?

code/modules/urist/gamemodes/scom/scom.dm
scrdest: update_dyndifficulty() too, but it's a scom proc, possibly oversight

Glloyd:
Steam is shit: Because it wouldn't work well with that few people. The maps are too big, the hq is too big, too many enemies even with low spawning and noone to heal or do science. With one person, the game will end if they log out, automatic alien victory. Or die for that matter.

scrdest:
I made the ANTAG base there, actually, then copied it over to the main. Essentially a backup in case the main map had to be reverted

Sure, but that makes it a very quick mode for low pops. IMO, that kinda is the advantage of the mode - it's PvE, so you can toss in those measly 3 players who could barely if at all play a regular round and have 'em have something to do for a while. With dynamic scaling, as soon as the player count gets reasonable, they can still have a challenge, and if they fail - well, they can very quickly get a station round started.

code/modules/urist/gamemodes/scom/scom.dm
scrdest: done

Glloyd:
:+1:

PR #240, opened 2015-06-24
Further ANTAG tweaking.

scrdest:
Now with 300% less rouge.

Literally just changes the base of the outfit into the waistcoated detective suit to allow wearing waistcoat + armor and/or have webbing be drawn over the waistcoat. Might add old variant into the armory.


Glloyd:
:+1:

PR #241, opened 2015-06-25
Makes Cold Resistance protect from the snow.

scrdest:
Because right now you get freezer burn even if you don't feel uncomfortable in 0K water.

Also, MUH SPAN CLASSES. Only in the snow, CBA to do the rest right now.


Glloyd:
:+1:

PR #246, opened 2015-07-06
Van Dyke fixing.

scrdest:
A port of the /tg/ PR, we had it back on tgcode so its kinda old, that makes the Van Dyke an actual Van Dyke, and the... thing that currently holds that title is now named a Fu Manchu. The sprite is a port of the /tg/ sprite as well. Also included is an edit with a downpointed instead of straight 'stache and a thinner, slightly curled goatee (only visible sideways and CBA to make all side examples).

vandycktgs vandyckalt

Left is the tg one, right is the alt.


Glloyd:
Put it in whatever .dmi has the urist hair and change it in the file too.

code/modules/mob/new_player/sprite_accessories.dm
```diff
@@ -510,6 +510,10 @@
 		name = "Van Dyke Mustache"
```
Glloyd: you can remove this

Glloyd: git didn't update, nvm

PR #248, opened 2015-07-12
Removes angle(elevation) offset from telesci

Flip-Sandwich:
Simply put, this makes telescience actually usable and adds consistency with every other codebases's telescience.

Only the angle(elevation) offset is removed, both the bearing and power offsets still remain. You will still have to calibrate telescience, it's just not practically impossible to do so anymore.


scrdest:
Sounds good. Not much code to break, anyway. :+1:

Glloyd:
Remind me to add a refactor lable.

Issue #247, opened 2015-07-09
Kitchen code, Y U DO DIS?


scrdest:
Custom foods need a serious cleanup. Off the top of my head:

- [x] aside from sammiches, which, possibly due to being baycode feature, work pretty fine (with a minor exception), custom food items completely ignore filling color and instead go with eye-bleedingly ugly RGBY colors. I have a suspicion which line of code causes it.

- [ ] sammiches, as well as other custom foods, don't transfer their contents to other custom foods they are added to, instead the whole object creates nested custom foods. While in most cases hilarious, you can easily break reality by adding a max size sandwich onto a slice of bread, collapsing all the contents of the old sandwich into Quantum Deliciousness Storage. Feedback suggests keeping it.

- [ ] grill removes ALL overlays, glitching the graphics for overlay-based foods.

- [ ] oven-based machinery can magically teleport food to and from the user, probably due to the sleep() preceding setting loc without a sanity check for the mob being adjacent.

scrdest:
I figured out why ice cream showed an empty cone icon. @Glloyd you forgot to fix the icon path to point to urist/kitchen.dmi rather that base code kitchen.dmi.

Glloyd:
Could you do that for me?

On Sat, Jul 11, 2015 at 7:44 AM, scrdest <notifications@github.com> wrote:

> I figured out why ice cream showed an empty cone icon. @Glloyd
> <https://github.com/Glloyd> you forgot to fix the icon path to point to
> urist/kitchen.dmi rather that base code kitchen.dmi.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/UristMcStation/UristMcStation/issues/247#issuecomment-120606935>
> .
>



-- 
Graham Lloyd

scrdest:
Did already. I will have a PR by tomorrow with some changes, just need to test the last changes.


Sent from Samsung Mobile

-------- Original message --------

scrdest:
Should be fixed by #249

PR #249, opened 2015-07-13
Custom food and drinks fixes/rewrites/sanity

scrdest:
Addresses some stuff from #247. Fixes the icon path for ice cream, resizes the gummy bears and jelly beans to no longer be giant - at least for the latter, no tears were shed over quality loss, including renaming them to be non-giant.

Rewrites handling of filling colors. I initially intended to just fix it not using filling_colors at all [which was a stupid bug caused by a check being if(!condition) rather than if(!(condition))] but I was unsatisfied with how it would just plain overwrite the previous filling color with a new ingredient's, so it's now blending a new color with an old one using an average blend operation (with some special features, check the code for specifics).

Since this wasn't an icon operation, and I am very, very efficient and smart like that (-_-), instead of just going back and converting filling_color values to RGBs I wrote a bunch of helpers to deal with handling hex-to-RGB colors and mixing them. I know helpers/game already did that to an extent, but this is simpler to use. Also, see first sentence in this paragraph. I also moved that and Vamp's is_on_ears helper to their own file.

Finally, in addition to the color handling code same as for food, reworked how custom alcohols are created. In addition to a much smaller flat amount, a new proc, ferment(), simulates the still insta-fermenting nutriment and sugar in the food added into alcohol.

The alcohol is also no longer hardcoded in a snowflake New() for EACH AND EVERY SUBTYPE it's controlled by a variable given to the subtype. It actually doesn't have to be an alcohol; you could make it water for custom flavored water or any other existing reagent. It will default to the veeeery unpleasant Hooch (realism!), so you better specify it.


Glloyd:
:+1: I can see those helpers coming in handy.

PR #250, opened 2015-07-15
Map Fixes

JacobLee1:
+ Fixes most of the pumps in atmos being on at roundstart. This should reduce the power drain significantly.
+ Fixes the AI's SMES being drained by power sinks. The AI core was connected to the main powernet; not sure if I oversaw this when first mapping it or not, but only lately I heard of the bug.
+ Fixes the AI upload's turrets being uncontrollable. The same bug probably exists in the AI core, which I also fixed.


Glloyd:
My hero returns!

Nienhaus:
>Still hasn't updated the IAA office for me. RIP

Glloyd:
:+1:

PR #251, opened 2015-07-15
Another fix

JacobLee1:
Fixes the incredibly important issue of the homogeneous folders in IA. There's now a red, blue, yellow, and white folder instead of four blue ones. Gllerd is a ninja and pulled my PR before I could get this pushed.


Nienhaus:
Gllerd= Glloyd+nerd?

Nienhaus:
Oh, also :+1: x :100:

Glloyd:
:+1:

scrdest:
Not the hero we deserve, but the hero we need right now, right here.

PR #254, opened 2015-07-19
Left 13 Spess

scrdest:
Adds a sub-variant of the regenerating zombies - plague zombies, able to zombify corpses of humanoids and SCOM civvies (not other 'human' simple_animals, simply because I didn't want to add a ton of istypes) and associated procs to support said transformations. 

They can be called by non-zombies for not-Chryssalids and whatnot, and have input-able mobpath but will only accept zombie subtypes for consistency, defaulting to plague type for invalid paths.

The infection is spread by a plague zombie attacking a transformable corpse and eating his brain. Yeah, it doesn't check whether it has one, spess magic. This spawns a zombie mob specified in the Zombify/SAZombify's mobpath argument. If the victim is a humanoid, the zombie retains the victim's appearance and name (with the prefix 'zombified') for extra spoopy.

If a player-controlled human is infected and doesn't ghost, the zombie is controlled by the player and gets a verb to infect corpses, since the AI zombies' behavior is dependent on their AI routine. Say code is overrided so that whatever the player tries to say comes out as 'BRAAAAINS!'.

Furthermore, the plague zombie is created *dead*, including the ones spawned with infectees. As it's a regen zombie type, it will resurrect in 1.5 minutes and while it's dead, it's assigned a temporary name and description of a 'corpse' to camouflage what they are. It's one part support for admin shenanigans, one part balancing so the infected don't immediately join the fray.

**THIS IS A 'GUYS, END THE ROUND ALREADY'/'EVENT TIME!'-TIER BADMINNERY. THIS IS AN UNKILLABLE HOSTILE MOB WHICH CAN RAPIDLY SPREAD. USE WITH CAUTION OR FACE PLAYER WRATH. SERIOUSLY.**


Glloyd:
Ahhh, this makes me really want to give some simple mobs the human damage system so you could delimb these guys. Is this a precursor to zombie mode?

scrdest:
Not in intention. More like the perfect example of feature creep. I suggest
you steal the death code for civvies, less spritework and resource filesize.

scrdest:
Speaking of losing limbs - it's a semi-bug I didn't mention, but if you zombify someone without both legs, the zombie will kinda float on its chest. I've been considering rejuving the target to prevent that, but ended up electing not to for now.

Glloyd:
I'm back now, so I'll take a closer look at this tomorrow and likely merge it then.

Also, code like this makes zombie mode not seem so far off :stuck_out_tongue_winking_eye:

scrdest:
It is. I thought about making it work similar to tg's playerblob - a picked antag starts 'infected' but not turned until a timer runs out - but I'm not touching gamemodes until the Bay does their damn freeze pull at last.

Glloyd:
We should definitely just pull from dev freeze at this point

Glloyd:
Should have some checks for if the mob has no limbs, but that can be done at a later date (or for zombie mode itself) :+1:

scrdest:
Considering the attack is a bite, a headless mob shouldn't work (might just add a brain check), arms don't matter and a legless one can be hooked to a var that will turn the sprite by default and bypass turning them on death.

PR #290, opened 2016-01-27
Goofchem Framework

zty8212000:
Adds the basic framework for temperatures and chemical  heaters.

### DIFF
--- a/code/game/machinery/chem_heater.dm
+++ b/code/game/machinery/chem_heater.dm
@@ -1,0 +1,113 @@
+/obj/machinery/chem_heater
+	name = "chemical heater"
+	density = 1
+	anchored = 1
+	icon = 'icons/obj/chemical.dmi'
+	icon_state = "mixer0b" //GET IN THE ICON
+	use_power = 1
+	idle_power_usage = 300
+	var/obj/item/weapon/reagent_containers/beaker = null
+	var/desired_temp = 300
+	var/heater_coefficient = 0.10
+	var/on = 0
+
+/obj/machinery/chem_heater/process()
+	..()
+	if(stat & NOPOWER)
+		return
+	if(on)
+		if(beaker)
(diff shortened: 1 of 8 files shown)


code/modules/mob/mob_grab.dm
```diff
@@ -37,7 +37,7 @@
 
 
 //Used by throw code to hand over the mob, instead of throwing the grab. The grab is then deleted by the throw code.
-/obj/item/weapon/grab/proc/throw()
+/obj/item/weapon/grab/proc/throw_target()
```
scrdest: ...why is this even affected?

zty8212000: As far as I can tell? BYOND decided throw() was now a reserved proc name and refused to allow it to compile without altering it. I've also changed all calls to the procedure, so everything should fine.

scrdest: Oooh. Yeah, it is, if you're using the updated BYOND, I forgot about it completely.

Glloyd:
this needs to be updated for the update.

code/game/machinery/chem_heater.dm
```diff
@@ -0,0 +1,113 @@
...
+			user << "<span class='notice'>You add the beaker to the machine.</span>"
+			icon_state = "mixer1b"
+
+/obj/machinery/chem_heater/attack_hand(mob/user)
+	if (!user)
+		return
+	interact(user)
+
+//Artifact from /tg/
+/*/obj/machinery/chem_heater/ui_act(action, params)
```
neersighted: Hah that's tgui, tgui's new UI system. You could just make this `Topic()` instead...

Glloyd:
You need to update this before any more progress can be made.

Glloyd:
Author is no longer on github

PR #317, opened 2016-02-11
Replaces arithrazine in anti-rads with hyronalin.

scrdest:
Arithrazine has extremely dangerous side-effects now, and a self-medicating engineer might find themselves having their... everything infected. At the same time, its danger is not obvious and with the engine changes the rad damage are less dangerous, so using last-resort anti-rads in autoinjectors feels rather extreme.


Glloyd:
:+1: get this in before the update.

PR #323, opened 2016-02-14
Switches Mentors back to Mods

scrdest:
Because it seems to be a requested change.


scrdest:
Uh. It appears BS12 removed this config option altogether and is just going with whatever rank is assigned in the config files.

PR #324, opened 2016-02-14
Lowers respawn time to 20 minutes

scrdest:
By popular demand.

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -371,7 +371,7 @@
 		var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10,1)
 		usr << "You have been dead for[pluralcheck] [deathtimeseconds] seconds."
 
-		if (deathtime < 18000)
+		if (deathtime < 12000)
 			usr << "You must wait 30 minutes to respawn!"
 			return
 		else
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -372,7 +372,7 @@
 		usr << "You have been dead for[pluralcheck] [deathtimeseconds] seconds."
 
 		if (deathtime < 12000)
-			usr << "You must wait 30 minutes to respawn!"
+			usr << "You must wait 20 minutes to respawn!"
 			return
 		else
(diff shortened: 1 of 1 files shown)


code/modules/mob/mob.dm
```diff
@@ -371,7 +371,7 @@
 		var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10,1)
 		usr << "You have been dead for[pluralcheck] [deathtimeseconds] seconds."
 
-		if (deathtime < 18000)
+		if (deathtime < 12000)
 			usr << "You must wait 30 minutes to respawn!"
```
liction15: Might need to change the line to say "You must wait 20 minutes to respawn!"

Zuhayr: Could also put the time in a var and do ````usr << "You must wait [(spawntime / 100) / 60] minute\s to respawn!"````.

scrdest: I didn't even notice that. I'll do a simple edit to 20 minutes, simply because I want to make as few edits as possible for merging's sake.

code/modules/mob/mob.dm
```diff
@@ -371,8 +371,8 @@
 		var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10,1)
 		usr << "You have been dead for[pluralcheck] [deathtimeseconds] seconds."
 
-		if (deathtime < 18000)
-			usr << "You must wait 30 minutes to respawn!"
+		if (deathtime < 12000)
```
neersighted: You could make this a config option

scrdest: Same as above, this is the least invasive edit I could possibly make, worst case Github reverts it right back during a merge.

Considering the state of Baycode, which we're currently tied to, Urist is currently working out the policy as to what to do, and one thing's for sure that after handling a good part of the code side of merging two Bay dev cycles at once I do *not* want to introduce another bulletpoint on the TODO list for the next one if possible.

I could macro it on top of this file, but it's literally only used here anyway.

Glloyd:
Yeah, this is fine for now until we decide definitively on the code future on Urist.

PR #327, opened 2016-02-15
Chameleon gun fix

### LINKED ISSUE #326: Chameleon Gun does not use mags
The chameleon gun does not seem to use magazines. It still works but you have to empty the mag and load the bullets into the gun one by one which is weird, slow and makes it painfully awful.

liction15:
Fixes #326 

Please check the code to ensure this would actually work. I had to dig around and found the gun under chameleon clothing, not sure if that would affect anything. Mostly copied over what it seemed to be missing and edited it to what it should be.

### DIFF
--- a/code/modules/clothing/under/chameleon.dm
+++ b/code/modules/clothing/under/chameleon.dm
@@ -443,7 +443,7 @@
 	ammo_type = "/obj/item/ammo_casing/chameleon"
 	w_class = 3.0
 	caliber = ".45"
-	load_method = MAGAZINE
+	load_method = 4 //Unable to set to MAGAZINE because the file's read before the macro def
 	origin_tech = "combat=2;materials=2;syndicate=8"
 	matter = list()
 	var/list/gun_choices = list()
--- a/code/modules/clothing/under/chameleon.dm
+++ b/code/modules/clothing/under/chameleon.dm
@@ -439,11 +439,12 @@
 	name = "desert eagle"
 	desc = "A fake Desert Eagle with a dial on the side to change the gun's disguise."
 	icon_state = "deagle"
-	w_class = 3.0
-	max_shells = 7
-	caliber = ".45"
-	origin_tech = "combat=2;materials=2;syndicate=8"
+	magazine_type = /obj/item/ammo_magazine/chameleon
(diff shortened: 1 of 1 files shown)


liction15:
okay it failed but Im learning it seems MAGAZINE needs to be defined

scrdest:
Something's wrong, it shouldn't read MAGAZINE as a variable, it's essentially a stage direction for the compiler to replace the word MAGAZINE with a static number - in this case, 4.

I'd hazard a guess it's a silly problem with the compiler not reading the macro definition because the clothing module is read before the projectile.dm file with the definition, though I might be wrong.

Also, MAGAZINE load_type guns don't seem to need ammo_type defined, the magazine's vars cover that.

Zuhayr:
````MAGAZINE```` is defined after this file is checked by the compiler, yep.

liction15:
the only thing I could think of to fix this is to move the chameleon gun bit to where all the other pistols are.

scrdest:
I'd edit that to simply load_mode = 4 and add a comment explaining you couldn't make it MAGAZINE because the file's read before the macro def.

liction15:
yay it seems that it compiled correctly, I is real coder nao?

scrdest:
Did you test it? 

If you didn't, you can edit the admin list in config to give yourself full priviledges, set up a local server with the Dream Daemon and spawn the gun and mag. Just don't commit the change for all that's holy. 

It looks good at a glance, but I know from experience porting guns to the new guncode can be a bit tricky to get right.

liction15:
Alternatively we could port over the existing file that's on bay which seems to turn the gun into a energy weapon, although we will probably need to change the chameleon kit to not include the mags. In any case I don't really know how to set up a local server to test this myself right now, I can probably figure out that later though.

scrdest:
It's pretty straightforward. Compile in DreamMaker (Build -> Compile), open Dream Daemon, enter the file path of the .dmb file, hit the green button, wait until it turns yellow (it may briefly hang on you), click it. Boom, you're in game on a local server.

liction15:
Just finished testing it. I can confirm that the gun now comes with a magazine loaded and also can be loaded and unloaded like other guns.

scrdest:
Alright then!

PR #329, opened 2016-02-15
Minor Bar/ANTAG remaps

scrdest:
Adds black curtains to the Bar windows inspired by some mid-round redecorating, should be openable from the inside (spawned on a tile inside the bar and y-offset to the window tile), start open.

Minor changes to the ANTAG base (shuttle chair -> red chairs, some magazines, a couple of walls replaced with the 'pillar'-type wall icon).

Cleaned using the map merge tools, hopefully it didn't make the map crash and burn.


Glloyd:
Looks good :+1:

PR #337, opened 2016-02-18
Vamp Fixing

### LINKED ISSUE #319: Vampire Turn power not working.
Followup to #312 - the fix (adding the call to the old snowflake proc dealing with adding the actual mind/vampire, powers and icon under add/create_antag rather than update_antag_mob (why I used it I don't even) managed to address the issues with gamemode spawning the antag on roundstart, but the Turn power still takes the 500 blood but doesn't properly add the target to antags nor vamps them properly via add_antag call in the handle_vampirize proc.

Since the antag refactor is a mess of procs with almost identical purpose and barely documented (guess what's the difference between add_antag, add_antag_mind and create_antag, for example), I don't know if I'll be able to fix this on my own before I get some kind of minor aneurysm from figuring out the system, so I'd be obliged if you could take a peek at the add_antag call @Glloyd

scrdest:
Screw you, antagonist system procs. Screw. You. Attempts to fix #319 

Furthermore, fix Glare not working on mobs *with* eyes (I screwed up the eyecheck, more == more protection) and remove_antagonist not properly cleaning the var from mind.

### DIFF
--- a/code/modules/urist/antagonist/vampire.dm
+++ b/code/modules/urist/antagonist/vampire.dm
@@ -20,7 +20,7 @@ var/datum/antagonist/vampire/vamps
 ///datum/antagonist/vampire/get_special_objective_text(var/datum/mind/player)
 //	return //"<br><b>Real Name:</b> [player.real_name].
 
-/datum/antagonist/vampire/create_antagonist(var/datum/mind/player)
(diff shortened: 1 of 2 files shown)


scrdest:
Rouge merge as a bugfix.

PR #340, opened 2016-02-19
Paranoia Mode 1: The Randomgreekname Conspiracy

scrdest:
Paranoia is a team-based mode similar to /tg/ Gang and Revolution and somewhat inspired by Mafia-style parlor games. 

Unlike Gang or Rev, it's not intended to be a mode about direct confrontation, focusing on subterfuge instead. Conspiracy agents do *not* receive icons to show them who their fellow agents are, only the leaders are identified, enabling agents to be flipped with the faction none the wiser. Thus, between the trust game and the (awfully, *awfully* named) conspiracies being abundant... Paranoia.

TODO list:

- [x] Create a basic functional skeleton of the mode.
- [x] Write up the strings for factions (welcome texts and the like)

- [x] Create a Spymaster's Uplink 
- [ ] and means to add TCs to it
- [ ] Create items for the Spymaster's Uplink //well, there's one, plus a gutted traitor uplink

- [ ] Test **everything**.

By popular request, I'm taking it off DNM and asking for a merge as a beta admin-set only mode like SCOM was.


code/modules/urist/antagonist/conspiracies.dm
```diff
@@ -0,0 +1,105 @@
...
+/datum/antagonist/agent/buildaborg
+	id = "Buildaborg"
+	role_text = "Buildaborg Group Leader"
+	role_text_plural = "Buildaborg Agents"
+	faction_role_text = "Buildaborg Agent"
+	faction_descriptor = "Buildaborg Group"
+	faction_verb = /mob/living/proc/convert_to_conspiracy
+	faction_welcome = "Help the cause. Do not harm your fellow agents."
+
+/mob/living/proc/convert_to_buildaborg(mob/M as mob in oview(src))
```
Glloyd: it would totally be possible to condense all the convert procs into one single proc. Which would be a lot nicer and easier to work with.

code/modules/urist/antagonist/conspiracies.dm
```diff
@@ -0,0 +1,105 @@
...
+	..()
+	buildaborgs = src
+
+/datum/antagonist/agent/buildaborg
+	id = "Buildaborg"
+	role_text = "Buildaborg Group Leader"
+	role_text_plural = "Buildaborg Agents"
+	faction_role_text = "Buildaborg Agent"
+	faction_descriptor = "Buildaborg Group"
+	faction_verb = /mob/living/proc/convert_to_conspiracy
```
Glloyd: in fact, it seems like that's what you intended to do, because the individual faction procs aren't actually used anywhere yet, you just use the basic conspiracy proc.

scrdest:
Oh, the verb not being the right one is an oversight. Yes, I actually planned to use one proc for it all... but it seemed all-but impossible to do natively in the antag system. Unlike vampire mode code, which has a neat, easy way to access a mob's antag data via a mind datum (mostly due to the update being very rough), the new system is more... hands-off on mobs.

I literally just now realized I could potentially do that by looping through all agent-datum children and calling if(src in current_antags) to grab the faction.

I also initially wanted to get the factions to be generic, initialized per spawned Leader antag, but that seems to be extremely painful to get working, especially without meddling in Baycode much.

code/modules/urist/antagonist/conspiracies.dm
scrdest: Fixed-ish now.

Glloyd:
Sure, we'll run it live to get feedback and whatnot for you.

scrdest:
Dammit, just a quarter too late. I did some things to improve it literally just now. You still here?

PR #346, opened 2016-02-25
Updated Security Rework

CoZarctan:
This is Skeeve2's pr which I applied to mapmerge tools to and fixed the most obvious errors. I had to redo access levels on most things, so it's entirely possible I screwed something up there. It's probably not ready to merge yet, but I'm putting up what I've done so far.

In addition to any concerns about the raised about the design in the original forum thread (http://www.prevess.com/viewtopic.php?f=4&t=216&sid=1b72b6087265b0e60d603b0a271b9315), I noticed the following issues:

-lethal injection syringes seem to have no sprite
-solitary has no flash and no lights
-lethal injection syringes are not locked up
-cell interior/exterior blast doors have their names reversed from what I expect

Obviously a few of these are balance concerns, so I leave them up for discussion.

PR #350, opened 2016-03-01
Updated Sec Rework (redux)

CoZarctan:
As per #346; rather than fix that branch it was easier to start over in a new one. Issues I listed there have been resolved to my preference, except for the invisible lethal injection syringes which should get fixed the next time you merge Bay. Still needs some testing, hopefully someone who isn't me can check for oversights.


Meunahlux:
Criticism from a argie with no life who plays sec and rarely anything else:

-The sec office should be connected to the armory via blastdoors as this makes handing out equipment infinitely easier and not a headache
-Armory should be connected to the warden's for sure, since it's nigh-impossible to detect anybody breaking into it now since there's spess outside
-A permabrig or make the communal harder to break out, since we'll be putting the permabrigees.
-The hallway feels awkwardly long, but that might be just me
-Detective's appears to be missing reinforced walls, not sure if that's on purpose. Makes breaking into sec itself be a 2-tool job.

Glloyd:
I'll do my best to use that feedback and I'll set up a PR to @CoZarctan 's repo

CoZarctan:
Well, the armory does have blast doors that can be opened, and as I understand it, having a permabrig that's allows more interaction with the crew and is easier to break out of is one of the goals of this change.

The point about the detective's office is well taken, though.

Meunahlux:
Gotcha, i can get behind the Perma-Communal brig if i think about it, just was midly concerning that most perme prisioners would try to attack the others, but if i think about it, probably won't happen

Skeeve2:
@Meunahlux Breaking into security was always that easy; it's not limited to the detectives office alone. There is a much easier way through the main hallway maintenance corridor. Similarly, it is far too easy (now that everybody knows) to break out of the permanent brig.

 Communal brig gives a chance for permabrigged people to interact more then anything else, often they are just left to rot. Of course, there will be other drawbacks which are just not preventable, but it will certainty make permabrigging / long sentences more interactive and give the warden a much more involved job. Of course, were experimenting here, were _pretty_ sure this will work... maybe. Only one way to find out!

CoZarctan:
@Glloyd I'd add the black warden outfit to the map, but the warden's helmet is currently defined as /obj/item/clothing/head/helmet/urist/**hopformal** and is thus being overwritten by the HoP's formal cap. I figure you'd want to fix that separately from this PR.

Glloyd:
oops. I'll touch that up and add it locally. For now though, this is good. Seems to have taken the feedback into account, and I'll go over it locally as well to make sure everything's in order. For now though, :+1:

PR #356, opened 2016-03-06
Weather

scrdest:
Adds fog (smoke2 sprite - same as the normal smoke but rectangular and without the animation), rain and an optional raindrop splash weather effects. Moves the blowing snow effect to the common weather effect parent and file as well.

All added sprites support color for variety and adminbuse (either Nien or Corai suggested blood rain, for example).

Fixes being able to pull weather - the effects are now anchored, and makes the mouseover ignore them to avoid the headaches of clicking in between the sprite transparencies.

Rough demonstration of the rain, courtesy of Nienhaus:
rain
The actual effect is faster, doesn't have the odd stutter, less bluish and somewhat transparent, but that's probably gif-creation artifacts. It does a good job of giving a general idea how it looks; it has full tesselation both upwards and sideways.


Glloyd:
I finally have some more time this week to do spess stuff, so I cna go through all the open PRs now. This is perfect because I can use it for the planetside stuff that'll be hitting soon.

code/modules/urist/modules/weather.dm
```diff
@@ -0,0 +1,53 @@
...
+	name = "blowing snow"
+	icon_state = "bsnow"
+	var/safe = 0 //safe == not arctic; ugly, but better than changing the vars on the map
+
+/obj/effect/weather/blowingsnow/Crossed(O as mob)
+	..()
+	if(!safe)
+		if(istype(O, /mob/living/))
+			var/mob/living/M = O
+			if(COLD_RESISTANCE in M.mutations)
```
Glloyd: not necessarily under the scope of the PR, but something that will need to be done is to make this check for clothing with the appropriate temp resistance as well.

code/modules/urist/modules/weather.dm
```diff
@@ -0,0 +1,53 @@
...
+/obj/effect/weather/rain
+	name = "rain"
+	icon_state = "rain"
+	color = null //fully supports colors, the null works but is very subtle
+
+/obj/effect/weather/rain/Crossed(O as mob)
+	..()
+	if(istype(O, /mob/living/carbon/))
+		var/mob/living/carbon/M = O
+		M.ExtinguishMob()
```
Glloyd: good thinking. should also put out hotspots, but that's something that can be added later, for now this is good.

Glloyd:
I just merged #350 so you'll need to update the map file before I can merge this.

Glloyd:
Shit, just merged those map fixes. Take two?

Nienhaus:
I suggested rain blood.

Glloyd:
Actually, just take out the map changes and I'll do it quick locally.

Issue #353, opened 2016-03-05
Map Mispiping


Vakothu:
In the HoS' room, the waste loop is attached to the vent, and the distro loop to the scrubber, and in Telecommunications, most of the distro loop is missing.

Skeeve2:
Possibly fixed in  #350

Issue #351, opened 2016-03-05
Medical lacks syringe gun ammo


liction15:
Due to the changes to guncode syringe guns need special darts now instead of normal syringes. Medical needs to have some darts added to their department at roundstart. This is a mapping issue.

PR #362, opened 2016-03-08
Various map fixes

### LINKED ISSUE #361: Pods sign obstructed
Obstruction

Just need to remove the r-wall (it will also make the 1x1 area there look better,) and rearrange some props.

I'm also spotting a few wiring issues (ex. there is a more efficient route to the wardens APC) and an unused wire going nowhere at the end of the communal cell.

Skeeve2:
Fixes #361  #353 #351

-351, adds box of syringe gun darts under the syringe gun


Glloyd:
:+1:

PR #363, opened 2016-03-09
Monkey genetics fix

### LINKED ISSUE #314: Genetics acting up
Any human type mob put in the DNA scanner yields an ingame error like a Husk or mob with no DNA (DNA structure is ruined beyond recognition etc etc) until they are irradiated (targetted or pulse).

Monkeys have only 27 SEs instead of full 50; editing the 27th SE humanizes them in that they get 50 SE blocks, but remain monkeys.

Spawned humans and monkeys humanized by the method above both have every single SE block set to 000, but UI are untouched and properly randomized.

CoZarctan:
Here's a sort of ugly hack that will fix #314. I'm not sure how we're feeling about ugly hacks to fix Baycode issues, so merge only if you feel it's acceptable. 

My feeling is that really /mob/living/carbon/human/new() or /mob/living/carbon/human/setspecies() should initialize DNA fully for everyone and set the monkey mutation for any primitive species that's spawned, but I don't know if that would have unintended consequences, and also it would be more work (getting this functional took me long enough, but most of that was just trying to remember how Byond code worked).


code/game/dna/genes/monkey.dm
```diff
@@ -37,8 +37,8 @@
 
 
 	var/mob/living/carbon/human/monkey/O = null
-	if(H.species.primitive)
-		O = new H.species.primitive(src)
+	if(H.species.primitive_form)
```
scrdest: I'm fairly sure this If won't do what you want it to; primitive IIRC being a var that stores if a mob currently is in primitive_form, and primitive_form storing what that form would be - so that If would fire for every single race who has any primitive_form set to anything not null. Especially considering the primitive var hasn't been colon'd so it has to exist on compiletime.

Then again, I'm going by memory, I didn't look into what the mob vars are in a while.

CoZarctan: There is no variable named primitive in the species datum. My guess is that it was renamed when species were changed to a new system, and this file wasn't updated.

code/game/dna/genes/monkey.dm
```diff
@@ -83,6 +83,7 @@
 		I.implanted = O
 //		O.update_icon = 1	//queue a full icon update at next life() call
 	qdel(M)
+	O.active_genes |= /datum/dna/gene/monkey //For some reason, this isn't done otherwise
```
CoZarctan: Now if you want something to complain about, it's this right here. There's something weird going on with the whole domutation2 loop that I couldn't figure out, and this is the (terrible) work-around.

code/game/dna/genes/monkey.dm
scrdest: *That's even weirder.*  If there's *no* variable like that, it should throw an error on compile unless someone specifically used a colon to tell it to trust the var will exist.

Naxodile:
The repo janitor has arrived. If this wasn't fixed somewhere in 5 years then maybe it was never meant to be

PR #365, opened 2016-03-12
Paranoia 2.1: The Hephaistos Intervention

scrdest:
Well, I accidentally opened a branch on the main branch, honk. Also, having way too much fun with the shitty airport-thriller titles in case it's not obvious enough.

A ton of last-minute (again >_>) bugfixes and QoL stuff like feedback on getting TCs.

Adds an outfit to dress up as a rather thin and minty human as an alien with a mask that works as a one-use voice changer.


Glloyd:
Good enough for now :+1:

Glloyd:
Sheit

PR #369, opened 2016-03-13
Paranoia 3: The Omega Code

scrdest:
Assuming the antag spawning works, that is, I can't really test it by myself and if the last test is anything to go by, it could take a couple hours out of four people's life to do so. But optimistically, the mode should be fully playable.

**1)** Adds handling of periodic (10-15 minutes) intel spawns. I couldn't decide which method to use so I just used them all and var'd their usage so you can enable or disable landmark spawns and leader/laptop spawns independently of each other on runtime via gamemode debug.

**The leader spawns** are straightforward; it ALWAYS spawns an intel item of the respective faction; if there is a faction leader, it spawns in his storage items or, failing that, his hands, and failing THAT on the ground. The same spawning behavior was added to the uplink laptops.

Speaking of laptops, if there isn't a leader, the same intel will be spawned on and only on the first laptop of the appropriate faction the gamemode can find.

**The landmark spawns** are trickier. They spawn random-type intel on a prob; it can be set on the individual landmark for fine-tuning of spawning, e.g. lower the odds for slightly higher-traffic areas to encourage exploration, or for areas requiring more specific access; defaults to 50%, both as the landmark's var and in case of null value in that var.

At the same time, to allow to potentially add a jillion landmarks for variety, there's a var controlling how many intel pieces can spawn in one 'drop', so that even if all prob(10) landmarks you plopped down miraculously triggered, no more than X will spawn. It's pretty generous for now, it will probably need tweaking.

**LANDMARKS ARE NOT MAPPED IN YET!** I want to do it in a separate PR, because, y'know. Maps.

2) _Hopefully_ fixes the antag spawns. In yesterday's test, only Men in Grey spawned. I noticed I changed the antag id's (because laptops originally used that for the faction name and you pointed out they were lowercase there) but forgot to change antag_tags to match, but since MIG's id was already uppercase, it wasn't changed for them. It, uh, also possibly means the mode might runtime now. :x 

If that doesn't work, it's probably the initial_spawn vars counting all agent subtypes for spawning a particular subtype, which would be painfully dumb to do but unsurprising.

### DIFF
--- a/baystation12.dme
+++ b/baystation12.dme
@@ -1684,6 +1684,7 @@
 #include "code\modules\urist\gamemodes\paranoia\conspiracyitems.dm"
 #include "code\modules\urist\gamemodes\paranoia\conspiracyleader.dm"
 #include "code\modules\urist\gamemodes\paranoia\paranoia.dm"
+#include "code\modules\urist\gamemodes\paranoia\paranoiahelpers.dm"
 #include "code\modules\urist\gamemodes\scom\alienspawns.dm"
 #include "code\modules\urist\gamemodes\scom\civilian_mobs.dm"
 #include "code\modules\urist\gamemodes\scom\machinery.dm"
--- a/code/modules/urist/gamemodes/paranoia/conspiracyitems.dm
+++ b/code/modules/urist/gamemodes/paranoia/conspiracyitems.dm
@@ -55,8 +55,10 @@
 	var/basedesc = "A file containing top-secret data."
 	var/faction = "Broken Code Initiative"
 
-/obj/item/weapon/conspiracyintel/New()
+/obj/item/weapon/conspiracyintel/New(var/presetconspiracy)
 	..()
+	if(presetconspiracy)
+		faction = presetconspiracy
 	var/datatype = pick("blueprints","financial records","operational reports","experimental data","access codes","personnel files","schematics","operation plans","communication frequencies")
 	var/valuedesc
 	value = pick(2,4,6)
@@ -297,3 +299,10 @@
 	..()
 	new /obj/item/clothing/suit/urist/fleshsuit(src)
 	new /obj/item/clothing/mask/gas/voice/fleshmask(src)
+
+/obj/effect/landmark/intelspawn
+	var/probability = 50 //so that it can be tweaked for areas with various amounts of traffic
+
+/obj/effect/landmark/intelspawn/New()
+	invisibility = 101
+	return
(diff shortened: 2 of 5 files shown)


Glloyd:
I still think the better way to do it would be to spawn it on a random floor tile in a specific list of areas, but that can always be added as a third option later on. My only issue is with people memorizing the locations, and with spamming landmarks all over the map, which is annoying in itself as a mapper.

scrdest:
Fair enough on the landmark spam. But it's a tradeoff; either landmark spam or people memorizing the locations, you have one or the other - between the prob and the cutoff point, only a handful of points will spawn intel per round, so you can't really camp on these. If this ever becomes a problem, I could be dynamically editing the landmark spawn probs so that for the next round the odds are even lower for that point, and bring them back to default once a round of not-spawning there passes.

I could edit the landmarks' alpha (without breaking the code for a change) so they aren't as obnoxious on the map.

Area spawning is a pretty rough method, though, since it's a very, very rough method; even if I added some sanity to make them avoid spawning in the walls, you'll have them appearing in the middle of random hallways, looking way out of place, and would probably have to involve going through every tile in every area it has to spawn to boot, which would be hella slow.

liction15:
how about instead of spawning in random/set ares it spawns every time you recruit someone on a random person who is apart of your group or even on a neutral person?

scrdest:
It'll probably sound pedantic, but in this case I really need to know ;_;

Do you mean 'a part', so it should spawn *in* the faction, plus neutrals, or 'apart', as in it would **not** spawn for anyone in the faction, as in opposing factions plus neutrals possibly? 

The former is kind of a thing, except on timer rather than recruitment, with the leader/laptop spawns.

The latter is kind of an interesting prospect, because counterbalance - if it spawns on the other faction - and if it spawns on a neutral, it drags them into the whole mess, which seems rather fun and involving - some rando suddenly gets something very, very valuable to some very, very dangerous men, there's a plot right here.

liction15:
I meant a part of the group and neutrals though your idea is interesting although eventually the neutrals might run out as they are recruited or maybe not, still though a system that helps every crew become involved is nice.

scrdest:
Neutrals running out isn't a major issue - there's already a similar thing in place for the leader spawning for a situation with no leader mob (incidentally, that made me realize I need to investigate some vars, because I might need to check if the leader is not a ghost to avoid ghost spawns).

I can't really pull the second off; even laptops within a faction are largely independent from each other (hell, as a matter of fact, don't tell anyone but the uplink-bought ones don't even get a faction assigned properly because I'm not sure how to put arguments for New() for the laptop in the New() for the datum of the uplink item of the laptop :3).

It would require making some kind of global controller separate from the gamemode specifically for handling laptops, whereas the same general effect is achieved by simply needing to use another faction's intel first.

Glloyd:
Sorry, been a hella busy week. Looks good to me! I'll see if Vac can update tomorrow.

Issue #367, opened 2016-03-13
In the new brig, one of the purple and white tiles around the table is not purple and white.


Vakothu:
This makes me irrationally annoyed. :P

Glloyd:
Oddly enough there's no corner sprite for that one, so I just changed up the pattern.

Skeeve2:
Fixed in #370

PR #370, opened 2016-03-13
Security : Visual, Balance update (& Hotfixes)

Skeeve2:
**Changes**
- Improves the visuals of security, removing OCD inducing areas (#367)
- Increases size of wardens by 1 tile to make it less claustrophobic
- Balances the communal brig so it is more open and removes the horrible looking 1x1 hallway from the cells.
- Balance to visitation so it is required to access it from the security lobby. A window with privacy shutters has been added to compensate for the new required access, this way, it is less of a hazard and looks better
- Removes unnecessary wires

**Hotfixes**
- Adds missing scrubbers and vents to some areas
- Adds missing air alarms
- Rewires some areas to be less intensive
- Adds missing security requirement to north maintenance entryway!! (Area was incorrectly set.)
- Fixes some blast-door button ID issues
- Fixes a disconnected scrubber

It should be unnecessary to update any maps, as the majority of changes are just slight


Skeeve2:
Additionally, I think the N2 tank shouldn't be at the scrubber location, security should be forced to take it out of the storage in maintenance. People will run out of their cells and activate the N2 tank.

Skeeve2:
Do not merge until fixed

Skeeve2:
new pull request soonish

PR #371, opened 2016-03-15
Security Fixes & Changes

Skeeve2:
**Changes**
- Improves the visuals of security, removing OCD inducing areas (#367)
- Increases size of wardens by 1 tile to make it less claustrophobic
- Balances the communal brig so it is more open and removes the horrible looking 1x1 hallway from the cells.
- Balance to visitation so it is required to access it from the security lobby. A window with privacy shutters has been added to compensate for the new required access, this way, it is less of a hazard and looks better
- Removes unnecessary wires

**Hotfixes**
- Adds missing scrubbers and vents to some areas
- Adds missing air alarms
- Rewires some areas to be less intensive
- Adds missing security requirement to north maintenance entryway!! (Area was incorrectly set.)
- Fixes some blast-door button ID issues
- Fixes a disconnected scrubber

It should be unnecessary to update any maps, as the majority of changes are just slight


Glloyd:
traaaaaavis

Glloyd:
:+1:

PR #373, opened 2016-03-17
Security final balance changes & fixes

Skeeve2:
Changes
- ~~Increases size of maintenance around security slightly~~ Only one corridor now
- Changes brig exterior window to shutters controllable from the warden desk, in case of rowdy prisoners
- Adds locally-stored books, allowing them to be more complex and not rely on a webpage, includes short documentation / template. May be used later to introduce uristlaw.
- Moves some objects around for better visuals
- ~~Removed mini-hoe as it is too viable as a weapon~~

Fixes
- Visitation shutter direction given proper alignment
- Fixes broken doors
- Adds more lighting to dark areas
- Removed laptops until fixed
- Fixes table that was both reinforced and not
- Adds proper networks to security cameras ++
- Removes non-working / unnecessary flasher buttons

[] @Glloyd **Suggested fixes/changes** Completed

- Added prison/exodus networks (even to some missed cameras)
- Minihoe kept (?) 
- Flasher fixe 
- Fixed non-electrified window
- Adds missing visitation button in visitation (not on picture)
- Puts even more tables to make north-east corner feel less empty

**There will be NO MORE SECURITY CHANGES OR BALANCE UPDATES after this update. ONLY patches!**


Skeeve2:
Do not merge until https://docs.google.com/spreadsheets/d/1xmbbub0F1bMAeLayV5Av55qpIYJRnCeXpFL2laGDz18/ is evaluated.

Skeeve2:
Also makes all books centered in table. Cool I guess?

Skeeve2:
- Also adds two minor hot-fixes to armory, adds a missing light and fixes the table (it's one type only now!)

Skeeve2:
Don't merge this yet

scrdest:
I think Sec armory may be overpressuring on the current iteration.

Skeeve2:
> I think Sec armory may be overpressuring on the current iteration.
This has already been fixed in previous update (not yet pushed to server)

Glloyd:
Pictures please.

Glloyd:
Looked at it locally. The networks on the cameras should be both prison and Exodus so they can be viewed from regular camera consoles, the east windows of the brig are no longer electrified, and that section just feels empty for some reason. The shutters in the centre of the visitation room have the only matching button in the Warden's office, which shouldn't be the case. You also removed the only other prison-wide flash, which should also not be the case. The brig should be able to function near normally if no player picks Warden or HoS, and this is not currently the case. Finally, those two-tile wide spots in maint are really unnecessary for such a small unimportant strip in maint, and that 4 tile spot should be cut down to three. Maint's meant to be one tile outside of secrets, pumps, airlocks, solars and certain rooms. If you really want, that second 2 tile stretch to the north can be kept, but either way please remove that one tile with the box on it.

Also, you say you removed the laptops and the minihoe, but they're still there? Either way, the mini-hoe should be kept, it's only force 5, and the old permabrig had much worse things in it.

Skeeve2:
Apparently, there can only be one prison-wide flash at any given time. I do not know why this is. I tried making a new one and placing it in various locations, but nothing seemed to work. I even directly copied the one from the wardens office and it still didn't work. 

Also, for some reason, this isn't all the changes I had made, I have to redo them

Skeeve2:
all donnnnnnnnnneeee

Glloyd:
beauty!

PR #377, opened 2016-03-20
Adds some more clothing.

Nienhaus:
I will add pictures some time.


Glloyd:
Pictures?

Nienhaus:
Green plaid is the one from an NPC for SCOM
wvzkcvd
The black is like this but black.
fd6hxqq
79vwy8x
pogpgto

Glloyd:
Sexy

PR #390, opened 2016-03-28
Anti-spam

scrdest:
Uses THE ANTI-SPAM MEASURE WE ALREADY HAD FUCKING IMPLEMENTED BUT DIDN'T ENABLE. ARGH.

Enables /client/proc/handle_spam_prevention() handling for OOC, ahelps et al which automutes successive identical messages above a threshold. It can be worked around by a spammer with half a braincell to show (so, pretty much none of them), but it's a matter of tweaking the existing filter.

### DIFF
--- a/config/config.txt
+++ b/config/config.txt
@@ -221,7 +221,7 @@ SOCKET_TALK 0
 #TOR_BAN
 
 ## Comment this out to disable automuting
-#AUTOMUTE_ON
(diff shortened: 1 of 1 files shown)


Glloyd:
Huh. I totally forgot this existed. gg

Glloyd:
Fucking github won't let me merge hit from my phone anymore, goddammit.

PR #391, opened 2016-03-28
Rebalances Secret odds

scrdest:
Everything was doubled for more granularity; Autotraitor unchanged, Wiz/Ling bumped up to 3 from 2 to roughly maintain the odds, 'normal' non-traitor modes bumped down to 5 from 6 to make them slightly rarer than traitor but still common, Malf a shit so bumped down to 1 from 2.


Glloyd:
What this means is that Malf and Ling will almost never be picked by secret, and when you hit the pop req for modes like nuke and whatnot, it'll be more likely to go to traitor. I dunno, I feel like this is something that should be polled on, but at the same time it seems kind of inconsequential because not too many people like ling and malf.

Addle:
I feel that reducing the chance for highpop modes like merc would be for the better. We've had an absolute overwhelming amount of those, and every time people get angry. Would be best to put it on a burner for a bit. We have gamemode votes if they need their merc fix.

scrdest:
I've done the math; the odds for Merc et al are not dramatically decreased in comparison; with all modes available, it's 16% chance for either of them vs. 19% for Autotraitor when both are available.

Malf and Ling being nigh-unpickable by Secret is wholly and entirely by design; New Malf is not even enjoyable *for the malf AI*, basically being Extended until the last 15-ish minutes when everyone dies in a nuclear fire or the AI mercifully researches Recall Shuttle just to cancel autorecall. I initially wanted to keep ling unchanged, but Corai held a vote, and it was a landslide in favor of that change.

Porting tg-ling from Polaris may make it less shit and make it restored back, but for now, there's always mode votes and a 3% chance for them happening literally every round, since both have basically no Secret requirements.

Glloyd:
we'll give this a shot for now, see what people think.

PR #403, opened 2016-04-03
Minimaps: Cave, Chapel

scrdest:
Two new minimaps for the jungle surface: a small cave surrounded by thick foliage, illuminated by glowshrooms, with a very small chance of loot - should be a decent hideout, and a Cult chapel with a fully-functioning Sacrifice rune and in a rather disturbingly good repair.

Cave
Chapel


Glloyd:
Yay! I'm glad people are getting excited about this system. There's a lot of fun things that can be done with it, I've got some good ideas for underground maps that I'm going to mess with later in the week. For now though, this is great!

PR #406, opened 2016-04-03
Minimap: Crashed escape pod

CoZarctan:
An escape pod crashed on the planet. Looks to have been a rough landing.

capture


Glloyd:
I love it! but you should remove the docking controller from there because it will likely runtime. And the display panel might show the ETA for the shuttle coming in.

CoZarctan:
Oh right, I forgot status displays work without power. I'll take them both out.

scrdest:
Maybe make the pilot have stumbled back and not to the side and shrink the map on the y? There isn't much there otherwise, it kinda feels like a waste.

Also: how about some spilled fuel trailing?

CoZarctan:
Spilled fuel is a good suggestion. My intention with the long y was to make the corpse not immediately visible to someone who stumbled upon the pod. I guess it's sort of waste, but that space would still likely be jungle if I shrank the map.

Glloyd:
The y is only 12, not that big of an issue, and CoZ is right that the space would still be jungle if the map was shrunk.

scrdest:
Unless it's a river. That's why I'm asking - I've been spawning those a bit today, and fitting them to the river is a bit of an issue.

Glloyd:
The river is static now though, it's not random anymore.

scrdest:
Yeah, I mean it's tough to place it to miss the river, since the thingy centers on you.

Glloyd:
true, but i've set it up so the templates only spawn to the east of the river, and if all maps are kept within the 35x35 size limit, there won't be any issues.

Glloyd:
I'm alright with merging this as it is. If you want to add anything more to it, just set up a PR and I'll merge it super quick.

PR #408, opened 2016-04-03
Minimaps 2: Sunken Chapel, Underground Cave-in

scrdest:
Notice a theme yet? Guess my mind is stuck on those ideas today... I forked up forking, so it's accidentally from Urist.

Adds two more new minimaps: a caved-in cave with an xenoarch artifact and some trapped miner corpses, and a chapel that has suddenly sank into the cavern below, sustaining heavy damage, taken over as a NPC vampire hideout, with two of the initial four having been killed already and only present as remains - mostly so it's not ALL safe on the planet.

Tweaks to the maps added earlier today: the surface cave's interior has been retex'd to the asteroid floor icon, changed the placement of glowshrooms and shrunk the map (only the jungle turfs), and the cult chapel has had the sand replaced with clear light jungle, since it looked ugly with no border between the sand and jungle.

caveinshowcase
sunknchapel


Glloyd:
While you're at it, remove the glowshrooms. They still runtime apparently.

scrdest:
Shit. That explains some things.

scrdest:
OK, removed 'em.

Glloyd:
Awesome. I'll set up template spawning for the underground section either tomorrow or tonight.

PR #413, opened 2016-04-04
Makes the planet surface use dynamic lighting

scrdest:
Turfs emit light via the light system proper, since luminosity setting was hacky.

Makes the cave floor a legit turf type and not an icon edit. Edits the rock wall not to be grass in desc. Adds an (unused) sun effect.

Tested, works. Minor stutter at the roundstart to update it all, but no issues later.

### DIFF
--- a/code/modules/urist/modules/jungle/jungle_areas.dm
+++ b/code/modules/urist/modules/jungle/jungle_areas.dm
@@ -16,7 +16,7 @@
 	name = "jungle"
 	icon = 'icons/jungle.dmi'
 	icon_state = "area"
-	lighting_use_dynamic = 0
+	lighting_use_dynamic = 1
 	luminosity = 1
 	base_turf = /turf/simulated/jungle/clear/grass1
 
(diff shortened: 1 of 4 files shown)


Zuhayr:
What is the purpose behind using an object rather than just adapting the starlight code?

scrdest:
I'm not using it right now. But the idea is to make it directional, to give visual feedback as to the sun's current position. Potentially adapt it to the vampire check_sun(), too.

scrdest:
Well shit, I accidentally rouged it while trying to push an unrelated fix. Want me to revert it?

Glloyd:
well shit. i never got a chance to look at it.

On Mon, Apr 4, 2016 at 2:56 PM, scrdest <notifications@github.com> wrote:

> Well shit, I accidentally rouged it while trying to push an unrelated fix.
> Want me to revert it?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/UristMcStation/UristMcStation/pull/413#issuecomment-205446810>
>



-- 
Graham Lloyd

scrdest:
I can say for sure that it works as intended, since I tested it. The only thing of interest would be that the turfs' New() calls update_lights (otherwise it was all dark aside from stuff that lights up on roundstart like the RD Outpost), so there might be slightly more stutter on roundstart, but it might as well be caused by running the overlays on the whole map. No additional mid-round lag.

Issue #349, opened 2016-02-27
Hydroponics is Buggy


liction15:
If you try to water the plant, it goes into the plant as a reagent, not into the tray. Also unable to replace the nutriment in the tray.

scrdest:
Worked for me. Nutriment sloooowly upped, but worked too.

cgale212:
Sprayed a whole thing of plant-b-gone and the plant was still healthy. From the plant analyzer: - plantbgone, 100 unit(s). I believe if you try to add fertilizer to a tray with no plant it will tell you the tray is full.

liction15:
Hydroponics is still broken which is a shame as it makes xenobotony very hard to do for long period of time as well as hydroponics.

Glloyd:
should be fixed

Issue #415, opened 2016-04-06
Thown fruit injects it's reagents directly into the bloodstream.


Vakothu:
We now have the "touching" chem-holder for reagents, so thrown fruit should probably put the reagents they have in that, instead of direct injection into the blood stream.

Glloyd:
fixed

Issue #303, opened 2016-02-06
Botany stuff


Werty892:
Issue 1:No rolling papers. What's the point of drying your things if you can't blunt em?

Issue 2:All trays have a light level of 0 lumens, despite being in the light.

Edit: We don't even have drying racks AAAAAAAAAAAAAAAAAAAAHHHHHHHHH

Glloyd:
Added a drying rack to the map, re issue one we'll need to add those, I can't find them in the code.

scrdest:
They weren't added in that commit on BS12. They might've been intended to be added in a separate PR, IIRC, but then, well, Bay coderbus lost its wheels.

Glloyd:
i think that's all. good shit.

PR #416, opened 2016-04-06
Hydroponics fixes

CoZarctan:
Fixes issues #349 #415 #303 

-Duplicates hydroponics tray fix from Baystation issue 11661
-makes splattering fruit use the reagent splash() proc rather than adding its contents directing into mobs' blood
-Adds rolling papers from /tg/ for use with the drying rack. I think that ambrosia produces less space drugs than they used to since the plant update, so I had to raise the chem ingestion multiplier for cigarettes so that you could actually get high from rollies. This may have to be tweaked further.


scrdest:
Screw making Botany usable, #303 is fixed by this PR. :+1: :heavy_multiplication_x: :100: 

Seriously though, it does look alright code-wise at a glance. Tested?

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -117,7 +117,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 					if(H.species.flags & IS_SYNTHETIC)
 						return
 
-				reagents.trans_to_mob(C, REM, CHEM_INGEST, 0.2) // Most of it is not inhaled... balance reasons.
+				reagents.trans_to_mob(C, REM, CHEM_INGEST, 1) // Most of it is not inhaled... balance reasons.
```
Glloyd: this is the only part that i'm worried about, but we'll see how it plays out. maybe .5 will be better, but i'm not sure how much ambrosia was nerfed, so as with most balancing things this'll have to be tweaked.

Glloyd:
Besides that :+1:

CoZarctan:
Hard to test every possible edge-case, but things like throwing thermite tomatoes at people and then setting them on fire works. 
I added a pack of 10 rolling papers to botany, but after those run out the only way to replenish them is by praying to Stoner Jesus. How do y'all feel about a cargo crate or something?

Glloyd:
Yeah. If there's a cargo crate with hydro supplies, toss them in there.

scrdest:
Why not add them to autolathe as a wood product?

CoZarctan:
Well, IRL I believe rolling paper is not made from wood, but instead other plant fibers. (Should we care about that?) I could see it potentially as a bio-generator item. This is why I asked, there are a few places they could conceivably go.

Glloyd:
If you're going to add it to something, add it to the pulp and paper processor

liction15:
rip deadly ggnore fruit

PR #418, opened 2016-04-07
Adds carpentry tools to mining cyborg modules

bknight286:


Glloyd:
good stuff! I'll add a check for silicons to the woodworking machines.

Glloyd:
uh, i tested it on the server and it worked. i dunno what you were doing wrong earlier. Either way, less work for me!

PR #425, opened 2016-04-10
Misc map fixes

CoZarctan:
-Fixes a broken conveyor belt in disposals
-Removed a conveyor switch in the mail office that wasn't hooked up to
anything
-The lockers in the virology access airlock had no biohazard suits


CoZarctan:
I put the change in the third commit separately in case it needed to be reverted for some reason. Does this thing have a purpose?
capture
It's a vent hooked up to scrubber pipes so I don't think it actually does anything right now, but is it supposed to?

Issue #428, opened 2016-04-10
No Power Net Sensor for engine outputs


liction15:
There is no way to measure the power output of both engines via console without rewiring it to an existing power sensor. The current sensor only measures the power input between the supermatter SMES and APC. The singularity input lacks a sensor completely.

CoZarctan:
You can always use a multitool on the wire leading into the SMES. Because of the way the sensors function, I'm not sure there's a nontrivial way to just stick one in there and have it work.

PR #431, opened 2016-04-10
Engineering Map Fixes

CoZarctan:
-Adds a gas sensor to the supermatter chamber, so that the monitoring console actually works
-Adds a powernet sensor to the singularity input as suggested in #428
-Fixes some blast doors with the wrong id
-The button to remote control the engine emitter now functions
-Fixes an error where you'd initially have to press the button twice to
unbolt the door in the SMES room


CoZarctan:
I'm not too sure about having those buttons I fixed inside the engine monitoring room, since pushing them doesn't have an immediately visible effect. The emitter button does at least give a feedback message, but it doesn't seem to have any uses besides making it easy to accidentally cause a melt-down.

liction15:
its a holdover from when you can see directly into the supermatter engine but there is always the camera console next to it to check in an awkward way.

Issue #430, opened 2016-04-10
[REQUEST] A break random event, and species locking the racial hardsuits.


Vakothu:
I sent these suggestions to Glloyd via Steam, but he might have forgotten by now.  Racial hardsuits being the Skrellian Voidsuits, and the Unathi NT Breacher RIG.  As I've seen people, as humans, grab the racial suits simply because they have good armour or because of how they look.

liction15:
It should be pretty easy to do this, I think it would only need adding one line of code. I'll look into trying to do this myself.

liction15:
okay so looking at the code it seems that skrell hardsuits are already restricted to certain species but its restricted to both skrell and humans. Im not really sure if this is intended but I can make a pull request just in case glloyd thinks that's fine.

Glloyd:
they're on my list, don't worry.

On Sun, Apr 10, 2016 at 6:20 PM, Michael <notifications@github.com> wrote:

> I sent these suggestions to Glloyd via Steam, but he might have forgotten
> by now. Racial hardsuits being the Skrellian Voidsuits, and the Unathi NT
> Breacher RIG. As I've seen people, as humans, grab the racial suits simply
> because they have good armour or because of how they look.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/UristMcStation/UristMcStation/issues/430>
>



-- 
Graham Lloyd

PR #434, opened 2016-04-11
[REQUEST] Species locking the Skrellian hardsuits

liction15:
Does half of #430


liction15:
ffff build failed species restriction is undefined arg.

liction15:
I figured out what i did wrong I changed the wrong file for the RIGs

liction15:
I ran a server to test the change I made to skrell hardsuits and i can now confirm that the change appears to be working as intended.

liction15:
welp seeing as this is already on glloyd's todo list i guess ill close this for now.

scrdest:
Eh, I don't see any reason to double the work.

Glloyd:
I'll take a look at this tomorrow, should be good though.

Glloyd:
Mind adding the restriction to the Unathi rigs on here?

liction15:
I would if i could figure it out but I cant because unathi rigs seem to use a different path thus why adding species_restricted throws an undefined variable error.

Glloyd:
Alrighty.

Vakothu:
You'll probably need to put the species_restricted flag on the body of the Unathi rig, not the back-worn control panel.

PR #457, opened 2016-04-30
Balaclavas with mouth-holes!

scrdest:
Adds a balaclava with a mouth-hole item (called ski mask for differentiation).

balaclava
balaclava_onmob
bal_onmobs


Glloyd:
Should add an attackby for the existing balaclavas so you can cut a mouth hole into them.

scrdest:
There's an idea! There's a minor continuity issue though, since the basic balaclava actually has a single large eye-hole, and this one has two separate ones. So I guess you also stitch it up too?

I mostly made this a separate item instead of a replacement for ease of merging.

PR #476, opened 2016-06-04
Update config.txt

Nienhaus:
Fixes site links.


scrdest:
oh noes it'll totes break even the features that weren't even coded

PR #483, opened 2016-06-12
Adds toeless jackboots to Xeno and Sec wardrobes

EyesNoEyes:


scrdest:
I don't see a reason to put them in Xeno wardrobe. It has an actual gameplay effect (damage against downed targets increased), and is equippable by anyone, so people might steal them as non-Unathi just for that.

That said, it's as easy to get the jackboots via loadout.

EyesNoEyes:
Right. I understand. I figured it would make little difference considering, as you mentioned, you can start with it, and just wanted to increase the number that were available on the station for those that were unfortunate enough to lose theirs, since it was impossible to acquire them any other way beyond prayer..

Glloyd:
@EyesNoEyes take the jackboots out if the Xeno locker and I'll merge it.

EyesNoEyes:
Done

PR #484, opened 2016-06-13
Waistcoat as an accessory

Kydrasz:
If I did it right then the title is enough info. Seeing as I did this on my phone, I doubt it.


Kydrasz:
Note to self: Remember to use wcoat instead of waistcoat

scrdest:
Once you sort out the mapping errors, please test to see if equipping them on roundstart/admindress works. I know no accessory I'm aware of starts attached on roundstart currently.

Kydrasz:
I'll re-do this once I fix the mapping errors.

scrdest:
You don't need to close the PR to fix it, but your call.

Glloyd:
It'd be easier to make a separate waistcoat as an accessory. Changing paths will break stuff with Bay pretty hard.

PR #494, opened 2016-09-14
Update admins.txt

Nienhaus:


Glloyd:
As I said on discord, needs to be Trialmin.

Nienhaus:
I was mostly showing this as an example how easy it is.

PR #510, opened 2016-10-11
Spoop cleanup

scrdest:
Genericizes and cleans up zombie code. Subtypes are just var presets, types are interchangeable with some VV. Regenplagues no longer start dead. 

Vampire NPCs now are gendered randomly, may introduce gender overrides later if someone wants them. Also both vamps and zombs are now hostile/urist subtypes, mostly for cleanup purposes.

### DIFF
--- a/code/modules/urist/mob/spoopymobs.dm
+++ b/code/modules/urist/mob/spoopymobs.dm
@@ -1,4 +1,4 @@
-/mob/living/simple_animal/hostile/zombie
+/mob/living/simple_animal/hostile/urist/zombie
 	name = "zombie"
 	desc = "Dead man walking - and hungry for your flesh."
 	speak_emote = list("groans")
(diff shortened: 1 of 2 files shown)


code/modules/urist/mob/spoopymobs.dm
```diff
@@ -206,8 +218,16 @@
...
 	minbodytemp = 0
-
-/mob/living/simple_animal/hostile/vampire/female
-	icon_state = "vampire_f_s"
-	icon_living = "vampire_f_s"
-	icon_dead = "vampire_f_d"
+	ranged = 0
+
+/mob/living/simple_animal/hostile/urist/vampire/New()
+	var/danglybits = pick(0, 1) //random gender
```
Glloyd: nice

Glloyd: looks good. Still holding out for zombie mode someday.

Glloyd:
looks good. Still holding out for zombie mode someday.

PR #516, opened 2016-10-19
Halloween outfits

Nienhaus:
I tested them, they actually work surprisingly. To lazy for pictures w/e.


Glloyd:
I don't see any issues with this.

PR #520, opened 2016-11-30
Gives coders (read: myself) +BUILDMODE

scrdest:
Because they already have all the component parts, only have to do it the annoying piecemeal way. Will close if consensus is nope.avi.


Glloyd:
yeah, I don't really see why not.

PR #522, opened 2016-12-03
Random back-end fixes:

scrdest:
!!Makes Hostile SA's open doors (among others) by adding a step_towards so they Bump the immediate first obstacle. May result in some sub-optimal paths, but worth it IMO.

Removes all boilerplate lumcounters from Vamp, Harvester, Stalker mob and the Shadow power, replaced with a new common helper, shadow_check and the helper it uses, get_light_amt. Can take an optional parameter to ignore red lights to make them scarier.

Makes skeltals not need atmos on account of not having lungs.

Fixes stalkers occasionally picking new_players as victims, makes them pick new victims if previous is dead (or nulled), along with safeguards against endless loops if they can't find it on lowpop. Normally only picks players with minds attached.

Admin magic can make the stalkers use a new behavior of making lights flicker around the victim by varediting a flickerlights var (whodda thunk).

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/hostile2.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile2.dm
@@ -8,7 +8,7 @@
 	var/atom/target // /vg/ edit:  Removed type specification so spiders can target doors.
 	var/attack_same = 0 //Set us to 1 to allow us to attack our own faction, or 2, to only ever attack our own faction
 	var/ranged = 0
-	var/rapid = 0
+	var/rapid = 0 //how many *additional* shots per burst, so set to 1 for 2-shot bursts etc. 2 is original behavior (3 rd. burst)
 	var/projectiletype
 	var/projectilesound
 	var/casingtype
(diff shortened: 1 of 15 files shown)


code/modules/mob/living/simple_animal/hostile/hostile2.dm
```diff
@@ -169,6 +169,7 @@
 	LostTarget()
 
 /mob/living/simple_animal/hostile/proc/Goto(var/target, var/delay, var/minimum_distance)
+	step_towards(src, target) //weird but necessary so they try to bump openable obstacles
```
Glloyd: I'll have to take a more in depth look tomorrow when I have more time, but right now it seems odd to be calling both walk_to and step_towards. I feel like there's a lot of issues that could come out of calling both. For instance, would this not override the use of minimum_distance below?  If a mob has a minimum_distance set, won't they just get stuck walking back and forth between two tiles to maintain their minimum distance? Have you tested movement much with this?

scrdest: It won't override retreat distance (so the distance that forces retreat), because Goto isn't called **at** retreat distance, so if it got called, there's at least one step from retreat distance:

```
			if(target_distance <= retreat_distance)//If target's closer than our retreat distance, run
				walk_away(src,target,retreat_distance,move_to_delay)
			**else
				Goto(target,move_to_delay,minimum_distance)**//Otherwise, get to our minimum distance so we chase them
```

They are already prone to 'cycling' without this commit (in testing, I discovered it can be used for an interesting hit-and-run melee behavior if you set retreat higher than minimal), but I threw in a min_dist check in there. They won't open doors if at minimum distance right now, but they won't cycle.

I would have done it with get_step_towards, but the fucking thing uses shitty built-in BYOND procs with no documentation on its architecture so I can't emulate its failure behavior well enough and I don't want to make snowflake checks for doors.

Glloyd:
theres a helper proc somewhere, get_step_towards2 that might address that. May be worth looking into.

code/modules/mob/living/simple_animal/hostile/hostile2.dm
Glloyd: theres a helper proc somewhere, get_step_towards2 that might address that. May be worth looking into.

scrdest: Nope, get_step_towards2 looks for and returns free turfs on the path to target respecting dense objects though it does use a blunt get_step() if no free step is found. 

The issue is that step/walk_to, which Goto uses, does **not** try to path through dense objects to see if they can be opened, just gives up.

I could go and try to implement walk2 procs that use g_s_t2 (they're probably very simple loops of get_step and step) buuut it would be a bit of a bitch without much gain. Especially now that there's a check so a plain ol' step_towards won't override minimum distance.

For that matter, turns out we do have A* implemented. It just isn't used by simple_animals, but it is used by, say, Beepsky, which explains why he could open doors. TMYK. Maybe I'll just make a not_so_simple_animal class one day, would be cleaner to merge anyways...

scrdest:
Any further concerns?

PR #527, opened 2016-12-26
[READY] Weather Revamp

scrdest:
Completely revamps the weather effects. They are now dynamic (if you want them to) and act over time rather than on Crossed().

This is accomplished by a three component system:

* New weather controller. The performance cost is kept relatively low thanks to a 'ping-pong' system where a localized weather object reports to the controller's list for processing when initialized, is nudged to see if it's responsive/active (for weather change/effects respectively) and removed from processing overwise, keeping it GC-able. It oversees weather changes and inflicting effects where the need arises.

* Weathertype object. Handles overlays and stores effects so that multiple can be kept per tile without spamming twenty objects directly onto the turf. Obj so it can have alpha, color and layer edits mostly.

* Weather object. The direct Thing What Makes You X. Its job is mostly to sit in some loc and be persistent there. Has two bools, dynamic = 1 and safe = 0 that respectively disable the weather change behavior and the effect inflicting behavior when flipped. They are not (unnecessarily) processed by the controller at all in the respective calls then.

* Area var additions: weather list var, so that weather is consistent between objects in a zone, and climate list var, a whitelist of weathertypes for the area. A maximum of two is picked currently.

A number of weather effects and corresponding static presets have been added, and some of the old ones changed. Those are arbitrary distinctions, the reagent/temperature behaviors can be put anywhere you want, not just these subtypes.

* Rain-type weathers splash mobs (only mobs rn) with reagents. Variants: acid, blood and fallout (w/ uranium).

* Snow and wind type work via exposure (heat/cold) by hijacked atmos code, so you can make atmos simulacra. They respect cold protection gear, which has been added to a ton of Urist clothes. Three cold types for various degrees of severity plus a heating one (quite weaksauce).

* Sandstorm deals direct brute reduced by clothing coverage. Plus there's splash and fog, aesthetic only.

Also fixes cryo hardsuit icon issues since same file changed, but no great loss if it doesn't get in through here.


Glloyd:
Will review tonight, just visiting family right now.

code/modules/urist/modules/weather/weather_defines.dm
```diff
@@ -0,0 +1,168 @@
...
+
+/obj/weathertype/proc/GetWeatherEffect(var/turf/T) //handles specific effects of weather on... stuff
+	return 1
+
+/*hijacked from human/life.dm's handle_environment and simplified;
+all-purpose cold/hot weather helper for exposure effects, wear a hat */
+/obj/weathertype/proc/TemperatureEffect(var/mob/living/M)
+	//Body temperature adjusts depending on surrounding atmosphere based on your thermal protection (convection)
+	var/temp_adj = 0
+	if(weathertemp < M.bodytemperature)			//Place is colder than we are
```
Glloyd: So the weather doesn't actually change the temperature of the turf? That's good, the less atmos overhead, the better.

code/__defines/items_clothing.dm
```diff
@@ -148,7 +148,7 @@
 #define BODYTEMP_HEATING_MAX            30  // The maximum number of degrees that your body can heat up in 1 tick,   when in a hot  area.
 
 #define BODYTEMP_HEAT_DAMAGE_LIMIT 360.15 // The limit the human body can take before it starts taking damage from heat.
-#define BODYTEMP_COLD_DAMAGE_LIMIT 260.15 // The limit the human body can take before it starts taking damage from coldness.
+#define BODYTEMP_COLD_DAMAGE_LIMIT 273.15 // The limit the human body can take before it starts taking damage from coldness.
```
Glloyd: I'm not sure this change is really necessary. 0 degrees celsius shouldn't start inflicting damage on somebody, especially when compared to the heat threshold.

code/modules/mob/living/carbon/human/species/species.dm
```diff
@@ -100,7 +100,7 @@
 	                                                  // (also affects temperature processing)
 
 	var/heat_discomfort_level = 315                   // Aesthetic messages about feeling warm.
-	var/cold_discomfort_level = 285                   // Aesthetic messages about feeling chilly.
+	var/cold_discomfort_level = 295                   // Aesthetic messages about feeling chilly.
```
scrdest: That is **BODY** temperature. It works on a differential, so to get bodytemp to 21 C you need something at least in the 10C range. Optimistically; the mild snow stabilizes bodytemp at about 283K only standing still (moving actually raises bodytemp, TMYK) and it's almost -10C. Mind you, stadard atmos probably warmed them a bit, but it will too on unsim/planetsim.

It had been absurd that you could have mobs' bodies be chilled well to sub-zero temperatures before they so much as get uncomfortably cold.

code/modules/urist/modules/weather.dm
```diff
@@ -3,52 +3,134 @@
...
+/obj/effect/weather/invariant/blowingsnow
+	name = "blizzard"
+	active_weathers = list(/obj/weathertype/snow/arctic)
+
+/obj/effect/weather/invariant/rain
+	name = "rain"
+	active_weathers = list(/obj/weathertype/rain)
+
+/obj/effect/weather/invariant/bloodrain
+	name = "blood rain" //ow the edge
```
scrdest: This and the sandstorm.

code/modules/urist/modules/weather/area_climates.dm
```diff
@@ -0,0 +1,12 @@
+//'Climates' (definitions of weather probabilities per area as lists) go here. You can use weights as for pick().
+//NOTE: /clear weather's effect is overriden by all other types! If you have rain and clear, it's effectively rain.
+
+/area
+	var/list/climate = list(/obj/weathertype/clear) //all potential area weather types
+	var/list/weather = list(/obj/weathertype/clear) //current pick for all objects in area
+
+/area/jungle
```
scrdest: You can, it uses a decorated pick(), I had them but removed them for testing. I didn't check if for() respects them as one list object.

code/__defines/items_clothing.dm
scrdest: See above, 0C weather won't even bring you near the *DANGER ZOOOONE* and even then the damage on the first threshold stabilizes in around 1 point of Burnloss, at least relatively short-term. Also those defines literally do nothing since species data are in, I believe :^)

I didn't want to edit heat threshold because I wasn't sure about balance side-effects with fires and such or I'd lower it too.

scrdest:
Good point on simple_animals; worst case they can be excluded from processing, but I needed to run a profile on it anyway.

code/modules/urist/modules/weather/weather_defines.dm
scrdest: That's the idea. Should work even on unsim, too.

scrdest:
Okay, large-scale initialization is, uh. Unadvisable, right now. It's not even an issue of actual weather's performance, the initialization is slow as fuck right now, but I have an idea how to remedy that.

Glloyd:
What're your plans to improve the initialization?

scrdest:
The main processing hog right now is all the new() calls by weatherobjects instantiating their weathertypes. I'll have to remember to stagger area updates by throwing a prob() in there. 

Plus I talked with Bay coders and new() increases in cost with the amount of objects, so having one object new another is not even **linearly** expensive.

I am trying to cut it down by attaching the weather instance to the areas instead, since they are one object each. I'd much rather just work with types altogether, but turns out you can't quite call type procs directly. Gonna hack at it some more.

Glloyd:
That was one of my concerns with the obj route, spawning that many instances is going to cause a fair bit of slowdown no matter how much you optimize it. We could use a weather toggle for stuff like assault and scom where the jungle isn't in use.

scrdest:
Ah, well, uh...

Turns out it wasn't quite the case. I just put a very ill-conceived trick in there (sleep()ing to wait for the controller setup) that locked up the world setup. On its own, it's not much worse that the spawning of all the trees and bushes for the jungle.

Right now the issue for upscaling this is staggering the updates in time, since the jungle is one xbox huge area. That, and waiting to for the controller setup now that I had to change it.

Glloyd:
Will take a look at this tomorrow.

Glloyd:
*6 days later*

data/mode.txt
```diff
@@ -1 +1 @@
-extended
```
Glloyd: This file still needs to be removed

Glloyd:
Okay, remove that file and we're good to go.

scrdest:
I decided not to push it, but if you want the jungle to have weather, just throw in a weather_enable() in the New() and wallpaper the plating outside with the enablers.

Glloyd:
Sounds good, I may or may toss that in along with some of the other stuff I'm working on.

scrdest:
>May or may.

So, will you toss that in, or perhaps you should consider tossing that in? :^)

Glloyd:
wew. May or may not, but probably will.

PR #528, opened 2016-12-29
Guncargo Enhancements

scrdest:
Adds boobytrapped secure crate subtype. Attempting to break them open (with lasers or explosions) triggers a trap, either a grenade, a mine, or if none set, a 0,2,4,5 explosion. If a var is set, the trap item is deleted once the trap is disarmed so that there's no free frags.

Changes all the murderboney Cargo Sec crates to use those crates. Emagging and EMPing works as normal, so antag QM can still get weapons, but regular guncargo will appreciate the term 'secure crate' painfully.

By default both types will use fragbombs, so the effect will be painful, but not stationfucky. If you want to peer pressure certain crates from guncargo extra hard, use the trap = null variant.

### DIFF
--- a/code/datums/supplypacks/security.dm
+++ b/code/datums/supplypacks/security.dm
@@ -35,7 +35,7 @@
 					/obj/item/weapon/storage/box/flashbangs,
 					/obj/item/weapon/storage/box/teargas)
 	cost = 40
-	containertype = /obj/structure/closet/crate/secure/weapon
+	containertype = /obj/structure/closet/crate/secure/boobytrapped/weapon
 	containername = "\improper Weapons crate"
 	access = access_security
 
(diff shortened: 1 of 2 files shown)


Glloyd:
In theory I like it, in practice, I'm not as sure. We're replacing an infinite amount of guns with an infinite amount of potential explosives, so instead of guncargo, we'll have ISIS cargo, setting off crate bombs all over the station. Combine one of these babies with a fuel tank and you've got some deadly shit on your hands.

But w/e. I'll take a look at the code tomorrow.

scrdest:
Good point, using the tank/crate as a combo is something I didn't think of. 

On the other hand, Allahu Klausbar Cargo is, unlike Guncargo on overt antags, is clearly in the realm of BWOINK. I'd be hard-pressed to accept '*we blew up half the station because we thought the EMT was a changeling*' as an excuse myself.

And if the cargo guy is the antag, it's not much worse than they could already do. I could remove the ex_act from it, especially since I just remembered some weird things a'la singlo called it on stuff, it was mostly there since I didn't know then what exactly emitters called on the things.

Glloyd:
Once you remove that ex_act as you mentioned, I'll merge this bitch.

Glloyd:
👍

PR #536, opened 2017-02-28
Dynamic window damage

moonheart08:
* Windows now take damage dynamically based on mob size and throw multiplier


code/game/objects/structures/window.dm
```diff
@@ -152,10 +152,9 @@
 	..()
 	visible_message("<span class='danger'>[src] was hit by [AM].</span>")
 	var/tforce = 0
-	if(ismob(AM))
-		tforce = 3
-		if(ishuman(AM))
-			tforce = 40
+	if(ismob(AM)) // All mobs have a hit multiplier and a size
+		var/mob/I = AM
+		tforce = I.mob_size * 2 * I.throw_multiplier
```
scrdest: null checks, null check-47s for everyone

code/modules/mob/mob_defines.dm
```diff
@@ -153,11 +153,12 @@
 	var/list/shouldnt_see = list()	//list of objects that this mob shouldn't see in the stat panel. this silliness is needed because of AI alt+click and cult blood runes
 
 	var/mob_size = MOB_MEDIUM
+	var/throw_multiplier = 1
```
scrdest: Miscommunication; this really really doesn't call for adding a new var at **/mob/ level** - it would affect *observers and AI eyes* which can't even be thrown, 99% of the time nobody will bother changing that, and it's kind of magical.

scrdest:
Uh. Considering you got it in on Bay, we're gonna have those changes in the next merge *anyway*.

PR #546, opened 2017-03-27
Random Venom Snakes

scrdest:
Adds a venom sac item. It's a pseudo-organ item; I'd have made it a proper organ but then it wouldn't play nice with being in simple_animals. It generates 10u amount of each random (with blacklist) chem on initialization, from one up to the specified number of chem types.

Refactors snakes' AttackingTarget to *actually get the mob's target rather than always null* thus making generic snakes' toxin work; also moves that toxin effect to a separate proc called in AttackingTarget, bite(), which handles bite effects.

Adds brown-colored snakes that use the above changes to spawn with random venom that is injected with each bite into a victim, provided it's a carbon-type. Robit armor stronk I guess.

However, since the venom is NOT used up, and can be nearly any chem, this has some abuse potential; nonetheless, so do the jungle fruits - probably moreso. Also, ATM the blacklist doesn't always trigger properly for the typesof-added elements for some reason.

### DIFF
--- a/code/modules/urist/mob/uristanimals.dm
+++ b/code/modules/urist/mob/uristanimals.dm
@@ -115,3 +115,47 @@ Please keep it tidy, by which I mean put comments describing the item before the
 	name = "Mule"
 	desc = "The QuarterMaster's turtle. Look out for bites!"
 
+/*holds a random chem for animal venom
+  should be used WITH copy=1 on transfer */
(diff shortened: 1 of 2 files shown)


Glloyd:
Nice branchname. It looks good though, I'll replace the snakes on Bear Island with these guys.

PR #548, opened 2017-03-30
Travis changes

Naxodile:
Makes travis check other maps again.


maps/glloydstation/glloydstation.dm
```diff
@@ -1,5 +1,8 @@
 #if !defined(USING_MAP_DATUM)
 
+	#include "../../code/modules/urist/area&turf/areas.dm"
+	#include "../../code/modules/urist/machinery/misc.dm"
```
Glloyd: put this file back where it was. Move the camera definition in that file to glloydstation_presets .dm

scrdest: Relative pathing is bad practice, better to just move these files wholesale to shared/urist or something along these lines

Glloyd:
yeah, having those relative paths there are iffy. You could just move the areas from that file to glloydstation_areas.dm and delete the old area file. That's what we should've just done anyways

Glloyd:
and now its runtiming from the awaymaps on the baymaps. Is it even worth it having those run through Travis?

PR #559, opened 2017-05-03
Misc fixes

Naxodile:
Fixes the welding mask having it's inhands be the same and fixes a case where you could make it permanently invisible.
Fixes curtains not starting with the plane of their closed version.
Fixes cloning so that clones aren't perfect and take longer.
Fixes plating on the syndicate shuttle, and some other windows that didn't have plating.
Fixes a spelling mistake in bone surgery.
Fixes assault mine boxes by changing them to box/large/mines
Adds the other cmo labcoat to medical.
Adds medical webbing to medical doctor lockers.
Adds a missing firelock for medical's front door, also slightly changes the area shape.
Ghosts can no longer knock on windows in non-spooky rounds.
Templates that used loot_spawner now use random/loot and maint/clean.


scrdest:
Looks alright for me.

code/game/machinery/cloning.dm
```diff
@@ -127,7 +127,7 @@
 	H.real_name = R.dna.real_name
 
 	//Get the clone body ready
-	H.setCloneLoss(H.maxHealth * (100 - config.health_threshold_crit) / 100) // We want to put them exactly at the crit level, so we deal this much clone damage
+	H.adjustCloneLoss(H.maxHealth * (100 - config.health_threshold_crit) / 100) // We want to put them exactly at the crit level, so we deal this much clone damage
```
scrdest: I hate whoever made those two methods and made the one call the other with a negative value.

code/modules/urist/modules/ghost_powers.dm
```diff
@@ -11,10 +11,11 @@
 
 /obj/structure/window/attack_ghost(mob/observer/ghost/user as mob)
 	. = ..()
-	playsound(src.loc, 'sound/effects/glassknock.ogg', 80, 1)
-	user.visible_message("Something knocks on the [src.name].",
-							"You knock on the [src.name].",
-							"You hear a knocking sound.")
+	if(round_is_spooky())
```
scrdest: TBH, rather than add those overrides, I probably should have just defined /atom/proc/spooky() and make attack_ghost() call that after this check, would save us the boilerplate. Would auto-examine when you wanted to spoop, but still.

Glloyd:
If @scrdest is satisfied, I'm comfortable merging, the rest looks fine to me.

PR #561, opened 2017-05-06
Locker Tweak

bknight286:
Lazy copypasta that makes personal lockers change names to reflect ownership.
Something is odd with this, I only changed one thing but some other stuff says it was changed, it claims I removed a line then put it one line down, twice.  I trust Glloyd can handle that.

### DIFF
--- a/data/mode.txt
+++ b/data/mode.txt
@@ -1,1 +1,1 @@
-secret
+extended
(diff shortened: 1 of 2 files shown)


data/mode.txt
```diff
@@ -1 +1 @@
-secret
+extended
```
Glloyd: When did this get changed back to secret? Odd.

bknight286: I have no idea what is going on with my PR right now to be honest but I think that's a text file that lists the local computer's current hosted game mode

Glloyd: Yeah, it is. It should be extended, no idea when it was changed to secret. It's all good.

Glloyd:
I'm fine with this 👍

PR #574, opened 2017-06-10
Planet + Assault additions

Glloyd:
Bugfixes for some planet stuff, planet outpost remap, toxins remap, assault balancing and planet assault mode (WIP).

Setting up the PR for Travis.


code/modules/urist/_maps/templates.dm
```diff
@@ -216,3 +215,7 @@ var/list/datum/map_template/underground_templates = list()
 
 	qdel(src)
 
+
+/obj/effect/template_loader/gamemode/assault
+	var/maptype = 0
+	gamemode = "assault"
```
scrdest: It might be better in the long run to make it into a macro, i.e. define MODE_ASSAULT 1 or whatever, for those if you want them to be dynamically loaded for each gamemode.

It'll be equally readable, prevent issues due to misspells, and will be significantly faster when you actually want to check the value of gamemode the loader holds, since string comparisons are slow - and if there's multiple loaders to iterate on, this WILL matter.

code/modules/urist/gamemodes/assault/assault.dm
```diff
@@ -10,6 +10,8 @@ var/global/remaininggens = 6
 	var/aliensurvivors = 0
 	votable = 0 //WEW
 	auto_recall_shuttle = 1
+	var/maptype = 0 //0 = station, 1 = planet
```
scrdest: Would macro this too, except here just for legibility.

PR #576, opened 2017-06-15
Small map improvements

rockwithagun:
adds a light to the research lobby, replaces the nuke in the vault with /obj/machinery/nuclearbomb/station, adds a glass table and an emergency toolbox to the cryogenic arrivals room so people can get out when there's no power, and replaces the iron wall in atmos with a rwall.


Glloyd:
Sure. We'll see how the nuke goes.

PR #584, opened 2017-07-19
Science outpost changes

rockwithagun:
Heavily expands xenobotany on the planet outpost, as well as reduces the amount of medium jungle turf, to prevent people (especially scientists) from getting mauled by panthers without traveling near the eastern river.


rockwithagun:
@Naxodile is this good for merging?

PR #590, opened 2018-06-24
Nerva map setup

Naxodile:
Converts the maps and swaps Z1 and Z3 for multi-z to function.
Sets up most of the important map datums.
The Glloydstation Travis fix is also here


maps/wyrm/wyrm_define.dm
```diff
@@ -81,7 +79,6 @@
 		"Aether Atmospherics",
 		"Zeng-Hu Pharmaceuticals",
 		"Hephaestus Industries",
-		"Commonwealth of Ahdomai"
```
Glloyd: What's this one?

Glloyd:
Thanks for this. I didn't realise that the order had to be like that for multi-z, but this is the most extensive work I've done with the new system, so I'm learning as I go.

maps/wyrm/wyrm_define.dm
Naxodile: Last I checked Tajara were pretty nuked.

PR #593, opened 2018-06-27
Bay Merge to July 10th, 2018

Glloyd:
As it says on the tin, @Naxodile, is there much left to do here?


Naxodile:
Conversion is done, and most everything I found is fixed. I do want to wait for the multi-tile airlock fix on Bay to get merged before this though.

Glloyd:
What are your feelings on the 3/4 sprites? I'd prefer to stick with the current ones for doors and walls, at least for now, but the other stuff I'm happy with.

Naxodile:
Walls look fine when they're not painted jet black like on the Torch. Airlocks are pretty bad for sides 4 & 8, but I'd rather have bad sprites that fit with everything else.

Glloyd:
That's fine with me, I haven't seen them outside the ones on the torch.

Glloyd:
What's the status on this? I'd like to get it in sooner rather than later so it's less stuff to convert on Nerva.

Glloyd:
Looks like the only thing left is those 2 travis issues with the gear datums. That should be an easy fix

PR #596, opened 2018-07-22
Several minor bug fixes and Wyrm updates.

Vakothu:
- Fixes the un-updated Wyrm fire alarms,
- Changes human sprite back to the Goon/Eris hybrid we were using,
- Adds CMO access to doctors on the Wyrm, so they can open certain orderable crates,
- Roboticists now spawn with the nifty roboticist lab-coat on both Wyrm and Glloydstation,
- Loadout roboticist lab-coat is no longer unable to be selected on Wyrm,
- Fake carp grenades now work properly,
- Fixes the mining voidsuit being invisible on-mob,
- Adds the ear-worn flowers to loadout,
- Icarus away now uses the same light colour = skybox colour that the Glloydstation planet uses.


Glloyd:
Very nice. I'll take a second look at the code in a bit, but the rest looks good.

Glloyd:
:+1:

PR #603, opened 2018-08-20
Scienceoutpostv2

rockwithagun:
baystation 12 - nss urist 2018-08-20 111202
Expands xenobotany from 4 trays to 12 trays, and reduces the chance of scientists getting mauled when leaving the outpost.


rockwithagun:
dreammaker_2018-08-20_11-21-14
dreammaker_2018-08-20_11-25-20
Old and new /turf/simulated/floor/planet/jungle/med compared.
Also issues in the pic have been fixed.

Glloyd:
You need to use the mapmerge tools in the tool folder before I can merge this. There's a readme in there, and I can give you more instruction if you need it. Also, you committed the backup, which is unnecessary, and it looks like you have some duplicate c-tags on the cameras.

rockwithagun:
I'm nearly certain I did use mapmerge but okay.

Naxodile:
You did in fact use mapmerge, however the map was changed from tgm back to dmm.

rockwithagun:
Yeah this ain't happening

Glloyd:
It would still be welcome if the mapmerge issues were fixed.

PR #610, opened 2018-10-01
Couple of fixes

Vakothu:
- Fixes the Trajan's airlock not working, and renames all the controller's parts to Trajan, not Calypso.
- Adds a smart-fridge to the kitchen and moves the shutter button under the light-switch.


Vakothu:
Make sure you change the config.txt to you being the host and what have you, I pushed the config I used, where I'm host.

PR #622, opened 2019-03-22
Assorted Teshari Updates

Vakothu:
Changelog:
  Gave Teshari innate dark vision to balance out their vision penalty in light.
  Added a Culture entry and a couple of Factions for Teshari to conform to Bay's new system for that.
    Addendum: This change fixed Teshari random names using the Human name generator.
  Fixed the descriptors defaulting to human sized for height and build on a Teshari.


Glloyd:
Looks good :+1:

PR #637, opened 2019-04-12
Makes the Bearcat NOT slow.

Andromeda-K22:
🆑 
Bearcat is now something resembling fast, so it doesn't take THREE MINUTES to move a single overmap tile.

:fuckbay:
🆑


Glloyd:
You're the saviour of @Vakothu

Glloyd:
I just realized it's 0.1 minutes and not seconds. Weird that it does that when every other ship uses seconds, but w/e.

PR #642, opened 2019-04-13
Medical balance and tweaks

Naxodile:
Makes it so arterial bleeding cannot occur in encased regions (head, chest) until the bones there are broken.
Re-balances lung recovery rates to be significantly faster.
Puts potassium chlorophoride behind carpotoxin as they both deal with stopping the heart.
Brings everything in the tactical medkit into line with standard naming conventions.
Gives bicaridine a buff to deal with cases of extreme damage (100~ to 200~) in a semi-timely fashion.


Andromeda-K22:
👍 

Seems all in order to me.
Can we maybe see about deobfuscating medical scanner results, or adding in actual damage readouts instead of generic 'this person is hurt a little/a bit/a lot in this part of the body'?

PR #645, opened 2019-04-15
Medical bugfixes

Naxodile:
Death now causes your heart to stop, just in case your body tried to live life on its own.
Human subspecies are now considered slightly less inferior to humans, and are acceptable for human-only prosthetics.
Attempting to fix decaying/necrotic organs with glue is now properly prevented.
Skrell and Unathi now have visible neural laces.


Glloyd:
I'll try to get to this tomorrow.

PR #663, opened 2019-04-28
[WIP] Torpedo IEDS

Andromeda-K22:
> THOMPSON: Twenty-five.
>SANBORN: Twenty-five meters, roger that. You are now in the kill zone.
>THOMPSON: Thanks for reminding me.
>SANBORN: That's what I'm here for, baby.

**Torpedo IEDS - Otherwise known as How I learned to Stop Worry and Love Cargonia**
I can't wait to see how much abuse these are going to get. 

Anyway.

This adds code to allow devices (timers, voice analyzers, signalers) to detonate torpedo warheads, and the code to modify them. 

They may currently be too powerful, i'l leave that up to others to decide.

Steps to rig a torpedo warhead to be an IED:
1. Crowbar the cover open.
2. Modify the internal circuitry using wirecutters.
3. Rig a wire to the detonator.
4. Disable the safeties with a multi-tool.
5. Attach a unsecured timer/signaller/voice analyzer to them.
You've got an IED now. It's armed.

Steps to *disarm* a torpedo warhead IED.
1. Open the cover.
2. Enable the safeties
3. Remove the device with wirecutters.
4. Snip the rigged wire with the wire cutters.
5. Undo the internal modifications with the wirecutters.
6. Crowbar the cover closed.

As usual, questions, comments, and constructive code criticism are welcome.


Andromeda-K22:
I'm hoping to have some sprites done for these, or rig up some of my own within a day or two.

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -282,6 +282,112 @@
...
+		visible_message("<span class='danger'>[src] beeps stubbornly, refusing to detonate!</span>")
+		playsound(src.loc, 'sound/machines/buzz-sigh.ogg', 25, 0, 10)
+	if(!safety)
+		if(prob(15)) // Small chance for the warhead's safeties to engage briefly.
+			visible_message("<span class='danger'>[src] beeps stubbornly, refusing to detonate!</span>")
+			playsound(src.loc, 'sound/machines/buzz-sigh.ogg', 25, 0, 10)
+			return
+		visible_message("<span class='danger'>[src] pings, begining a short countdown!</span>")
+		playsound(src.loc, 'sound/machines/ping.ogg', 25, 0, 10)
+		spawn(40) // 4 seconds to run away.
```
Naxodile: Ditch the comment, just use `spawn(4 SECONDS)`
This can also be applied through the PR to other do_afters/etc.

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -282,6 +282,112 @@
...
+	var/is_rigged = 0 //Has this thing been jerry-rigged?
+	var/riggedstate = 0 //What stage of rigging are we in?
+	var/obj/item/device/attached_device //Whatever is attached that'll detonate us.
+
+/obj/item/shipweapons/torpedo_warhead/examine(mob/user)
+	..(user)
+	switch(riggedstate)
+		if(1) to_chat(user, "<span class='notice'>It's control circuitry is exposed.</span>")
+		if(2) to_chat(user, "<span class='notice'>It's control circuitry is exposed, and the internal wiring appears to have been modified.</span>")
+		if(3) to_chat(user, "<span class='notice'>It's control circuitry is exposed, the internal wiring appears to have been modified, and a wire has been rigged to the detonator circuit.</span>")
```
Naxodile: ... and a wire has been rigged to \The [attached_device].

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -282,6 +282,112 @@
...
+	else if(istype(I, /obj/item/device/multitool))
+		if(!riggedstate)
+			return
+		if(!safety)
+			to_chat(user, "<span class='notice'>You begin to re-engage the built-in safeties.</span>")
+			if(do_after(user,40))
+				to_chat(user, "<span class='notice'>The warhead beeps softly, indicating you have re-enabled it's safeties!</span>")
+				safety = 1
+				playsound(src.loc, 'sound/machines/buttonbeep.ogg', 25, 0, 10)
+		if(safety)
```
Naxodile: All of this can become a single toggle with the following changes.
`safety = !safety`
and 
`to_chat(user, "<span class='notice'>You begin to [safety ? "disable" : "re-engage"] the built-in safeties.</span>")`

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -282,6 +282,112 @@
...
+			playsound(src.loc, 'sound/items/Crowbar.ogg', 50, 1)
+		else if(!riggedstate)
+			to_chat(user, "<span class='notice'>You carefully lever open the warhead's circuitry panel.</span>")
+			riggedstate++
+			playsound(src.loc, 'sound/items/Crowbar.ogg', 50, 1)
+		else
+			to_chat(user, "<span class='notice'>You can't close the panel. Remove the [attached_device] first.</span>")
+
+	else if(istype(I, /obj/item/weapon/wirecutters))
+		if(riggedstate == 3)
```
Naxodile: For the purposes of fun and reducing repetition I'm going to request you add a `/datum/wires` to this and replace the disarming step with that.

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -282,6 +282,119 @@
...
+			to_chat(user, "<span class='notice'>You carefully lever open the warhead's circuitry panel.</span>")
+			riggedstate++
+			playsound(src.loc, 'sound/items/Crowbar.ogg', 50, 1)
+			icon_state = "torpedowarhead-open"
+		else
+			to_chat(user, "<span class='notice'>You can't close the panel. Remove the [attached_device] first.</span>")
+
+	else if(istype(I, /obj/item/weapon/wirecutters))
+		if(is_rigged && attached_device)
+			to_chat(user, "<span class='notice'>You carefully begin to remove [attached_device] from the warhead's internals.</span>")
```
Naxodile: For the sake of fun and reducing repetition I'm going to suggest you combine this disarming stage and the one below to use a `/datum/wires` interface.

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -282,6 +282,119 @@
 	desc = "It's a big warhead for a big torpedo. Shove it in a torpedo casing and you've got yourself a torpedo." //torpedo
 	icon = 'icons/urist/items/ship_projectiles.dmi'
 	icon_state = "torpedowarhead"
+	var/safety = 1 //Integrated safeties are functioning as intended. This thing's inert.
+	var/is_rigged = 0 //Has this thing been jerry-rigged?
+	var/riggedstate = 0 //What stage of rigging are we in?
```
Naxodile: While not required you may want to use defines for these states.

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -282,6 +282,119 @@
...
+	var/obj/item/device/attached_device //Whatever is attached that'll detonate us.
+
+/obj/item/shipweapons/torpedo_warhead/examine(mob/user)
+	..(user)
```
Naxodile: The arguments are automatically passed by calling `..()`. You'll also need to check if `..()` returned false and not display anything in that case. Finally, you'll want to set the return value `.` equal to `..()` for the case some future type also overwrites this proc.

Andromeda-K22:
@Naxodile I'm currently not comfortable enough with wires and such to implement a wire-hacking system. In the future, it is a possibility.

code/modules/urist/modules/shipbattles/weapons.dm
Andromeda-K22: Not sure how to swing this and modify the icon state at the same time.

code/modules/urist/modules/shipbattles/weapons.dm
Andromeda-K22: Not comfortable enough with wiring to do so, unfortunately. In the future.

code/modules/urist/modules/shipbattles/weapons.dm
Andromeda-K22: Seems somewhat silly to use defines for these.

code/modules/urist/modules/shipbattles/weapons.dm
Naxodile: Icon_state is just a string, same as the rest.
`icon_state = "torpedowarhead-open-mod[safety ? "" : "-armed"]"`

code/modules/urist/modules/shipbattles/weapons.dm
Naxodile: Just makes it more readable by having `if(riggedstate == PANEL_OPEN)`,etc. You would only have them defined for this file alone.

code/modules/urist/modules/shipbattles/weapons.dm
Andromeda-K22: In the next commit.

code/modules/urist/modules/shipbattles/weapons.dm
Andromeda-K22: In the next commit.

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -282,6 +288,111 @@
...
+	if(!safety)
+		if(prob(15)) // Small chance for the warhead's safeties to engage briefly.
+			visible_message("<span class='danger'>[src] beeps stubbornly, refusing to detonate!</span>")
+			playsound(src.loc, 'sound/machines/buzz-sigh.ogg', 25, 0, 10)
+			return
+		visible_message("<span class='danger'>[src] pings, begining a short countdown!</span>")
+		playsound(src.loc, 'sound/machines/ping.ogg', 25, 0, 10)
+		playsound(src.loc, 'sound/items/countdown.ogg', 25, 0, 10)
+		spawn(4 SECONDS)
+			explosion(src, 4, 5, 6)
```
Naxodile: It'll cause a discrepancy compared to the real explosive effect of it, but devastation range is a lot bigger deal inside a ship compared to just hitting the hull. 
```suggestion
			explosion(src, 2, 4, 6)
```

Naxodile:
Beyond balance considerations the code here is fine.

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -282,6 +288,111 @@
...
+	if(!safety)
+		if(prob(15)) // Small chance for the warhead's safeties to engage briefly.
+			visible_message("<span class='danger'>[src] beeps stubbornly, refusing to detonate!</span>")
+			playsound(src.loc, 'sound/machines/buzz-sigh.ogg', 25, 0, 10)
+			return
+		visible_message("<span class='danger'>[src] pings, begining a short countdown!</span>")
+		playsound(src.loc, 'sound/machines/ping.ogg', 25, 0, 10)
+		playsound(src.loc, 'sound/items/countdown.ogg', 25, 0, 10)
+		spawn(4 SECONDS)
+			explosion(src, 2, 4, 6)
```
Glloyd: I would maybe make this 1, 3, 5. That's about the only change I can think of though. Otherwise it seems okay, balance-wise. We'll have to see how it plays out in-round though.

Naxodile:
I'll get around to making this use a wire system sometime this weekend.

PR #666, opened 2019-04-29
Makes cryogenic storage rad shielded

Glloyd:


Glloyd:
ah fuck, i ruined #666

PR #669, opened 2019-05-02
Improvised Rifle Fixes & Changes

Cheedows:
Removes the jamming chance from improvised rifles and improvised short rifles, as well fixes the bug where their inhand sprite wouldn't show up ingame. Now it looks like a space ak-47 as originally intended apparently.


Andromeda-K22:
I like these changes. Gives reason to build something other than fscking zip guns.

Glloyd:
We could probably use better in_hand sprites for this thing

PR #672, opened 2019-05-06
General Bugfixes Volume 2

Naxodile:
Portable atmospherics will no longer take power from APCs when their equipment is turned off.
Spilled welding and flamethrower fuel is no longer persistent.
Ghost writing now goes above the blood its made from.
The mysterious 'item' sometimes found in maintenance should no longer appear.
The ability to take certain languages is now checked on switching species.
Fixing suit breaches with tape now uses the damage's proper descriptor.
Xenophages and Promethians no longer have invisible chests/heads.
The circuit printer can make things that would take the exact amount it contains.
Runtime fixes for unused species.
Vines will now attempt to be somewhat aligned.
Vines will also now go down stairs. Going back up is still beyond them. This should hopefully fix issues with being stuck in vines above your zlevel.
The time it takes to get out of vines is now somewhat possible.
Plastic flaps on the Trajan will become airtight again if moved.
Short muskets will be loaded with something they can actually fire.
Gas that has overlay sprites will now appear over floor decals, and anything else that used to get in its way.
The Bearcat now has its intercomms direction properly set.
The Verne has had its areas changed to be remotely accurate to area usage standards. This fixes any issues with gravity in space and falling randomly.
Cards against the Galaxy has regained its sprites.


chaoko99:
If I might make a suggestion, regarding muskets: Please uncomment the code regarding one's ability to actually craft musket balls. I don't know why it's commented out, those guns blow now.

scrdest:
I would suggest breaking up these kinds of PRs into commits in the future. This would document what solution fixed which problem, and so when the merge or whatever inevitably fucks something up again in the same way, you could just look at the commit history and reapply the fix.

Naxodile:
Proper documentation? In my PRs? Impossible.

martinlyra:
For the future reference, it would be awesome to put the bugfixes and their info in a bullet list. Makes it more ordered and clean to read.

Added bonus if you can put extra effort in adding location of what files were changed for one specific bugfix. Not like it is important, but it helps.

Naxodile:
File location is included in all commits. Adding it in the description would be excessive.

code/modules/hydroponics/spreading/spreading_response.dm
```diff
@@ -61,7 +61,7 @@
 			unbuckle()
 		else
 			user.setClickCooldown(100)
-			var/breakouttime = rand(600, 1200) //1 to 2 minutes.
+			var/breakouttime = rand(30 SECONDS, 1 MINUTE)
```
Glloyd: blessed change

Glloyd: could even make it more reasonable tbh, 20 seconds to a minute. You end up standing there for so long.

code/modules/species/station/adherent.dm
```diff
@@ -49,7 +49,7 @@
 	slowdown = -1
 	hud_type = /datum/hud_data/adherent
 	pixel_offset_y = -4
-
+/*
```
Glloyd: thanks, I'd done this locally but hadn't pushed it yet.

code/modules/urist/structures/uriststructures.dm
```diff
@@ -448,7 +448,7 @@ Please keep it tidy, by which I mean put comments describing the item before the
 	buildstate = 4
 
 //special flaps for shuttles
-/obj/structure/plasticflaps/mining/special/clear_airtight()
+/obj/structure/plasticflaps/mining/special/become_airtight()
```
Glloyd: oof, i'm an idiot

PR #678, opened 2019-05-13
Protolathe / Circuit imprinter UI Overhaul

Andromeda-K22:
🆑 
Circuit Imprinters and Protolathe designs are now broken up into various categories (8 for the protolathe, 6 for the imprinter).
This works off defined categories, and each option will generate a list of designs which have that category. Anything with an unspecified category is found under the 'general' section of each machine.
Protolathe Menu:
image

image
🆑


Andromeda-K22:
For something so simple, this was a surprising amount of work.

PR #680, opened 2019-05-15
Circulator Griefproofing

Andromeda-K22:
🆑 
Circulators can now no longer be unbolted while pressurized. 
This is a griefproofing measure, largely.
🆑

### DIFF
--- a/code/modules/atmospherics/components/binary_devices/circulator.dm
+++ b/code/modules/atmospherics/components/binary_devices/circulator.dm
@@ -89,6 +89,11 @@
 
 /obj/machinery/atmospherics/binary/circulator/attackby(obj/item/weapon/W as obj, mob/user as mob)
 	if(isWrench(W))
+		var/air1pressure = air1.return_pressure()
+		var/air2pressure = air2.return_pressure()
+		if(air1pressure || air2pressure < ONE_ATMOSPHERE)
+			to_chat(user, "<span class='warning'>You can't unbolt the circulator while it's pressurized!</span>")
+			return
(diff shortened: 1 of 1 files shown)


code/modules/atmospherics/components/binary_devices/circulator.dm
```diff
@@ -89,6 +89,11 @@
 
 /obj/machinery/atmospherics/binary/circulator/attackby(obj/item/weapon/W as obj, mob/user as mob)
 	if(isWrench(W))
+		var/air1pressure = air1.return_pressure()
+		var/air2pressure = air2.return_pressure()
+		if(air1pressure || air2pressure < ONE_ATMOSPHERE)
```
Naxodile: Current reads as "if air1pressure is true (aka, exists) OR air2pressure is less than one atmosphere."
```suggestion
		if((air1pressure > ONE_ATMOSPHERE) || (air2pressure > ONE_ATMOSPHERE))
```

Naxodile:
A few other minor things. First is that compared to pipes two atmosphere maximum pressure this seems a little off, but that's not too critical. Second is that you don't have to define vars just for your if statements. You can call procs in the statement and they'll get resolved then evaluated, however, this is entirely up to what you think is more readable.

code/modules/atmospherics/components/binary_devices/circulator.dm
Andromeda-K22: Ah. I wasn't entirely sure on the use of ||.

Andromeda-K22:
It was purely down to readability, hence why I wrote it like that.

PR #681, opened 2019-05-15
HE pipes now check space area/turf instead of just turf

Andromeda-K22:
🆑 
Space radiators now check for being in the space area and turf, instead of just turf. They will still work if they are in the space turf but a different area.
🆑


Glloyd:
This won't change anything functionally. The exterior areas that are covering part of the radiator become defined as /area/maintenance/exterior on runtime. I'm extending the radiator in my upcoming mapping PR, but this change won't alleviate those issues.

chaoko99:
Check for AREA_FLAG_EXTERNAL

PR #633, opened 2019-04-08
Baymerge to March 11th, 2019

Naxodile:
Only a little late.
- [x] Medical needs the new equipment (auto-cpr, body scan displays, heating/cooling elements) added
- [x] There's now a functional BSA that acts as a ship weapon and is used for mining if you're insane.
- [x] The Trajan needs a full set of ship computers and machinery, this is already in the current commit
- [x] Re-add missing crafting recipes
- [ ] Re-add mentors
- [x] Fix missing locker appearence datums for first & second officer
- [x] Fix req_access uses in **every single map**
  - [x] ~~Fix req_access for Glloydstation~~ -Not a priority at the moment
- [x] Possibly revert the ammo caliber consolidation
- [x] Remove the delay before the pregame countdown
- [x] Initialization time has somehow doubled to near 400 seconds in my local tests
- [x] Find the source of the random crashes?
- [x] ~~Giant spiders aren't compatible with our hostile mob system.~~ 
  - They still produce some strange behavior, but it's at least functional
- [ ] Fix the admin body scan report runtime.
  - The cause of this is still unknown, there's not much debugging information from it
- [x] Add in the new ballistics sprites.
- [ ] Adjust the new Travis tests to our maps.
- [x] Repaths are needed for all our templates, particularly the NT Trading Station.

PR #682, opened 2019-05-15
Smarter airlock initialization

Andromeda-K22:
Last PR of the night.
🆑 
Airlocks now pull name and access from area if name is not defined on initialize.
Areas now have a req_access variable you can set, so you don't have to worry about setting that either, if it's configured in the area!
🆑

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -7,6 +7,7 @@
 	var/global/global_uid = 0
 	var/uid
 	var/area_flags
+	var/list/req_access = list()
 
 /area/New()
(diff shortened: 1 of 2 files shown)


Glloyd:
Correct me if I'm wrong @Naxodile, but won't this immediately be made redundant by #633?

Naxodile:
Yeah, the current setup in #633 uses a system where airlock access is determined by all the areas its touching. Also, I believe there's an autonaming door somewhere under airlocks already.

Andromeda-K22:
Alright, closing then.

PR #685, opened 2019-05-17
TEG Refactor

Andromeda-K22:
🆑 
The oil in Engineering's Roboco vendor now has a use. You can lubricate to prevent overloads.
Lubricating the TEGs increases thermal efficiency by 15%. 

Overloads now, functionally, are the same. However, they have some 'scary' aesthetic effects.
When a TEG overloads, it'll make a grumpy message and a noise. There's a chance it'll go a step farther, make a loud BANG, a warning buzzer, shake the camera of anyone in view, and make some smoke.

.... It's perfectly harmless, however!
🆑

### DIFF
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -144,7 +144,7 @@
 						playsound(src.loc, 'sound/effects/meteorimpact.ogg', 50, 1, -3)
 						for(var/mob/living/M in view(7, src))
 							shake_camera(M, 1, 2)
-						spawn(5)
+						spawn(0.5 SECONDS)
 							SM.start()
 							playsound(src.loc, 'sound/effects/smoke.ogg', 50, 1, -3)
 	//Power
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -40,7 +40,7 @@
 /obj/machinery/power/generator/examine(mob/user)
 	..()
 	to_chat(user, "Auxilary tank shows [reagents.total_volume]u of liquid in it.")
-	if(reagents.total_volume == 0)
+	if(!lubricated)
 		to_chat(user, "It seems to be in need of oiling.")
 
 //generators connect in dir and reverse_dir(dir) directions
@@ -84,7 +84,6 @@
 		return
 
 	updateDialog()
-
 	var/datum/gas_mixture/air1 = circ1.return_transfer_air()
 	var/datum/gas_mixture/air2 = circ2.return_transfer_air()
 
@@ -129,7 +128,7 @@
 	//remember to lubricate your engines kiddos
 	if(!lubricated)
 		if(effective_gen > max_power && prob(5))
-			var/datum/effect/effect/system/spark_spread/s = new /datum/effect/effect/system/spark_spread
+			var/datum/effect/effect/system/spark_spread/s = new()
 			s.set_up(2, 1, src)
 			s.start()
 			stored_energy *= 0.5
@@ -138,15 +137,15 @@
 				var/malfsound = pick(soundlist)
 				playsound(src.loc, malfsound, 50, 0, 10)
 				if(prob(20))
-					var/datum/effect/effect/system/smoke_spread/SS = new()
-					SS.set_up(5, 0, src.loc)
+					var/datum/effect/effect/system/smoke_spread/SM = new()
+					SM.set_up(5, 0, src.loc)
 					playsound(src.loc, 'sound/machines/warning-buzzer.ogg', 50, 1, -3)
 					spawn(2 SECONDS)
 						playsound(src.loc, 'sound/effects/meteorimpact.ogg', 50, 1, -3)
 						for(var/mob/living/M in view(7, src))
 							shake_camera(M, 1, 2)
 						spawn(5)
-							SS.start()
+							SM.start()
 							playsound(src.loc, 'sound/effects/smoke.ogg', 50, 1, -3)
 	//Power
 	last_circ1_gen = circ1.return_stored_energy()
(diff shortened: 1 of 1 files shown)


code/modules/power/generator.dm
```diff
@@ -113,12 +125,29 @@
...
+			var/datum/effect/effect/system/spark_spread/s = new /datum/effect/effect/system/spark_spread
+			s.set_up(2, 1, src)
+			s.start()
+			stored_energy *= 0.5
+			if(prob(55))
+				visible_message("<span class='danger'>[src] [pick(soundverb)]!</span>")
+				var/malfsound = pick(soundlist)
+				playsound(src.loc, malfsound, 50, 0, 10)
+				if(prob(20))
+					var/datum/effect/effect/system/smoke_spread/SS = new /datum/effect/effect/system/smoke_spread
```
scrdest: You can just do new() without specifying the type, it's implicit on assignment.

code/modules/power/generator.dm
```diff
@@ -113,12 +125,29 @@
...
-		var/datum/effect/effect/system/spark_spread/s = new /datum/effect/effect/system/spark_spread
-		s.set_up(2, 1, src)
-		s.start()
-		stored_energy *= 0.5
-
+	//exceeding max power is supposed to be bad, have there be a small chance of some visual and audio effects to stress this
+	//remember to lubricate your engines kiddos
+	if(!lubricated)
+		if(effective_gen > max_power && prob(5))
+			var/datum/effect/effect/system/spark_spread/s = new /datum/effect/effect/system/spark_spread
```
scrdest: That's overly verbose and inconsistent with the way you do it for smoke a couple lines below. Speaking of which - you have a /effect/system for sparks called `s`, and one for smoke called `SS` - this is gonna be confusing to maintain.

code/modules/power/generator.dm
```diff
@@ -135,6 +164,13 @@
 		lastgenlev = genlev
 		update_icon()
 	add_avail(effective_gen)
+	if(reagents.has_reagent(/datum/reagent/lube/oil))
+		reagents.remove_any(0.01)
```
scrdest: Subtle bug - you're not guaranteed to actually use up the _oil_ for lubrication if any other reagent winds up in machine.

code/modules/power/generator.dm
```diff
@@ -135,6 +164,13 @@
...
+		thermal_efficiency = 0.80
+		lubricated = 1
+	if(!reagents.has_reagent(/datum/reagent/lube/oil))
```
scrdest: `else` seems like it would make more sense for this.

code/modules/power/generator.dm
```diff
@@ -152,6 +188,15 @@
 		else
 			disconnect_from_network()
 		reconnect()
+	if(istype(W, /obj/item/weapon/reagent_containers))
+		var/obj/item/weapon/reagent_containers/R = W
+		if(!R.reagents.has_reagent(/datum/reagent/lube/oil))
+			audible_message("<span class='warning'>[src] blips in disappointment</span>")
+			playsound(get_turf(src), 'sound/machines/synth_no.ogg', 50, 0)
+			return
+		R.standard_pour_into(user, src)
```
scrdest: You're only checking if the container _has_ oil in it for validation. This means you can put a drop of oil in a bucket of human blood and dump the concoction in the engine just fine. Which is pretty metal. 

Coupled with the other check I commented on, you can lubricate the engine with literally anything from water, through adminordazine, to molten iron.

code/modules/power/generator.dm
```diff
@@ -22,15 +23,26 @@
...
+	create_reagents(120)
 	..()
 	desc = initial(desc) + " Rated for [round(max_power/1000)] kW."
 	spawn(1)
 		reconnect()
 
+/obj/machinery/power/generator/examine(mob/user)
+	..()
+	to_chat(user, "Auxilary tank shows [reagents.total_volume]u of liquid in it.")
+	if(reagents.total_volume == 0)
```
scrdest: `if(lubricated)` would probably be more representative.

code/modules/power/generator.dm
Andromeda-K22: There is a check on proc that handles adding the oil.

code/modules/power/generator.dm
Andromeda-K22: I can't see a decent way around this, although with the changes on the reagent check in process() i can just make the TEG dump any reagent that isn't oil.

Naxodile: The way to do it is to `reagents.remove_reagent(/datum/reagent/lube/oil, amount)`, however, it'd be pretty odd if somehow only the oil was coming out of something you were pouring.

code/modules/power/generator.dm
```diff
@@ -113,12 +124,29 @@
...
+				playsound(src.loc, malfsound, 50, 0, 10)
+				if(prob(20))
+					var/datum/effect/effect/system/smoke_spread/SM = new()
+					SM.set_up(5, 0, src.loc)
+					playsound(src.loc, 'sound/machines/warning-buzzer.ogg', 50, 1, -3)
+					spawn(2 SECONDS)
+						playsound(src.loc, 'sound/effects/meteorimpact.ogg', 50, 1, -3)
+						for(var/mob/living/M in view(7, src))
+							shake_camera(M, 1, 2)
+						spawn(5)
```
Naxodile: 0.5 SECONDS works just as well. I'm eternally averse to spawn after hearing about its performance, but given that this is nested enough to be classified as a bird's nest I'm sure it'll be fine.

Andromeda-K22: Where do you think the Teshari come from?

Glloyd:
@scrdest Are you happy with this? If so, I'll merge.

PR #686, opened 2019-05-19
Adds circuitry tools to roboticist belt storage

Racerbiscuit:
Added integrated circuitry tools to robotics belt's can_hold list.

### DIFF
--- a/code/modules/urist/items/clothes/belt.dm
+++ b/code/modules/urist/items/clothes/belt.dm
@@ -19,7 +19,11 @@
  	/obj/item/device/flashlight,
  	/obj/item/stack/cable_coil,
  	/obj/item/device/analyzer,
- 	/obj/item/device/mmi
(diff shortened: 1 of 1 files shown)


code/modules/urist/items/clothes/belt.dm
```diff
@@ -19,7 +19,11 @@
  	/obj/item/device/flashlight,
  	/obj/item/stack/cable_coil,
  	/obj/item/device/analyzer,
- 	/obj/item/device/mmi
+ 	/obj/item/device/mmi,
+	/obj/item/device/integrated_electronics/debugger,
```
Naxodile: Adding /obj/item/device/integrated_electronics instead of every subtype will work just fine.

Racerbiscuit: Got it, thanks.

PR #692, opened 2019-05-23
Adds pipenet collectors

Andromeda-K22:
🆑 
Adds pipenet enable collectors. Position them over a pipe cap, and wrench into place.
Uses phoron from the pipenet instead of a tank. Adminspawn only currently.
🆑

### DIFF
--- a/code/modules/power/singularity/collector.dm
+++ b/code/modules/power/singularity/collector.dm
@@ -173,7 +173,12 @@ var/global/list/rad_collectors = list()
 /obj/machinery/power/rad_collector/pipenet/Destroy()
 	QDEL_NULL(removed)
 	pipe = null
-	..()
+	. = ..()
+
+/obj/machinery/power/rad_collector/pipenet/Initialize()
+	if(anchored)
+		locate_pipenet()
+		connect_to_network()
 
 /obj/machinery/power/rad_collector/pipenet/Process()
 	//so that we don't zero out the meter if the SM is processed first.
--- a/code/modules/power/singularity/collector.dm
+++ b/code/modules/power/singularity/collector.dm
@@ -163,3 +163,101 @@ var/global/list/rad_collectors = list()
 		flick("ca_deactive", src)
 	update_icons()
 	return
+
+/obj/machinery/power/rad_collector/pipenet
+	name = "Modular Radiation Collector Array"
+	desc = "A variant of the radiation collector array which connects to a pipecap underneath it."
+	var/obj/machinery/atmospherics/pipe/cap/pipe = null
+	var/datum/gas_mixture/removed = null
+
+/obj/machinery/power/rad_collector/pipenet/New()
+	..()
+	rad_collectors += src
+	if(anchored)
+		locate_pipenet()
+
+/obj/machinery/power/rad_collector/pipenet/Destroy()
+	rad_collectors -= src
+	. = ..()
+
+/obj/machinery/power/rad_collector/pipenet/Process()
+	//so that we don't zero out the meter if the SM is processed first.
+	last_power = last_power_new
+	last_power_new = 0
+	if(!pipe || !anchored)
+		return
+	var/datum/gas_mixture/env = pipe.return_air()
+	removed = env.remove(0.75 * env.total_moles)
+	if(pipe && active)
+		var/rads = SSradiation.get_rads_at_turf(get_turf(src))
+		if(rads)
+			receive_pulse(rads * 5) //Maths is hard
+
+	if(pipe && removed)
+		if(removed.gas["phoron"] == 0)
+			investigate_log("<font color='red'>out of fuel</font>.","singulo")
+			active = 0
+			update_icons()
+		else
+			removed.adjust_gas("phoron", -0.001*drainratio)
+			env.merge(removed)
+	return
+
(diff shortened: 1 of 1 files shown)


Andromeda-K22:
Not my runtime. Re-run travis pls.

code/modules/power/singularity/collector.dm
```diff
@@ -163,3 +163,101 @@ var/global/list/rad_collectors = list()
...
+	var/obj/machinery/atmospherics/pipe/cap/pipe = null
+	var/datum/gas_mixture/removed = null
+
+/obj/machinery/power/rad_collector/pipenet/New()
+	..()
+	rad_collectors += src
+	if(anchored)
+		locate_pipenet()
+
+/obj/machinery/power/rad_collector/pipenet/Destroy()
```
Naxodile: Needs to handle destruction of the pipe reference and the gas mixture.

code/modules/power/singularity/collector.dm
```diff
@@ -163,3 +163,101 @@ var/global/list/rad_collectors = list()
...
+				src.locked = 0 //just in case it somehow gets locked
+				to_chat(user, "<span class='warning'>The controls can only be locked when the [src] is active</span>")
+		else
+			to_chat(user, "<span class='warning'>Access denied!</span>")
+		return 1
+	return ..()
+
+
+/obj/machinery/power/rad_collector/pipenet/proc/locate_pipenet()
+	var/T = src.loc
```
Naxodile:
```suggestion
	var/turf/T = get_turf(src)
```

code/modules/power/singularity/collector.dm
```diff
@@ -163,3 +163,101 @@ var/global/list/rad_collectors = list()
...
+			"You [anchored? "secure":"undo"] the external bolts.", \
+			"You hear a ratchet")
+		if(anchored)
+			connect_to_network()
+			locate_pipenet()
+		else
+			disconnect_from_network()
+			P = null
+		return 1
+	else if(istype(W, /obj/item/weapon/card/id)||istype(W, /obj/item/modular_computer))
```
Naxodile: You can just let `return ..()` handle this.

code/modules/power/singularity/collector.dm
```diff
@@ -163,3 +163,101 @@ var/global/list/rad_collectors = list()
...
+
+/obj/machinery/power/rad_collector/pipenet
+	name = "Modular Radiation Collector Array"
+	desc = "A variant of the radiation collector array which connects to a pipecap underneath it."
+	var/obj/machinery/atmospherics/pipe/cap/pipe = null
+	var/datum/gas_mixture/removed = null
+
+/obj/machinery/power/rad_collector/pipenet/New()
+	..()
+	rad_collectors += src
```
Naxodile: Left over from the parent proc.

code/modules/power/singularity/collector.dm
```diff
@@ -163,3 +163,101 @@ var/global/list/rad_collectors = list()
...
+	if(anchored)
+		locate_pipenet()
+
+/obj/machinery/power/rad_collector/pipenet/Destroy()
+	rad_collectors -= src
```
Naxodile: Same as above. It'll already be removed in the parent call.

code/modules/power/singularity/collector.dm
```diff
@@ -163,3 +163,85 @@ var/global/list/rad_collectors = list()
...
+/obj/machinery/power/rad_collector/pipenet
+	name = "Modular Radiation Collector Array"
+	desc = "A variant of the radiation collector array which connects to a pipecap underneath it."
+	var/obj/machinery/atmospherics/pipe/cap/pipe = null
+	var/datum/gas_mixture/removed = null
+
+/obj/machinery/power/rad_collector/pipenet/Destroy()
+	QDEL_NULL(removed)
+	pipe = null
+	..()
```
Naxodile: Missed this, all Destroy()'s need to return their hint.
```suggestion
	. = ..()
```

code/modules/power/singularity/collector.dm
```diff
@@ -163,3 +163,85 @@ var/global/list/rad_collectors = list()
 		flick("ca_deactive", src)
 	update_icons()
 	return
+
+/obj/machinery/power/rad_collector/pipenet
+	name = "Modular Radiation Collector Array"
+	desc = "A variant of the radiation collector array which connects to a pipecap underneath it."
+	var/obj/machinery/atmospherics/pipe/cap/pipe = null
+	var/datum/gas_mixture/removed = null
+
```
Naxodile: You went a little bit overkill and removed the connecting to a pipenet in New() part. Also you can just stuff it in Initialize instead.

code/modules/power/singularity/collector.dm
```diff
@@ -163,3 +163,90 @@ var/global/list/rad_collectors = list()
...
+/obj/machinery/power/rad_collector/pipenet/Destroy()
+	QDEL_NULL(removed)
+	pipe = null
+	. = ..()
+
+/obj/machinery/power/rad_collector/pipenet/Initialize()
+	if(anchored)
+		locate_pipenet()
+		connect_to_network()
+
```
Naxodile: It's that time again. All Initialize()'s must return their parent's init hint or a custom one.

PR #693, opened 2019-05-23
Adds Telepads

Andromeda-K22:
🆑 
More or less directly ports telepads from Virgo.
By request of Lance77am.
🆑


code/modules/telesci/quantumpad.dm
```diff
@@ -0,0 +1,197 @@
...
+	origin_tech = list(TECH_POWER = 6, TECH_ENGINEERING = 4, TECH_BLUESPACE = 4)
+	req_components = list(
+							/obj/item/weapon/stock_parts/subspace/crystal = 1,
+							/obj/item/weapon/stock_parts/capacitor = 2,
+							/obj/item/stack/cable_coil = 5,
+							/obj/item/weapon/stock_parts/console_screen = 1)
+
+/obj/machinery/power/quantumpad/New()
+	..()
+	component_parts = list()
```
Naxodile: Fun fact, you can make all this component mess into build_default_parts(/obj/item/weapon/circuitboard/telepad)

PR #695, opened 2019-05-26
[WIP][DNM] Charge Weapon framework

Andromeda-K22:
Adds the basics for charge weaponry, still a WIP.

One weapon included for testing.

Charge weapons work based on a SMES unit in a certain proximity of the weapon, their damage is based on that and some other factors (a power conversion factor, shield/hull damage ratios).

They use 20% of the linked capacitor's charge per shot, so balancing is likely required.

### DIFF
--- a/baystation12.dme
+++ b/baystation12.dme
@@ -2879,6 +2879,7 @@
 #include "code\modules\urist\modules\resourcetech\hunter.dm"
 #include "code\modules\urist\modules\resourcetech\planer.dm"
 #include "code\modules\urist\modules\resourcetech\tools.dm"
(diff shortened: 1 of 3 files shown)


Andromeda-K22:
+ Added a shield buster laser projectile, for use against supershielded targets.

Glloyd:
If you're taking a certain percentage of a SMES, you're going to have some issues. SMESs can have different max storages, so if someone builds a SMES there with a bigger storage capacity, it'll draw more than one with a smaller capacity. It should probably take a certain amount, rather than a percentage.

Andromeda-K22:
This was generally the intent, however I can probably clamp it down to a max charge draw for balance reasons.

PR #700, opened 2019-05-29
Fusion updates

Andromeda-K22:
🆑 
+ Gyrotrons are now constructable.
+ Circuits are sorted properly into the R&D categories
🆑


Glloyd:
700 :tada:

PR #701, opened 2019-05-30
Constructables: The Constructed Begining

Andromeda-K22:
🆑 Adds more constructables 

+ TEGS
+ Emitters
+ Radiation Collectors
+ Hydroponic Basins
+ Honey Extractors
🆑

### DIFF
--- a/baystation12.dme
+++ b/baystation12.dme
@@ -2830,6 +2830,8 @@
 #include "code\modules\urist\modules\donate.dm"
 #include "code\modules\urist\modules\ghost_powers.dm"
 #include "code\modules\urist\modules\mining.dm"
+#include "code\modules\urist\modules\constructable\constructable_circuits.dm"
+#include "code\modules\urist\modules\constructable\constructable_designs.dm"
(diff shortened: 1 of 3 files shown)


Naxodile:
You may want to add support for `default_deconstruction_screwdriver` and the rest of the machinery procs in some of the machines. As is they won't be deconstructable / be able to have their parts replaced.

Glloyd:
Travis error is related to your research designs not being unique. As an aside, I would love to see the results from your constructables survey, see if there's anything there I can toss in. Also, how the hell are hydroponic basins not already constructable.

Andromeda-K22:
@Naxodile I'l have to get around to adding functionality for those parts to these machines, although that's honestly probably best left to another PR on making the parts actually matter.

Glloyd:
The only real question here is one of balance, regarding the required tech levels. TEGs, rad collectors and circulators could probably be bumped up to power/engineering 3, although someone who is more familiar with RnD would be able to give better input than me.

PR #709, opened 2019-06-08
Improves QoL when fixing internal organs.

Naxodile:
The changes people probably want anyway.
Now instead of closing that incision you just spent the past 15 seconds making you'll think twice and consider your options. You can still cover it up on anything other than help intent.
The surgery itself is now also changed to work on an individual basis so that you won't scar someone's eyeballs to death when you're performing brain surgery.

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -86,7 +86,7 @@
 				if(used == amount)
 					break
 				if(W.is_surgical() && user.a_intent == I_HELP)
-					to_chat(user, "<span class='notice'>You refrain from closing the incision.</span>")
+					to_chat(user, "<span class='notice'>You refrain from closing the incision as your intent is set to help.</span>")
 					continue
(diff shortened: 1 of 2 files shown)


code/game/objects/items/stacks/medical.dm
```diff
@@ -85,6 +85,9 @@
 					continue
 				if(used == amount)
 					break
+				if(W.is_surgical() && user.a_intent == I_HELP)
+					to_chat(user, "<span class='notice'>You refrain from closing the incision.</span>")
```
Glloyd: I feel like this might work better as an input, might make things a little more clear for newer players. Either that or an addition to the message like "you refrain from closing the incision as your intent is set to help."

Glloyd: I feel like this might work better as an input, might make things a little more clear for newer players. Either that or an addition to the message like "you refrain from closing the incision as your intent is set to help."

PR #713, opened 2019-06-09
Big Huge Zombie Update + simplemob attack fix

scrdest:
Unfortunately the SA fixes and the zombie update got slightly tangled up together.

- Un-awfuls the hostile simplemob click/attack logic. Current implementation put attack effect handling literally _in the AI code_, so player-controlled and NPC mobs used separate logic for the same thing for no reason.
- Also un-awfuls simplemob priorities; they would attempt to perform two actions per tick, and so would magically be able to attack both the environment and the mob in one tick. We do not abide by such haxxorz, and as such they now first try to eat your face, and try to smash their way towards you only if the face proves temporarily unavailable.
- Restores Urist zombification procs; to prevent conflicts with upstream, I used the wonky uCamelCase (u for Urist) naming convention for procs. It's ugly, but Bay would have to actively try to make it conflict that way.
- Reworks plague zombies. Building on the first point, the special transformation verb for PCs is now removed; PC and NPC zombies automagically apply it by simply attacking the target.
- Furthermore, the transformation is now gradual and reagent-based. A bite transforms a fairly small random volume of a very evil reagent, which will try to zombify the victim. Upon a bite, the transformation is preventable, and small quantities may just be metabolized away, but since the reagent has a chance of multiplying, it requires either urgent advanced medical care, or more... drastic measures.

### DIFF
--- a/code/modules/urist/mob/spoopymobs.dm
+++ b/code/modules/urist/mob/spoopymobs.dm
@@ -24,7 +24,7 @@
 	environment_smash = 1
 	var/plague = 0 //whether biting dead people spawns new zombies
 	var/regen = 0 //if true, they won't stay down, but revive after a delay
-	var/regen_delay = 900 //delay for regen revive
+	var/regen_delay = 90 //delay for regen revive
 
 /datum/reagent/xenomicrobes/uristzombie
 	metabolism = REM
@@ -164,13 +164,15 @@
 /mob/living/simple_animal/hostile/urist/zombie/death()
 	. = ..()
 
+	src.transform = null
+	var/matrix/N = matrix()
+	N.Turn(90)
+	src.transform = N
+
 	if(regen)
-		to_chat(src, "You begin to regenerate. This will take about [regen_delay/600] minutes.")
-		spawn(regen_delay)
-			var/matrix/N = matrix()
-			N.Turn(270)
-			src.transform = N
-			rejuvenate()
+		to_chat(src, "You begin to regenerate. This will take about [regen_delay/60] minutes.")
+		to_chat(src, "If you ghost, you can re-enter the mob assuming it still exists.")
+		addtimer(CALLBACK(src, /mob/living/simple_animal/hostile/urist/zombie/proc/deathregen), regen_delay SECONDS, TIMER_STOPPABLE)
 
 	if(src.contents)
 		var/inv_size = contents.len
@@ -182,10 +184,11 @@
 		update_icons()
 
 /mob/living/simple_animal/hostile/urist/zombie/ghostize(var/can_reenter_corpse=1)
-	if(can_reenter_corpse)
-		if(src.client)
-			to_chat(src.client, "You can re-enter the mob as long as it still exists.")
-	. = ..(max(1, can_reenter_corpse)) // always re-enterable
+	return ..(max(1, can_reenter_corpse)) // always re-enterable
+
+/mob/living/simple_animal/hostile/urist/zombie/proc/deathregen()
+	src.transform = null
+	rejuvenate()
 	return
 
 /mob/living/simple_animal/hostile/urist/zombie/Destroy()
--- a/code/modules/urist/mob/spoopymobs.dm
+++ b/code/modules/urist/mob/spoopymobs.dm
@@ -21,13 +21,14 @@
 	move_to_delay = 7
 	stat_attack = 1
 	ranged = 0
+	environment_smash = 1
 	var/plague = 0 //whether biting dead people spawns new zombies
 	var/regen = 0 //if true, they won't stay down, but revive after a delay
 	var/regen_delay = 900 //delay for regen revive
 
-/datum/reagent/toxin/zombie/uristzombie
+/datum/reagent/xenomicrobes/uristzombie
 	metabolism = REM
-	target_organ = BP_BRAIN
+	var/target_organ = BP_BRAIN
 
 	var/symptom_msgs = list(
 		"Your teeth feel loose.",
@@ -56,10 +57,10 @@
 		"Your tongue swells up and turns black."
 	)
 
-/datum/reagent/toxin/zombie/uristzombie/affect_touch(var/mob/living/carbon/M, var/alien, var/removed)
+/datum/reagent/xenomicrobes/uristzombie/affect_touch(var/mob/living/carbon/M, var/alien, var/removed)
 	affect_blood(M, alien, removed * 0.5)
 
-/datum/reagent/toxin/zombie/uristzombie/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+/datum/reagent/xenomicrobes/uristzombie/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if (istype(M, /mob/living/carbon/human))
 		var/mob/living/carbon/human/H = M
 		var/true_dose = H.chem_doses[type] + volume
@@ -112,9 +113,9 @@
 							H.uZombify(1, 1, transformation_msgs)
 		else
 			// almost straight copy of plain reagent/toxin/affect_blood()
-			if(strength && alien != IS_DIONA)
-				M.add_chemical_effect(CE_TOXIN, strength)
-				var/dam = (1 + rand(5))
+			if(alien != IS_DIONA)
+				M.add_chemical_effect(CE_TOXIN, 5)
+				var/dam = 0.05 * rand(1, 20)
 				if(target_organ)
 					var/obj/item/organ/internal/I = H.internal_organs_by_name[target_organ]
 					if(I)
@@ -147,6 +148,10 @@
 		if(prob(Clamp(5+true_dose, 0, 20)))
 			H.make_jittery(5)
 
+/mob/living/simple_animal/hostile/urist/zombie/Aggro()
+	if(prob(35))
+		playsound(src.loc, pick('sound/hallucinations/wail.ogg', 'sound/hallucinations/screech.ogg', 'sound/hallucinations/growl1.ogg', 'sound/hallucinations/growl2.ogg', 'sound/hallucinations/growl3.ogg'))
+	..()
 
 /mob/living/simple_animal/hostile/urist/zombie/say()
 	var/acount = rand(2,8)
@@ -216,8 +221,8 @@
 
 			if(victim.reagents)
 				var/biosafety = victim.getarmor(null, "bio")
-				if(prob(Clamp(100-biosafety, 0, 100)))
-					victim.reagents.add_reagent(/datum/reagent/toxin/zombie/uristzombie, rand(5, 10))
+				if(!prob(Clamp(biosafety, 0, 100)))
+					victim.reagents.add_reagent(/datum/reagent/xenomicrobes/uristzombie, rand(5, 10))
 
 			uZombieInfect(victim)
 
@@ -332,7 +337,14 @@
 		else
 			drop_from_inventory(W)
 	new_mob.update_icons()
-	playsound(src.loc, 'sound/hallucinations/wail.ogg')
+
+	playsound(src.loc, pick(
+		'sound/hallucinations/wail.ogg',
+		'sound/hallucinations/screech.ogg',
+		'sound/hallucinations/growl1.ogg',
+		'sound/hallucinations/growl2.ogg',
+		'sound/hallucinations/growl3.ogg')
+	)
 
 	spawn()
 		qdel(src)
--- a/code/modules/urist/mob/spoopymobs.dm
+++ b/code/modules/urist/mob/spoopymobs.dm
@@ -26,7 +26,6 @@
 	var/regen_delay = 900 //delay for regen revive
 
 /datum/reagent/toxin/zombie/uristzombie
-	flags = AFFECTS_DEAD
 	metabolism = REM
 	target_organ = BP_BRAIN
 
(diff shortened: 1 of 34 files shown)


code/modules/urist/mob/spoopymobs.dm
```diff
@@ -21,11 +21,131 @@
...
+		"Greasy sand seems to fill your blood vessels.",
+		"Grey oil oozes out of your ears.",
+		"You smell spoiling meat. You drool at the thought.",
+		"Your tongue swells up and turns black."
+	)
+
+/datum/reagent/toxin/zombie/uristzombie/affect_touch(var/mob/living/carbon/M, var/alien, var/removed)
+	affect_blood(M, alien, removed * 0.5)
+
+/datum/reagent/toxin/zombie/uristzombie/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
```
Andromeda-K22: There's no way to delay or counteract this, dialysis isn't always avaliable - add dylovene as a delaying agent or peridaxon as a antidote?

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -81,91 +199,104 @@
...
-	else if(istype(T, /mob/living/simple_animal/hostile/scom/civ))
-		var/mob/living/simple_animal/hostile/scom/civ/victim = T
-		src.visible_message("<span class = 'warning'> <b>[src]</b> chomps at [victim]'s brain!</span>", "<span class = 'warning'>You munch on [victim]'s brain!</span>")
-		victim.Zombify(src.regen, src.plague, src.maxHealth)
 	return
 
-/*
-/mob/living/carbon/human/proc/Zombify(var/regens = 0, var/infects = 0, var/hitpoints = 40) //I swear officer, that Animalize() proc fell out the back of a truck.
+// do excuse the weird naming convention, but Bay *already* broke this proc once so I'm not taking chances.
+/mob/living/simple_animal/hostile/urist/zombie/proc/uZombieInfect(var/mob/living/infectee)
```
Andromeda-K22: This seems to ignore clothing like biosuits and sealed suits. That seems like a bad idea, since i doubt a decaying human's teeth can puncture a spaceproof suit enough to exchange a significant amount of fluids.

scrdest: If it did, it... well, did. Keep in mind, it's not your regular 'corpse except alive' zombie, this is explicitly regenerative - so much so that if it transforms the brain, the only solution is to pulverize the whole body. It just uses a lot of energy and didn't quite figure out higher brain functions yet, hence BRAAAAAINS.

code/modules/urist/mob/spoopymobs.dm
scrdest: It's not supposed to be fair, it's supposed to be _fucking terrifying_. Rabies was a major inspiration - once you show symptoms, its recovery rate is single-digit, with brain damage, after damn near replacing the person's blood with antivirals, and we don't even know those weren't strokes of luck.

Peridaxon and Alkysine do act as delaying agents, and Dylovene helps the patient just not plain die on you regardless of transformation. Failing that, you can apply 10ccs of circular saw to the head - this kind of decision is a genre staple and I was explicitly designing around encouraging a scenario where you have to make it.

Naxodile: As this is a subtype of reagent/toxin I'm almost certain dylovene is already an antidote for it. I'll check and see how effective this really is later.

scrdest: I checked the code, Dylovene removes all /reagent/toxin subtypes at a linear ratio of 1:4. On the other hand, the zombie chem exhibits pseudo-exponential growth (it's actually ridiculously complicated to find the expected value mathematically - it's a stack of two autoregressive models).

What this all amounts to is that a detox has a chance to work, especially if applied early and/or on a single bite or two. It still may work otherwise, but then it's basically Spess Milwaukee Protocol for Spess Zombie Rabies. And that balance makes me very happy.

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -81,91 +199,104 @@
...
+		var/mob/living/carbon/human/victim = infectee
+		if(victim)
+			src.visible_message(munch_msg_ext, munch_msg_self)
+			victim.uZombifyInstant(src.regen, src.plague, src.maxHealth)
+
+	else if(istype(infectee, /mob/living/simple_animal/hostile/scom/civ))
+		var/mob/living/simple_animal/hostile/scom/civ/victim = infectee
+		if(victim)
+			src.visible_message(munch_msg_ext, munch_msg_self)
+			victim.uZombify(src.regen, src.plague, src.maxHealth)
```
Naxodile: Scom civvies could never die or be transformed from what I saw in testing.

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -21,11 +21,131 @@
 	move_to_delay = 7
 	stat_attack = 1
 	ranged = 0
-	universal_speak = 1
 	var/plague = 0 //whether biting dead people spawns new zombies
 	var/regen = 0 //if true, they won't stay down, but revive after a delay
 	var/regen_delay = 900 //delay for regen revive
```
Naxodile: Couldn't find a good place to put this. Zombies do not break tables, barricades or anything else in their way, and as such I have to rate them 0/10.

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -21,11 +21,131 @@
...
 
+/datum/reagent/toxin/zombie/uristzombie
+	flags = AFFECTS_DEAD
```
Naxodile: This flag doesn't appear to function, transformations never progressed after death.

Glloyd:
That Travis crash on Glloydstation is something I've never seen before, what the hell

Glloyd:
There's nothing special about that specific turf, and it's crashing during init. I wonder if this is related to the server crashing somehow?

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -21,11 +21,131 @@
...
+					var/obj/item/organ/external/Head = H.organs_by_name[BP_HEAD]
+					if (Head)
+						if(!(Head.status & ORGAN_DEAD))
+							Head.status |= ORGAN_DEAD
+							for (var/obj/item/organ/external/C in Head.children)
+								C.status |= ORGAN_DEAD
+						else
+							H.uZombify(1, 1, transformation_msgs)
+		else
+			// straight copy of plain reagent/toxin/affect_blood()
```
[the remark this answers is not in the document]
scrdest: Considering I _literally personally copy-pasted it_ from that proc, I don't see how that's possible, unless there's a communication issue at play - only this particular bit is copied over; the code below in the same else block is custom. Perhaps I should add an endblock comment to clarify.

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -21,11 +21,131 @@
...
+			if(strength && alien != IS_DIONA)
+				M.add_chemical_effect(CE_TOXIN, strength)
+				var/dam = (1 + rand(5))
```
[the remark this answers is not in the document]
scrdest: Again, this is plain toxin code, the only real difference is a different metabolic rate which may be a culprit. 

Also - what race? I've tested it a LOT of times and I've seen it working fine for Skrell, Unathi and Resomi, although it's obviously trickier in the latter two due to their health handling being different.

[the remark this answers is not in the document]
Naxodile: [https://github.com/UristMcStation/UristMcStation/blob/master/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm#L18](url)
The extremely important difference between normal toxin code and this is that you aren't accounting for the metabolic rate, and are instead doing up to 75% of cyanide's effect at half the reagent use.
All of the testing done was on humans. By a single bite I notably mean getting bitten once then walking away.

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -21,11 +21,131 @@
...
+							"Your [E.name] feels stiff.",
+							"You smell something foul.",
+							"Flesh on your [E.name] turns black.",
+							"You can't feel the skin on your [E.name].",
+							"You can feel blood clotting in your [E.name].")
+						E.status |= ORGAN_DEAD
+						to_chat(H, "<span class='warning'>[decay_msg]</span>")
+
+						for (var/obj/item/organ/external/C in E.children)
+							C.status |= ORGAN_DEAD
```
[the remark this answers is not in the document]
scrdest: At this point, this is literally just for show - specifically, to apply the necrotic limb visuals; the mob gets deleted afterwards after the newly spawned zombbo steals their icon.

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -21,11 +21,131 @@
...
+			to_chat(H, "<span class='warning'>[pick(symptom_msgs)]</span>")
+
+		if(prob(true_dose))
+			H.reagents.add_reagent(src.type, rand(10*removed))
+
+		H.add_chemical_effect(CE_PAINKILLER, 40)
+
+		// transforming sanics the victim's metabolism:
+		H.add_chemical_effect(CE_PULSE, 4)
+		H.nutrition -= min(H.nutrition, true_dose)
```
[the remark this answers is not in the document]
scrdest: I was thinking about adding a clamp, but I've seen Bay do it as well - I presumed life(), or whatever, clamps that value, so it's not worth the overhead.

code/modules/urist/mob/spoopymobs.dm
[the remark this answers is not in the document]
scrdest: I think that's something I forgot to revert, good catch. Odd that we've got such different outcomes in human tests, though.

code/modules/urist/mob/spoopymobs.dm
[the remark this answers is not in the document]
Naxodile: Turns out it sure does.

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -81,91 +199,104 @@
...
-	set desc = "Eat a target corpse's brain to zombify him."
-
-	if(stat || !(plague))
-		src << "You cannot eat brains in your current state."
-		return
+		if(ishuman(A))
+			var/mob/living/carbon/human/victim = A
+
+			if(victim.reagents)
+				victim.reagents.add_reagent(/datum/reagent/toxin/zombie/uristzombie, rand(1, 15))
```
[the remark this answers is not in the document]
scrdest: Fair enough, I'll look into that.

code/modules/urist/mob/spoopymobs.dm
scrdest: That's not really endemic to this mob, nothing does AFAIK. Hostiles' code is pretty slapdash in general, although, the random fixes I've put in here already addressed some Big Issues it had.

code/modules/urist/mob/spoopymobs.dm
scrdest: They do, but not because of this flag; I think brainmed broke this flag -_-

code/modules/urist/mob/spoopymobs.dm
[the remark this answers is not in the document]
Naxodile: While that's true of other cases where this is used, this one happens quite early and survivors have a relatively high chance of having this happen to them.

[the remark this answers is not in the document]
scrdest: Oh, my bad, I thought you were referring to a different loop. Yeah, good catch.

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -21,11 +21,131 @@
...
+		if(prob(max(0, idlemsg_proba)))
+			to_chat(H, "<span class='warning'>[pick(symptom_msgs)]</span>")
+
+		if(prob(max(0, min(100, true_dose))))
+			H.reagents.add_reagent(src.type, rand(10*removed))
+
+		H.add_chemical_effect(CE_PAINKILLER, 40)
+
+		// transforming sanics the victim's metabolism:
+		H.add_chemical_effect(CE_PULSE, 4)
```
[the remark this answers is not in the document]
scrdest: Highest _effective_ elevation, i.e. doesn't give you more bang for your buck in terms of pulse rate.

This does not mean 4 does not make any difference from 3. The clamping of the pulse rate occurs _after_ all the modifiers are taken into account, which means things that would normally bring you down to, say, lvl 2 pulse just don't work.

code/modules/urist/mob/spoopymobs.dm
```diff
@@ -36,24 +156,15 @@
...
-		M.Turn(90)
-		M.Translate(1,-6)
-		src.transform = M
-
-		src << "You begin to regenerate. This will take about [regen_delay/600] minutes."
+	if(regen)
+		to_chat(src, "You begin to regenerate. This will take about [regen_delay/600] minutes.")
 		spawn(regen_delay)
 			var/matrix/N = matrix()
+			N.Turn(270)
```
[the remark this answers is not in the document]
scrdest: Gah. It used to work fine originally, but it got broken during the last merge. I changed it to 270 because before it was the opposite - they would be alive, but display lying down.

code/modules/urist/mob/spoopymobs.dm
Glloyd: This is something I'd like to see as well, although if hostiles aren't breaking tables or barricades, then that's something that should be made into an issue

Glloyd:
:+1: On the non-zombie changes, they look good to me. I didn't see any issues in the zombie code besides what @Naxodile has already brought up, so if those couple issues are addressed, then I'd be fine to merge this, but I haven't tested it as extensively as Nax has. Although if you could sneak in a fix for hostiles not breaking structures, that would certainly be a welcome addition.

code/modules/urist/mob/spoopymobs.dm
scrdest: Turns out Dylovene had some special toxin-subtype reagent-clearing logic attached. It's no longer a subtype of that; I used Xenomicrobes since it makes some sense and has a chance of being blacklisted properly that way.

code/modules/urist/mob/spoopymobs.dm
scrdest: FINALLY done. Argh.

Glloyd:
there is a conflict in code/game/objects/structures.dm that needs to be resolved before this can be merged

code/modules/urist/mob/spoopymobs.dm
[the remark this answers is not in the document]
scrdest: _Burned to the goddamn ground, **finally**._

scrdest:
Conflict should be resolved now, rebased on the latest master.

PR #720, opened 2019-06-13
AIs can no longer open canisters.

Andromeda-K22:
This is more of a grief magnet than anything else.


Glloyd:
Yeah, I'm not sure how anyone thought this was a good idea

PR #722, opened 2019-06-13
Fixes issue with MMI holder objects

martinlyra:
Title. Beware, this is a fix I've made after a long sleep-deprived night. While it is my responsibility to test the code and that it does indeed compile, I need help to test for validation of its intended function.

* Have a FBP player, and a non-synthetic player in the same session. The goal is to ash the FBP's head and transfer their mind into the MMI that eventually lands on the ground.
* A conscious and alive MMI should be able to talk. Or if the FBP becomes a ghost, they should be able to transfer back into the MMI by double-clicking on it.
* The best way to achieve this is using an ion rifle directly on FBP
* Additionally, check if ashing a laced player's head will still drop the lace.

I'll conduct my own testing when I have the time.

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -885,11 +885,10 @@ Note that amputating the affected organ does in fact remove the infection from t
 			new /obj/effect/decal/cleanable/ash(get_turf(victim))
 
 			// URIST EDIT BY IRRA, 2019-05-25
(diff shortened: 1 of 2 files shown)


martinlyra:
Just to reiterate a memo that I posted in #development channel few days ago. This bugfix has been finished and is now available for final review.

PR #723, opened 2019-06-13
Fixes the likely cause of the crashes in initialization

### LINKED ISSUE #703: Server crashes at initialization stage
#### Description of issue
Server crashes at the initialization stage, often leaving behind no logs or additional information around the event. There are two different flavours;
* Crashes at map init - narrowed down to cave and ore generation
* Crashes soon as a gamemode vote had started

#### Difference between expected and actual behavior
No crashes

#### Steps to reproduce 
The crashes are difficult to reproduce in the normal environment since they occur irregularly (fancy way to say random). Forcing the server to spawn more than two exoplanets reduces this irregularity and gives more instances of crashing at map generation. Disabling random exo-planets altogether dramatically reduces the crashes to zero.

#### Specific information for locating
The crashes occur without any other additional information. To recite the description;
* Map generation
* After gamemode voting's start

#### Length of time in which bug has been known to occur
Unknown length, but it is widely known by developers and players.

#### Client version, Server revision & Game ID
Only server revision applicable: e904f40589d89d4cbb562b2affccf38a374060d2

#### Issue bingo
- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/UristMcStation/UristMcStation/issues)

Naxodile:
What in the actual fuck was I doing back in that baymerge.

This doesn't even seem to fix the issue, it just lowers it from an illegal instruction to a runtime. Which also causes the wall to have a null sprite with a hidden wall overlay from the case I saw this happen.
Fixes #703 ?

### DIFF
--- a/code/game/turfs/simulated/wall_icon.dm
+++ b/code/game/turfs/simulated/wall_icon.dm
@@ -54,12 +54,13 @@
 		return
 
 	for(var/i = 1 to 4)
-		if(other_connections[i] != "0")
-			I = image('icons/turf/wall_masks.dmi', "[material.icon_base]_other[wall_connections[i]]", dir = 1<<(i-1))
-		else
-			I = image('icons/turf/wall_masks.dmi', "[material.icon_base][wall_connections[i]]", dir = 1<<(i-1))
+		I = image('icons/turf/wall_masks.dmi', "[material.icon_base][wall_connections[i]]", dir = 1<<(i-1))
 		I.color = base_color
 		overlays += I
+		if(other_connections[i] != "0")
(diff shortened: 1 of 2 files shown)


scrdest:
Still crashing in the `.so`s on Travis + Glloydstation.

Naxodile:
Posted Bay's encounter with this in #development, looks like it'll become more reliable once we get the merge completed.

PR #725, opened 2019-06-14
Holobarrier port from TG.

chaoko99:
🆑 
Adds atmos holobarrier. It blocks atmosphere.
Adds security and engineering holobarriers. Walk through them.
/ 🆑 

NONE of these are mapped yet.


chaoko99:
Don't know why this is failing.

Glloyd:
Because this more or less negates the point of inflatables, and you haven't mapped any, I think this would be better served as an RnD unlock. That way inflatables still have a point, and this can be viewed as a direct upgrade, and be able to be produced

Glloyd:
Because this more or less negates the point of inflatables, and you haven't mapped any, I think this would be better served as an RnD unlock. That way inflatables still have a point, and this can be viewed as a direct upgrade, and be able to be produced

chaoko99:
I agree! Unfortunately, on that fact alone they will *never* be used because nobody knows how to research, or is actually just that lazy. I was mostly waiting for input on that.
I would *like* to map one in because using inflatables is such a genuinely awful experience the moment you have to pull a dense object like, say, any atmos portable.

PR #727, opened 2019-06-14
Gas turbine is now mapped behind atmos. Buffed, and made compatible with subsystems.

chaoko99:
🆑 
The gas turbine is now aft of atmospherics! Find the optimal mix!
The gas turbine has been buffed to produce roughly 2.8x the previous amount of power, it used to cap out at roughly 50kw
/ 🆑


chaoko99:
Also completely reorders the file to me less painful to read.

chaoko99:
Need to change the SMES subtype.

PR #730, opened 2019-06-15
Removes SMGs from the armory.

Andromeda-K22:
image


Naxodile:
2 years later and the consensus has changed to a nerf. #1013

PR #733, opened 2019-06-16
Makes the plasma cutter useful, and actually perform like it seems to suggest it would.

chaoko99:
🆑 
Plasma cutter does brute damage now.
Plasma cutter can cause bleeding.
Plasma cutter damage nerfed, added AP to the projectile.
Plasma cutter now mines every rock it hits.
/ 🆑

### DIFF
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -198,7 +198,7 @@
 	armor_penetration = 10
 	sharp = 1
 	edge = 1
-	damage_type = BRUTE || BURN //It's basically a particle accelerator anyways.
+	damage_type = BURN
 	kill_count = 5
 	pass_flags = PASS_FLAG_TABLE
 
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/projectile/beams.dm
```diff
@@ -194,11 +194,11 @@
...
 	icon_state = "omnilaser"
 	fire_sound = 'sound/weapons/plasma_cutter.ogg'
-	damage = 15
+	damage = 10
+	armor_penetration = 10
 	sharp = 1
 	edge = 1
-	damage_type = BURN
-	check_armour = "laser"
+	damage_type = BRUTE || BURN //It's basically a particle accelerator anyways.
```
Naxodile, line 201: Almost certain this does not get handled properly.
```suggestion
	damage_type = BURN
```

Naxodile (approved):
Seems fine, that check_armour was a repeat anyway

PR #734, opened 2019-06-16
Allow Numbers in Names

LorenLuke:
You know, for vat-growns and stuff.


Vakothu:
Just letting you know, Vat Growns and machines can already have numbers in their names,

LorenLuke:
I've run into issues

PR #735, opened 2019-06-16
Adds Visitor to Assistant Alt-Titles

LorenLuke:

### DIFF
--- a/maps/nerva/datums/nerva_jobs.dm
+++ b/maps/nerva/datums/nerva_jobs.dm
@@ -225,7 +225,7 @@
 
 /datum/job/assistant
 	alt_titles = list(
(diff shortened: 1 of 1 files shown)


Lyeos:
I'm not really sure Visitor meshes well with the setting (But will defer to Glloyd's judgement). Passenger might work better, but, eh.

Naxodile (approved):
Seems fine to me.

Naxodile:
Noticed this is still open, any thoughts on it, Glloyd?

Glloyd:
I think I'll side with Raz here. It'd need to be implemented as a separate role probably to avoid them getting paid. We'll put this to rest for now, but it's an interesting idea.

PR #737, opened 2019-06-17
Nerfs all alcohol strengths by a factor of 5, so your booze doesn't just kill your dudes.

chaoko99:
🆑 
Bar dwellers and drunkards rejoice, alcohol takes WAY more to just kill you outright. Exactly 5 times as much.
/ 🆑 

Holy fuck booze code is weird.

Edit: I'd like to change the ordering of those strength values, but whatever.


Naxodile:
Issue is more so in the effective dose calc, and this is a non-ideal way of fixing it. I'll toss this along with vomiting on my todo eventually list.

PR #739, opened 2019-06-17
Dylovene counters sleeping chems.

chaoko99:
🆑 
Did you chloral someone? Fix it with dylovene. Works with Soporific too.
/ 🆑 

who broke this


chaoko99:
Oh, "who broke this" can be traced back to whoever repathed those chemicals to not be south of toxin.

Naxodile (approved):
This will almost certainly never be used, but is good to go. Also why does the web editor have to exist

Glloyd:
yeah, web editor isn't great

PR #742, opened 2019-06-17
adds a better sprite for, and maps the Reagent Sublimator into secure tech.

chaoko99:
🆑 
Reagent sublimator doesn't have a crappy codersprite anymore. Thanks /TG/.
Reagent Sublimator is now mapped into secure tech storage, and will require the use of a CE or hacking to obtain. Have fun with the possibilities!
/ 🆑 
Now we have at least one thing in Secure Tech that will ever be used!


Naxodile:
Never even seen this thing used in play testing. Pinging @Vakothu and @Dimasw99 for engi-nerd balance thoughts.

Glloyd:
I will also defer to the engi-nerds

chaoko99:
Balance isn't much of an issue as chemicals already modify based upon how
much is being consumed. Back when this PR was made it was found to be
relatively sane for emergent gameplay purposes. It did lead to the concern
about alcohol, but I seem to remember that they didn't cause any issue of
note.. Did I change how alcohol poisoning works again? I can't remember
it's been so long.

Also what's up with the sudden burst of activity? My email has been blowing
up with this all day lmao

chaoko99:
bruh

PR #750, opened 2019-06-19
Disables Unathi Regen.

chaoko99:
🆑 
Unathi Regeneration ability is disabled now.
/ 🆑 
Inb4 those people who only play Unathi for the regen crawling out of the woodwork. Just watch.

### DIFF
--- a/code/modules/species/station/lizard.dm
+++ b/code/modules/species/station/lizard.dm
@@ -68,11 +68,11 @@
 		"Your scales bristle against the cold."
 		)
 	breathing_sound = 'sound/voice/lizard.ogg'
-
+/*Urist Edit: Disables Unathi Regen.
 	base_auras = list(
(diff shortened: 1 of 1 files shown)


RainbowSwaggins:
add 4 lines of code and become a Villain, commit out 4 lines and become a hero.

code/modules/species/station/lizard.dm
```diff
@@ -72,11 +72,11 @@
 	base_auras = list(
 		/obj/aura/regenerating/human/unathi
 		)
-
+/*Urist Edit: Removes Unathi Regen.
```
scrdest: No it does not. You've just _disabled the toggle_. 

Meaning that if the regen started out enabled, there's no way to **dis**able it without having staff do it manually.

chaoko99: Crap okay I think I figured it out.

PR #753, opened 2019-06-21
Bear Mirrors

Mazianni:
Title is misleading, these are actually beam mirrors but i mis-spelled it and through it was hilarious.

Adds beam mirrors. They'll redirect any beam they're hit with to a set direction.

Pretty basic.


code/modules/urist/machinery/beam_mirror.dm
```diff
@@ -0,0 +1,41 @@
...
+	desc = "A specially designed mirror that redirects photonic energy into a singular direction."
+	icon = 'icons/urist/beam_mirrors.dmi'
+	icon_state = "reflector_base"
+	density = 1
+	dir = NORTH
+	var/degrees_from_north = 0 //needed for vectors
+	var/compass_directions = list("North", "South", "East", "West")
+
+/obj/machinery/beam_mirror/bullet_act(var/obj/item/projectile/Proj)
+	if(istype(Proj, /obj/item/projectile/beam/))
```
Glloyd: Just an idea, but maybe non-beam projectiles should break the mirror?

code/modules/urist/machinery/beam_mirror.dm
```diff
@@ -0,0 +1,41 @@
...
+
+/obj/machinery/beam_mirror/verb/rotate_clockwise()
+	set category = "Object"
+	set name = "Rotate Mirror"
+	set src in view(1)
+
+	if (usr.stat || usr.restrained() || anchored)
+		return
+	var/choice = input("What point do you want to set \the [src] to?", "[name]") as null|anything in compass_directions
+	switch(choice)
```
chaoko99: This can be completely removed via `dir2angle` and `turn` calls, as well as making this rotate clockwise verb into just that. It's a bit simpler for the player overall.

code/modules/urist/machinery/beam_mirror.dm
```diff
@@ -0,0 +1,41 @@
...
+		Proj.redirect(new_x, new_y, curloc, null)
+	else
+		return
+
+/obj/machinery/beam_mirror/verb/rotate_clockwise()
+	set category = "Object"
+	set name = "Rotate Mirror"
+	set src in view(1)
+
+	if (usr.stat || usr.restrained() || anchored)
```
chaoko99: The anchored part of this could cause issues. It's also not checking if the player is anchored, but the object itself. There's also no way to anchor it.

PR #755, opened 2019-06-22
Disables clerical borgs. We aren't baystation.

chaoko99:
🆑 
Clerical borgs are disabled.
/ 🆑 

Now here's a weird one! We're not baystation. We aren't paperwork memes. Please don't just join as a borg to waste a slot.


Glloyd:
No. If people want to play as clerical borgs, they are more than welcome to.

PR #757, opened 2019-06-23
Thruster Additions

Andromeda-K22:
...

This was harder than I considered.

Adds gas-efficient plasma thrusters. They use a LOT  of power. 
Adds reactionless thrusters. Which were the first attempt at making plasma thrusters.

### DIFF
--- a/code/modules/overmap/ships/engines/ion_drive.dm
+++ b/code/modules/overmap/ships/engines/ion_drive.dm
@@ -1,0 +1,72 @@
+//Gas nozzle engine
+/datum/ship_engine/plasma_thruster
+	name = "plasma thruster"
+	var/obj/machinery/atmospherics/unary/engine/plasma/nozzle
+
+//Actual thermal nozzle engine object
+
+/obj/machinery/atmospherics/unary/engine/plasma
+	name = "plasma nozzle"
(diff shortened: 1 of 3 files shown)


code/modules/overmap/ships/engines/ion_drive.dm
```diff
@@ -0,0 +1,72 @@
...
+	use_power = 1
+	idle_power_usage = 150		//internal circuitry, friction losses and stuff
+	power_rating = 80000			//7500 W ~ 10 HP
+	opacity = 1
+	density = 1
+	moles_per_burn = 1.5
+
+/obj/machinery/atmospherics/unary/engine/plasma/Initialize()
+	. = ..()
+	controller = new(src)
```
Naxodile: Already created in the parent call.

code/modules/overmap/ships/engines/ion_drive.dm
```diff
@@ -0,0 +1,72 @@
...
+
+/obj/machinery/atmospherics/unary/engine/plasma/Destroy()
+	QDEL_NULL(controller)
```
Naxodile: Same deal as above. Already handled.

Naxodile:
These are implemented on multiple codebases / upstream (I think) now, will add stealing their updated version to my eternal todo list.

PR #758, opened 2019-06-23
Rubbers and Flashrounds: Coming to a Nerva Near You

Cheedows:
Adds flashround and rubber variants for 5.56 mm, 7.62 mm and 10mm. Flashrounds use the default flash round all current pistol flashrounds use, 5.56 and 7.62 rubber variants use a new rubber rifle bullet that does +2 damage and +10 agony over regular pistol rubbers, so 7 damage and 40 agony.  10mm uses the default pistol rubber round. All these variants are added to the autolathe, only visible when hacked like the majority of pre-existing ballistics. Additionally, changes the HI2521 SMG (the white one in the armory) to spawn with 9mm rubber rounds by default, design rationale is the same as why the 9mm Sabre SMG from cargo starts with rubbers. Regular magazines can obviously still be slotted in from the ammunition locker. 

You might ask why this PR was made. Regular rounds are still indisputably better than rubber and flash rounds, it just adds more variation for LTL options for automatic weapons which previously didn't exist before. And to be honest it's really funny to fire 3 round bursts of flash rounds.

### DIFF
--- a/baystation12.dme
+++ b/baystation12.dme
@@ -2769,6 +2769,7 @@
 #include "code\modules\urist\gamemodes\vampire\vampire_powers.dm"
 #include "code\modules\urist\items\ai_modules.dm"
 #include "code\modules\urist\items\airlock_painter.dm"
+#include "code\modules\urist\items\ammo.dm"
 #include "code\modules\urist\items\cards.dm"
 #include "code\modules\urist\items\chairpainter.dm"
 #include "code\modules\urist\items\circuitboards.dm"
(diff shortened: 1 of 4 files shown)


Glloyd:
I would prefer if the new items were moved into a urist file. Preferably a new one, I'm going to do a big cleanup of urist/items/guns.dm as part of the merge, but calibres and everything are getting fucked with by bay anyways, so doing it this way will save some time with both the cleanup and the inevitable merge conflicts this will have.

Cheedows:
Alright thanks for taking a look, I'll try to figure that out in about a week once I get home for vacation, unfortunately until then I won't be able to do it. If it'll delay the merge for too long then feel free to just scrap the PR.

PR #760, opened 2019-06-27
Allows the protolathe to be filled easily

Lmandude:
Now you can just pour stuff in with beakers and whatnot.


Lmandude:
Traaaaaaaviiiiiiis
I tried looking at it, but I don’t think it’s something that would be related to this?

PR #766, opened 2019-06-28
Allows warheads to be removed from torpedoes

Lmandude:
The glue stuck onto each and every warhead, both present and future, has finally been scraped off. Now you can remove them from their case prisons.

### DIFF
--- a/code/modules/urist/modules/shipbattles/weapons.dm
+++ b/code/modules/urist/modules/shipbattles/weapons.dm
@@ -260,7 +260,7 @@
 		if(istype(user,/mob/living/silicon))
 			return
 		user.put_in_hands(warhead)
-		user << "<span class='notice'>You remove the torpedo warhead.</span>"
+		to_chat(user, "<span class='notice'>You remove the torpedo warhead.</span>")
 		warhead = null
 		loaded = 0
 		icon_state = "bigtorpedo-unloaded"
--- a/code/modules/urist/modules/shipbattles/weapons.dm
+++ b/code/modules/urist/modules/shipbattles/weapons.dm
@@ -247,6 +247,7 @@
 	icon = 'icons/urist/items/ship_projectiles48x48.dmi'
 	icon_state = "bigtorpedo-unloaded"
 	var/loaded = 0
+	var/obj/item/shipweapons/torpedo_warhead/warhead = null
 	matter = list(DEFAULT_WALL_MATERIAL = 2500)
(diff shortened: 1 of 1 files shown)


Andromeda-K22:
how am i going to get high off glue now >:(

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -247,22 +247,32 @@
...
+/obj/structure/shipammo/torpedo/attack_hand(mob/user as mob)
+	if(warhead)
+		if(istype(user,/mob/living/silicon))
+			return
+		user.put_in_hands(warhead)
+		to_chat(user, "<span class='notice'>You remove the torpedo warhead.</span>")
+		warhead = null
+		loaded = 0
+		icon_state = "bigtorpedo-unloaded"
+
```
Naxodile:
```suggestion
	. = ..()
```
In the strange off chance ship ammo ever needs to handle something.

Lmandude:
I was gonna go with the magically-appearing-warhead option at first, but then I realized if someone put in a warhead that was modified in any way, if the warhead was removed everything would be mysteriously gone. I’ll see what I can do about that and the other stuff, though. Thanks for reminding me the preloaded ones exist.

Lmandude:
Changes made, in case github hasn’t alerted you yet. Sure hasn’t for me.

code/modules/urist/modules/shipbattles/weapons.dm
```diff
@@ -277,6 +285,7 @@
 	loaded = 1
 	icon = 'icons/urist/items/ship_projectiles48x48.dmi'
 	icon_state = "bigtorpedo"
+	warhead = /obj/item/shipweapons/torpedo_warhead
```
Glloyd: This will probably runtime when someone tries to remove the warhead. You will likely need to set this as `warhead = new /obj/item/shipweapons/torpedo_warhead` in the New() of the loaded torpedo.

Lmandude: Worked fine when I tested it, plus it’s how stunbatons do it if memory serves, which is what I looked at for an example.

Lmandude: Right, I’m dumb and misunderstood what you said, but still already handled. New() already has something for it, which is what I was talking about with the stunbaton example.

PR #767, opened 2019-07-01
Ports Polaris Blob [WIP but playable]

spacemen-on-a-space-station:
🆑
add: Ports Polaris Blob
/🆑

Mostly a straight port of polaris blob by @Neerti with some minor edits to get it working here plus some quality of life for player blobs.

This means blobs can now produce factories/nodes/spores/shields/resources and come in a variety of types with special abilities and properties of their own. These blobs can be both AI and player controlled.

Most major changes from Polaris blob are that this blob can destroy any machinery or structure and that player blobs don't get extraneous verbs that don't work with their type.

**Known issues**:

Polling for player blobs doesn't respect any sort of antag preference

Ranged spore blob type is commented out mostly because I have not gotten around to finishing it yet

Fungus blob type is commented out because it was not compiling and I haven't gone back and figured out why yet

A repeated runtime related to girders (or walls?) getting eaten

## ERROR: proc name: PrivUpdateListenerLoc (/datum/sound_token/proc/PrivUpdateListenerLoc)
  source file: sound_player.dm,229
  usr: null
  src: /datum/sound_token (/datum/sound_token)
  call stack:
/datum/sound_token (/datum/sound_token): PrivUpdateListenerLoc(the girder (/obj/structure/girder), 1)
/datum/sound_token (/datum/sound_token): PrivAddListener(the girder (/obj/structure/girder))
/datum/proximity_trigger/squar... (/datum/proximity_trigger/square): on turf entered(the plating (129,101,2) (/turf/simulated/floor/plating), the girder (/obj/structure/girder), null)
Entered (/decl/observ/entered): raise event(the plating (129,101,2) (/turf/simulated/floor/plating), the girder (/obj/structure/girder), null)

Some machinery that is destroyed fails to garbage collect

I expect people to find bugs because I'm not familiar with this codebase and what objects might be interacting with the blob but from my testing it is playable.

Requested by @Andromeda-K22


Naxodile:
This was commissioned but never ¿finished? Weird stuff aside, @Glloyd your thoughts on this conceptually, if one of us has to finish it?

PR #768, opened 2019-07-04
Adds some light scaling to rad collectors

Naxodile:
Also acts as a subtle buff to low radiation production.

### DIFF
--- a/code/modules/power/singularity/collector.dm
+++ b/code/modules/power/singularity/collector.dm
@@ -33,7 +33,7 @@ var/global/list/rad_collectors = list()
 	if(P && active)
 		var/rads = SSradiation.get_rads_at_turf(get_turf(src))
 		if(rads)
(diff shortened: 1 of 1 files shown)


code/modules/power/singularity/collector.dm
```diff
@@ -33,7 +33,7 @@ var/global/list/rad_collectors = list()
 	if(P && active)
 		var/rads = SSradiation.get_rads_at_turf(get_turf(src))
 		if(rads)
-			receive_pulse(rads * 5) //Maths is hard
+			receive_pulse(rads * (1+4*(250/rads))) //Maths is hard
```
scrdest: This collapses to just `pulse = rads + 1000` if we ignore floating-point precision issues.

Naxodile:
None of the revisions here were what I was after. I'll remake this later once I have the time.

PR #769, opened 2019-07-05
Introducing the Bluespace RPED

Lmandude:
Thanks to our wonderful scientists, we have manged to reverse-engineer the designs for a Bluespace RPED from one of our sister stations. Please disregard any rumors that said designs were actually stolen by our spies, which we certainly do not have.

Basically this ports the BRPED from tg. It can hold a lot more parts, and upgrade machines from a distance without needing the panel open. It even works through cameras, because bluespace. Also, credit to Nyxodile for helping make it a bit cleaner, hopefully.


Lmandude:
Icon shoved over in tgitems.dmi

code/modules/research/designs.dm
```diff
@@ -230,6 +230,15 @@ other types of metals and chemistry for reagents).
 	build_path = /obj/item/weapon/storage/part_replacer
 	sort_string = "CBAAA"
 
+/datum/design/item/stock_part/BRPED
```
Glloyd: i should have picked up on this the first time, but if possible, this should be moved to code/modules/urist/datums/uristdesigns.dm to save further merging headaches. Besides that, code looks good.

PR #772, opened 2019-07-06
Buffs medical pouches

Naxodile:
Trauma pouches can heal trauma, burn pouches can fix missing blood, and toxin pouches might even save you from cirrhosis.


Glloyd:
@Naxodile Travis error while compiling: 

`code/game/objects/items/weapons/storage/med_pouch.dm:76:error: /obj/item/weapon/reagent_containers/pill/pouch_pill/deletrathol: undefined type path`

PR #775, opened 2019-07-08
Moves clothing cutting to a right-click verb.

Lmandude:
Rejoice as you no longer need to worry about accidentally cutting your clothes to ribbons with a syringe or screwdriver when you just wanted to put it in your webbing.

### DIFF
--- a/code/modules/urist/items/clothstuff.dm
+++ b/code/modules/urist/items/clothstuff.dm
@@ -23,16 +23,30 @@
 	req_amount = 1
 	time = 10
 
-/obj/item/clothing/under/attackby(obj/item/I, mob/user)
+/obj/item/clothing/under/verb/cut()
+	set name = "Cut up clothing"
+	set desc = "Cut some clothes into cloth."
+	set category = "Object"
+	set src in oview(1)
+
+	if (!can_touch(usr))
+		return
+
+	var/obj/item/I = usr.get_active_hand()
+	if(!I)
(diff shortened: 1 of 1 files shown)


code/modules/urist/items/clothstuff.dm
```diff
@@ -23,16 +23,30 @@
...
 	if(is_sharp(I))
-		user.visible_message("<span class='notice'>\The [user] begins cutting up \the [src] with \a [I].</span>", "<span class='notice'>You begin cutting up \the [src] with \the [I].</span>")
-		if(do_after(user, 50, src))
-			to_chat(user, "<span class='notice'>You cut \the [src] into pieces!</span>")
+		usr.visible_message("<span class='notice'>\The [usr] begins cutting up \the [src] with \a [I].</span>", "<span class='notice'>You begin cutting up \the [src] with \the [I].</span>")
+		if(do_after(usr, 50, src))
+			to_chat(usr, "<span class='notice'>You cut \the [src] into pieces!</span>")
 			for(var/i in 2 to rand(3,5))
 				new /obj/item/stack/material/cloth(get_turf(src))
 			qdel(src)
```
Glloyd: @Lmandude i should have thought of it, but the clothing/under should drop all its contents here to prevent storage objects from disappearing into thin air. would you be able to hack that up quick in a new PR?

Lmandude: I’ll see if I can either tonight before I sleep or tomorrow, yeah.

PR #776, opened 2019-07-08
Removes Zombie Toxins, Nanites, and Hell Water from exotic plants.

Vakothu:
As the title says, this just removes nanites, hell water, and both variations of zombie toxin from Exotic Plants.


Naxodile:
Travis failure is just due to the issue with exoplanet ruins being preloaded on Glloydstation.

PR #781, opened 2019-07-12
Adds/Adjusts loadout computers

LorenLuke:
Adds tablet computers to the loadout, slightly adjusts prices for the laptops.

### DIFF
--- a/code/modules/client/preference_setup/loadout/lists/utility.dm
+++ b/code/modules/client/preference_setup/loadout/lists/utility.dm
@@ -64,12 +64,31 @@
 modular computers
 ****************/
 
+/datum/gear/utility/cheaptablet
+	display_name = "tablet computer, cheap"
+	path = /obj/item/modular_computer/tablet/preset/custom_loadout/cheap
+	cost = 3
+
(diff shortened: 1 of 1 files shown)


code/modules/client/preference_setup/loadout/lists/utility.dm
```diff
@@ -64,12 +64,31 @@
...
+	cost = 4
+
+/datum/gear/utility/customtablet/New()
+	..()
+	gear_tweaks += new /datum/gear_tweak/tablet()
+
 /datum/gear/utility/cheaplaptop
 	display_name = "laptop computer, cheap"
 	path = /obj/item/modular_computer/laptop/preset/custom_loadout/cheap
-	cost = 5
```
scrdest: With the cost being equal, why would you go for laptop vs tablet, or vice versa?

LorenLuke: Portability. Size. Pocketing. Convenience. I never got laptops, but laptops can run certain programs that Tablets can't.

PR #784, opened 2019-07-17
Makes it so that you uncuff yourself instead of unbuckling.

LorenLuke:


Naxodile:
Pretty significant re-balance, arguably warranting player vote. Pinging @Glloyd for the nth time today on a verdict.

PR #792, opened 2019-08-14
Make Radioactivity mutation less stupid & empower Ryetalyn

martinlyra:
Previously, the radioactivity mutation worked directly on the `radiation` variable of mobs. Instead of using the `rad_act` function. This has been addressed.

Secondly, Ryetalyn couldn't fix all mutations, specially those that were granted from `active_genes` - a major source of 'unfixable' mutations that cannot be cured or nullifed by Ryetalyn. A part of this pull request proposes fixing this by making Ryetalyn more potent in deactivating genes.

----
#### Changes to ` code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm`
* Ryetalyn now deactivates the state of DNA blocks responsible for activating mutations and other powers. Successfully nullifying mutations and resetting genes back to their normal state. 
#### Changes to `code/modules/urist/modules/genetics/disabilities.dm`
* Radioactivity mutation now respects armour and affects more than just mobs. (using `atom` instead of `mob`)

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -409,7 +409,6 @@
 	overdose = REAGENTS_OVERDOSE
 
 /datum/reagent/ryetalyn/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
-	M.mutations = list()
 	M.disabilities = 0
 	M.sdisabilities = 0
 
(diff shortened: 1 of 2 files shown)


code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
```diff
@@ -409,16 +409,23 @@
 	overdose = REAGENTS_OVERDOSE
 
 /datum/reagent/ryetalyn/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
-	var/needs_update = M.mutations.len > 0
-
 	M.mutations = list()
```
Naxodile, line 412:
```suggestion

```
Already handled in the code below.

Naxodile (approved):
The shift from living to atom doesn't actually change anything gameplay-wise currently, but is fairly inconsequential. Genes and mutations could technically be split into separate checks, I'm not going to block for it however.

PR #800, opened 2019-08-16
Fix tgitems.dmi with a working copy

martinlyra:
On requrest by @Glloyd


Glloyd:
this would be #800 

@chaoko99 you somehow corrupted tgitems.dmi

PR #818, opened 2019-09-21
Overmap ship modules

Andromeda-K22:
Adds:
Teleporters
Repair Modules
Shield Disruptors
Cloaking

Redux of a previous PR before I got sidelined with real life and such.

### DIFF
--- a/code/modules/urist/modules/shipbattles/overmapships/components.dm
+++ b/code/modules/urist/modules/shipbattles/overmapships/components.dm
@@ -6,6 +6,7 @@
 	var/mob/living/simple_animal/hostile/overmapship/mastership = null
 	var/broken = FALSE
 	var/targeted = TRUE
+	var/last_activation = null
 
 /datum/shipcomponents/proc/BlowUp()
 //	qdel(src)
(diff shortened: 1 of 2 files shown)


code/modules/urist/modules/shipbattles/overmapships/components.dm
```diff
@@ -160,4 +167,242 @@
...
+	var/list/boarding_mobs = list(/mob/living/simple_animal/hostile/russian, /mob/living/simple_animal/hostile/russian/ranged) //What mobs do we spawn on boarding?
+	var/boarding_number = 4 //We'll spawn this many every time we board.
+	var/last_boarding = 0 //How long since we boarded last?
+	var/boarding_turf = null//Where do we spawn our mobs?
+	var/boarding_failure_chance = 0 //more shields = more chance of preventing teleportation.
+
+/datum/shipcomponents/teleporter/DoActivate()
+	if(last_boarding > world.time) //Are we too early?
+		return
+
```
Glloyd: I haven't read through all of this yet, but I think you should add a check here for if the mastership is being boarded already

something like `if(mastership?.boarded) return`

Glloyd:
@Andromeda-K22 Been super busy lately, but I'll give this a full final lookthrough tomorrow or saturday if you fix that compile warning.

Glloyd:
incorporated in #886

PR #828, opened 2019-12-06
Changes Grass Turfs to be more natural and less eye-melting.

ShibberinoShipperino:
* Grass no longer destroys your retinas.

- Images

grassnew


Glloyd:
Always hated the old grass sprite

PR #845, opened 2020-02-12
Fixes Weather Opacity + Wooden Grille Exploit

Laniakeaaaa:
### What does this do?

- Mouse_Opacity appeared to be set wrong for weather effects, now you can interact, pull and inspect your character while being snowed, rained, sanded on.

- Fixes missing icon_state sprites for Wooden Grilles.

**Good News:**
This probably makes weather for Exoplanets viable now, right?


scrdest:
<3

Likely still not viable though, optimizing it to the point where it's efficient on large scale is a pretty advanced technical problem.

PR #856, opened 2020-03-14
🤖🤖🤖🤖🤖🤖🤖 MALF AI IS GONE! 🤖🤖🤖🤖🤖🤖🤖

Laniakeaaaa:
image

### What this does:

- Malf AI is no longer votable at the start, or as an antagonist.
- Becoming a traitor as an AI still makes you malfunction, all this does is remove the obvious round start rush, and obsolete voting shit.

Looking at vampire as well, since it's one of the most complained gamemodes, but depends on how well this will go through, Vamp is very design flawed.


Naxodile (approved):
Looks good to me. Malf has been covered under traitor for a while now.

PR #859, opened 2020-03-23
Adds alert level lighting(?)

Torque4607:
Aaaaaaa it worked when I tested it on my private instance, I have no idea how it's going to react. Bear with me, literally my first 'project' in DM.

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -41,6 +41,8 @@
 #define APC_UPOVERLAY_LOCKED 8
 #define APC_UPOVERLAY_OPERATING 16
 
+GLOBAL_LIST_INIT(global_apc_list,list())
+
 // Various APC types
 /obj/machinery/power/apc/critical
 	is_critical = 1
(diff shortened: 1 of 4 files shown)


code/modules/power/apc.dm
```diff
@@ -1287,6 +1289,21 @@ obj/machinery/power/apc/proc/autoset(var/cur_state, var/on)
 	update_icon()
 	return 1
 
+	/obj/machinery/power/apc/proc/set_light_color(var/color) //just changes the colors of the ship's lighting.
```
martinlyra: You should remove the spacing/indentation before the function definition here.
```suggestion
/obj/machinery/power/apc/proc/set_light_color(var/color) //just changes the colors of the ship's lighting.
```

code/modules/power/apc.dm
```diff
@@ -1287,6 +1289,21 @@ obj/machinery/power/apc/proc/autoset(var/cur_state, var/on)
 	update_icon()
 	return 1
 
+/obj/machinery/power/apc/proc/set_light_color(var/color) //just changes the colors of the ship's lighting.
```
Naxodile, line 1292: Personally, I would have this proc take the `/decl/security_level` itself as an argument. Then, checking for nulls to reset, have `lightmode` just set to `security_level_lightmode`.

Glloyd:
@Torque4607 Any movement on this?

Andromeda-K22:
I should note that Torque *probably* isn't aware of how to make too many technical changes, as far as I know.

PR #873, opened 2020-05-18
nervafixes

miniusAreas:
Basic mapping fixes to Virology and the Bridge.

- Adds atmospherics to the Test Monkey room to prevent atmospheric alarms.

- Adds a button to the Sensor Shroud. One never existed this entire time....


Glloyd:
The huge diff isn't great, but it's not a dealbreaker.

PR #890, opened 2020-07-12
Fixes Wizard Centcomm Exploit

Laniakeaaaa:
### What does this do?

Prevents Wizards from using teleport hijinks to reach out of bounds in Centcomm, leads them into a very ouchie surprise if they try to.

Thanks to Dimas for reporting this :)


Dimasw99:
I might be able to cheese that too

PR #921, opened 2020-12-12
Disables Extended in Secret/Random Votes.

Laniakeaaaa:
**Disables the ability to get Extended from a Secret/Random vote.**

As per discussions.
Looks like a simple config change, which was a nice change.


Glloyd:
Made the change serverside so I don't have to mess around with merging the config file. Thanks!

PR #925, opened 2020-12-18
Fixes windows blocking air always

Naxodile:
Breaking full tile windows would cause them to be pushed out onto space, but continue to block air. This would then cause ZAS to never trigger an update for the newly exposed room.


Glloyd:
thanks Nax

PR #929, opened 2020-12-21
Rebalances oxyloss from lung damage

Naxodile:
Oxyloss is properly capped as a percentage of the lung's total damage. This significantly changes the outcomes for ruptured lungs and other low damage cases to be more weakening than killing. Ruptures specifically will only reduce oxygenation to 75% excluding outside factors. Given sufficient procrastination the low chance to trigger losebreath will eventually kill someone.

The warning messages that come along with this now make the distinction between mild and severe asphyxia.


Glloyd:
This is a very nice and much needed rebalancing.

PR #930, opened 2020-12-21
Say bugfixes

Naxodile:
Fixes cases where lung damage would improperly prevent speaking, and ensures cases where speech is impossible are communicated.

Also fixes the IC markup system.


code/modules/mob/living/carbon/human/say.dm
```diff
@@ -131,6 +131,7 @@
 
 /mob/living/carbon/human/handle_speech_problems(var/list/message_data)
 	if(silent || (sdisabilities & MUTE))
+		to_chat(src,"<span class='danger'>You are unable to speak!</span>")
```
Glloyd, line 134: thanks for getting this in too.

PR #932, opened 2020-12-22
Readds boron and fixes the RUST 

ghost:
Turns out in a merge somewhere our RUST reactions got messed up, the hardest reaction was overriding one of the weakest because they were the exact same.

Also ports the RUST changing colors the hotter it is, and the warning when you try to deactivate it when it's unsafe.


Vakothu:
I _may_ have forgotten to log out of my Github burner account here, but the ghost was me!

Glloyd:
Thanks for this Vak! I really appreciate it!

Issue #933, opened 2021-01-04
Science Issues (Access, Sprites, Etc.)


Laniakeaaaa:
#### Description of issue

- Science IDs lack sprites when using alternative titles.
- Fixovein needs added to Robotics for Augments.
- (Possibly outdated) - Science NPC needs new clothes.
- Adjust Scientists to no longer spawn with UXO/Whatever Brand Labcoats.
- Senior Scientist has no access to Loadout for Science.

#### Difference between expected and actual behaviour
- Science IDs lack proper sprites when using a different title.
- Robotics apparently need Fixovein for augments now? (I haven't used or tested augments at all.)
- Scientists should spawn with normal labcoats or Nerva Brand.
- Senior Scientists should be able to access Science loadout settings.
- (Possibly depreciated) Science NPCs also lack labcoats now, whoops.

#### Steps to reproduce 

- All of these occur normally with Science Department on any round.

#### Specific information for locating
Loadout
Science Equip Datum
urist/npc
Science Labcoat (urist/clothing/under) or (nerva/clothing)

#### Length of time in which bug has been known to occur
- Baymerge
- Some occurred due to oversights in: https://github.com/UristMcStation/UristMcStation/pull/867

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round

Laniakeaaaa:
I'll handle these soon, but just wanted to get it up in case any downstreams, etc had issues.

PR #935, opened 2021-01-04
Science General Fixes (#933)

Laniakeaaaa:
- Fixes multiple issues attached in issue report: #933

- [ ] Fix Icon Pathing


Glloyd:
There's a couple minor things I want to tweak with this, but I'll do that myself so you don't have to worry about it

PR #944, opened 2021-04-12
Resolve add_antag issues

Naxodile:
attempt_late_antag_spawn would previously never get antag datums. Ghosts would have a blank body created, but fail the second round of add_antag due to them already being made antags.


Glloyd (approved):
Love that branch name

PR #952, opened 2021-04-22
Fixed hand teleporter

CPU-Blanc:
Fixes an issue with the hand teleporter and z-level checking, causing the hand teleporter to not lock onto the target but teleport you randomly nearby instead.


Glloyd:
Thanks for this Blanc!

PR #954, opened 2021-04-24
Couple of medical scan bugfixes

CPU-Blanc:
- Fixes the Medical Scanner printouts to actually print the reports correctly
- Fixes the admin verb Analyse Human Health to use the correct medical data and actually work

### DIFF
--- a/code/modules/admin/verbs/debug.dm
+++ b/code/modules/admin/verbs/debug.dm
@@ -433,7 +433,7 @@
 
 	if(!H)	return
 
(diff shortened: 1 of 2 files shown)


Glloyd (changes requested):
You'll have to take the skill stuff out of this because we've disabled it everywhere else. Besides that, it should be good.

CPU-Blanc:
Thought as the exact check was being done during the scan itself it was disabled further upstream. My bad. Out now c:

PR #959, opened 2021-04-25
Make Mentors Useful Again

CPU-Blanc:
Re-added Mentors, finally.
If any permissions are missing or wonky, please do let me know as permission/rank handling in baycode is inconsistent at best and hard to test with just 1 alt. Though as far as I could test it all works.

Current perms:

- AdminPMs
- aghost
- Subtlemessage
- Check contents
- VV (No edit)
- PP
- msay
- dsay
- aooc
- Check antagonists
- jmp to mob/key


Glloyd:
👍 looks good

PR #961, opened 2021-04-29
Bringing social media back to the station (for real this time)

Cheedows:
Makes several changes to the email client to make it more similar to the old PDA messenger system. Overall streamlines the client, making sending messages faster with less hassle and button presses and will hopefully once again encourage people to use their PDAs to speed text when being held at gunpoint.

Main changes:

- Completely changes the "New Message" functionality, previously to send an email you had to fill out three fields separately, the Title/Subject, the Recipient, and the message body itself
- Now when selecting "New Message" you are immediately brought to the address book with all of the crewmembers listed. Selecting any of their names will open up the message prompt box, and after you hit ok after typing in your message the message immediately sends. Exactly like the old PDA system, fast and only needs 2 button clicks right to the point.
- Reply functionality is also greatly improved as well. Previously, replying would bring you right back to the new message screen and you would have to click again to fill out the message. Now, you just hit Reply when seeing the notification, and the new message prompt box pops up immediately. You type in your message, press ok and it's sent. For speedy texting on the go.
- Notification message has been changed slightly. Removes the noise of titles (since they are gone anyways) and bolds who is sending you the message so it stands out more.
- Still keeps the same new system with the modular computers and programs. All this changes do is change the overall look and functionality of the Email Client program to encourage it's use among the crew. Now it's basically old PDAs with the new nano UI.

### DIFF
--- a/code/modules/modular_computers/file_system/programs/generic/email_client.dm
+++ b/code/modules/modular_computers/file_system/programs/generic/email_client.dm
@@ -1,7 +1,7 @@
 /datum/computer_file/program/email_client
-	filename = "emailc"
-	filedesc = "Email Client"
-	extended_desc = "This program may be used to log in into your email account."
+	filename = "messengerc"
+	filedesc = "Messenger Client"
+	extended_desc = "This program may be used to log in into your messenger account."
 	program_icon_state = "generic"
 	program_key_state = "generic_key"
 	program_menu_icon = "mail-closed"
@@ -58,12 +58,12 @@
 		ui_header = "ntnrc_idle.gif"
 
 /datum/nano_module/email_client/
-	name = "Email Client"
+	name = "Messenger Client"
 	var/stored_login = ""
 	var/stored_password = ""
 	var/error = ""
 
-	var/msg_title = ""
+	//var/msg_title = ""
 	var/msg_body = ""
 	var/msg_recipient = ""
 	var/datum/computer_file/msg_attachment = null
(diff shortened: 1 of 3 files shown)


code/modules/modular_computers/file_system/programs/generic/email_client.dm
```diff
@@ -1,7 +1,7 @@
 /datum/computer_file/program/email_client
-	filename = "emailc"
-	filedesc = "Email Client"
-	extended_desc = "This program may be used to log in into your email account."
+	filename = "messengerc"
```
Glloyd, line 2: if you're changing the filename, you'll also need to change the filename in code\modules\modular_computers\computers\subtypes\preset_pda.dm. Line 23, needs to be updated.

Glloyd (changes requested):
Besides that, it seems pretty good after brief testing. You will also need to update the language on line 42 if you're going with the email > message change: https://github.com/UristMcStation/UristMcStation/blob/7b87302cc3aabd510ce54e0811a230669fb4dc3d/code/modules/modular_computers/file_system/programs/generic/email_client.dm#L42

code/modules/modular_computers/file_system/programs/generic/email_client.dm
Cheedows: Got it thanks, applied and pushed those 2 changes.

Glloyd (approved):

Glloyd:
Looks good! I'll get this in the server for tonight.

PR #971, opened 2021-05-14
Fixed Coolant

CPU-Blanc:
Fixes coolant breaking all laws of physics and thermodynamics by allowing it to cool gasses to absolute zero. Now coolant when sprayed in the open can cool to a max of 10C, making it a little more plausible.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Other.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Other.dm
@@ -361,7 +361,7 @@
 		return
 
 	var/datum/gas_mixture/environment = T.return_air()
-	var/min_temperature = 0 // Room temperature + some variance. An actual diminishing return would be better, but this is *like* that. In a way. . This has the potential for weird behavior, but I says fuck it. Water grenades for everyone.
+	var/min_temperature = 283.15 // 10C. Because coolant cooling gasses to 0 kelvin makes zero sense.
 
 	var/hotspot = (locate(/obj/fire) in T)
 	if(hotspot && !istype(T, /turf/space))
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Other.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Other.dm
@@ -361,7 +361,7 @@
 		return
 
 	var/datum/gas_mixture/environment = T.return_air()
-	var/min_temperature = 283.15 // 10C. Because coolant cooling gasses to 0 kelvin makes zero sense.
+	var/min_temperature = 263.15 // -10C.
 
 	var/hotspot = (locate(/obj/fire) in T)
 	if(hotspot && !istype(T, /turf/space))


Dimasw99:
The 10C is brutal as it makes it to near no diffrence to water. 100K seems much more reasonable.

CPU-Blanc:
100K is not reasonable. That's **-173.15** Celsius. Stupidly over powered making it stronger than heat exchange pipes to space, multiple upgraded coolers etc, all for basically free with no energy cost.

Dimasw99:
Heat exchangers can go below 100K as they are not being exposed to sun light, multiple coolers are also well capable of reaching that temperature and even below and the energy cost is not something you would notice anyway.

The main diffrence here is the time investment to achieve the temperature and engineering has the issue that setting everything up takes a lot of time even up to an hour because of how few people play engineering.

CPU-Blanc:
And those other methods are balanced by the time and energy cost (and in the case of the coolers, RnD effort)
A few sprays of coolant should **not** do as much, or more work than a fully upgraded cooler in just a few seconds. That is beyond poor balancing.

Dimasw99:
You should play engineering to see how insane the time investment already is before you start gutting tools that make the process quicker

CPU-Blanc:
I've played SS13 for over 10 years at this point. I'm _well_ aware of how much time and effort things take.
The benefits of using coolant/coolers isn't even a necessity like you're making it sound like it is. A solid engine setup can be done in 10 or so minutes. If you want to min/max it all, then yes, it'll take longer.

Dimasw99:
10 minutes is for the bare minimum to get the lights working while ignoring everything else like shield and thrusters. If you don't want to restart after a single ship combat as everything will be vented then you have to make a beefier engine and a more robust shield setup. 

I wouldn't be as opposed to this nerf if it wasn't for the low player count making engineering practically a solo job. Cutting down time anywhere is huge for the playing experience

CPU-Blanc:
I don't know what world you're living on, but a basic engine setup without spending ages on cooling gases is more than enough so sustain shields against pirate ships that may wander into the Nerva's co-ords, as well as allow thrust to navigate. Heck, if you _really_ wanted to, you can navigate just by opening the fuel valve to max without the SM even being energised.

Super chilling phoron is completely unnecessary but just an added bonus. 
Funnily enough Dimas, the ship doesn't fall apart every time your OP engine setup isn't used. It manages just fine and again, isn't necessary at all, just a bonus.

Dimasw99:
After the ship combat update all enemy ships are way more dangerous and even a pirate ship has enough fire power to penetrate one of my strongest engine and shield setups

CPU-Blanc:
If that's the apparent case, then that's a call for space combat to be rebalanced, not power generation c:

But I've made myself clear enough, any more and I'll just be running in circles. There's further power generation to be had from installing collectors, the solars and other creative means without resorting to hacky physics if it's _really_ needed.

Dimasw99:
That is the issue. Even with more collectors it isn't enough right now for the shields. Solars in this scenario is also just a droplet in a ocean when it comes to power output. 

And again to restate my point. The only consquence of this nerf is the increased time investment which is problematic with this pop count.

Dimasw99:
To further elaborate on my opinion here. I agree that going down to 0 Kelvin is bonkers here so I suggest to have it go down to a middle number like 100K or 150K which is low enough to justify its use but still give coolers a job to further improve on this number.
In essence it would cut down the time investment and give all parts a job

Glloyd:
Thanks Blanc! That's more reasonable and still gives coolant a use.

PR #972, opened 2021-05-16
Shuttles Redux

Vakothu:
Whole bunch of touchups to the shuttles around the map, every shuttle that could reasonably dock with the Nerva can now, and we only need one waypoint per docking port.
Also misc. bug fixes.


Vakothu:
Gonna close and open the PR to kick Travis, all those checks failed because Github was down.

Glloyd:
Map diff bot isn't working for some reason, so I'm just going to trust that everything's okay mapwise.

Vakothu:
Yeah, it's mostly variable changes to make the docking airlocks actually dock.

Glloyd:
Seems good from my limited testing. Only thing I can see is that the Merc shuttle can't land anymore. Although that might be on me and not on you, I'm not sure.

Vakothu:
Lemme test, I might have buggered something up.

Vakothu:
Okay, I checked and on the server's current version the nuke ops shuttle can't dock either.

Vakothu:
Found the problem, merc docking waypoint has the wrong base_area, and unrelated but the Merc shuttle also doesn't have a docking controller.

PR #973, opened 2021-05-16
Adds more stuff to R'nD than I originally planned to

Razberyl:
Can't compile this on my end to see if it works.
Adds the arc welder to RnD at Vak's suggestion.

Hopefully re-adds the bluespace power cell to RnD that we lost at some point, and changes the file with the bluespace trash bag to be slightly more consistent.

If I need to change any more files to make this work, lemme know. I asked already but got no answer.


Glloyd:
Looks good, thanks Raz!

PR #975, opened 2021-05-30
AI construction fixes

Dimasw99:
Adding glass to the AI core was fixed and transfer_to proc used for mind.dm has its skillset obtain disabled due to runtimes. Shouldn't be an issue until skills are used on Urist


Naxodile (commented):
The AI changes all check out, you've just got an extra commit on your branch.

PR #976, opened 2021-06-03
Surgery & Cloning bugfixes

Naxodile:
Arterial bleeding in encased organs (upper body, head) can now be fixed in situations where the organ is also broken.
Robotic organs in non-robotic external organs (read: laces) can now be repaired.
Cloning will properly calculate the amount of damage left to heal before a clone is released.


Glloyd:
Looks good! Thanks Nax

PR #980, opened 2021-06-07
Update Chef & Barkeep guides to our bay version

CPU-Blanc:
Because the current recipes don't work on our build.

### DIFF
--- a/code/modules/library/manuals/manuals.dm
+++ b/code/modules/library/manuals/manuals.dm
@@ -34,7 +34,7 @@
 				Put 3 processed eggs, 5 flour and 5 milk into a beaker. Then wait.
 
 				<h3>Cakebatter:</h3>
-				Add a pinch of sugar to 2 parts of regular batter.area
+				Add a pinch of sugar to 2 parts of regular batter.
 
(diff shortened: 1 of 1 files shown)


Vakothu (commented):
Oh boy, cooking, even in the updated version there's some, well, I won't call them mistakes, but misleading parts and the deprecated food processor.
On lines 25 and 34 you mentioned "three processed eggs", which while technically accurate, is a confusing way to put it, both of them need 3u of egg yolk, which you get from cracking a single egg into a bowl or beaker.
You mentioned spaghetti as "processed flour", the food processor has been deprecated for a while and now the way to get raw spaghetti is make some dough, roll it flat, cut the flat dough into dough slices, then cut the dough slices into raw spaghetti.
Line 58 also mentions the deprecated food processor, now you get potato sticks from chopping up a potato.

CPU-Blanc:
My bad. Uncommented code in processor.dm made me think this was still the way it was done. Fixed now!
Thanks!

Glloyd:
If @Vakothu is satisfied with the changes, I'll merge this.

PR #984, opened 2021-06-14
Multi Z-level repair additions

CPU-Blanc:
Currently, repairing hull breaches than span over multiple z-levels is a real pain. There's always one tile you've missed _somewhere_ that's almost impossible to find, running up the stairs and back down while the bridge which has inevitably been hit by a missile is venting atmosphere. And good luck if the breach is from Deck 1 into 2 with that maze of maintenance tunnels. So, I thought I'd try to make that a little easier. This is my first PR that isn't some kind of bugfix and is an actual addition, so feel free to yell at my incompetence or general issues with it.

This adds the following:

- **Rod stacks have a new verb** "Construct Lattice Above", available in the context menu and Object tab. This allows the player to construct a lattice on an open space tile above their current location. Negating the need to run upstairs and attempt to find the hole. Does not allow further construction other than lattice, ie catwalks.
- **Floor tiles also have a similar verb** "Place Tile Above", again available in the context menu and Object tab. This again allows the player to place a floor tile down onto a lattice above them. Does not allow full floor construction, ie placing a floor tile onto plating.
- **T-Scanners now gain a secondary scan mode**, toggled by either the context menu or in the Object tab. This mode, instead of revealing pipes and wires under tiles, will instead scan above you and overlay any tiles missing plating, currently denoted by a red X. This icon is a placeholder because I am incompetent at anything artistic so I just used something that fit. Has the same range as the standard mode. Side-effect of this is a fix of an oversight in the original code: The t-scanners do scan invisible mobs (line 124), but would be filtered out regardless in line 67. This tweak passes all atoms and instead relies on get_scanned_objects to filter type.

Example from above:
Screenshot from 2021-06-14 00-39-00

And the damage below with the mode active (again, placeholder icons):
Screenshot from 2021-06-14 00-39-24

Again, if anything is wrong or it's simply a bad idea, feel free to yell in my general direction.

### DIFF
--- a/code/game/objects/items/devices/t_scanner.dm
+++ b/code/game/objects/items/devices/t_scanner.dm
@@ -10,7 +10,8 @@
 	matter = list(MATERIAL_ALUMINIUM = 150)
 	origin_tech = list(TECH_MAGNET = 1, TECH_ENGINEERING = 1)
 	action_button_name = "Toggle T-Ray scanner"
-
+	
+	var/standard_mode = TRUE
 	var/scan_range = 3
 
 	var/on = 0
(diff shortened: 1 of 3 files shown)


Glloyd:
Looks good 👍 

I also added an alt-click switch to the t-ray scanner to make it easier to switch modes.

PR #991, opened 2021-06-21
Reverts baymerge atmos nerf

Naxodile:
Ruins the afk engine strategy.

@Glloyd It's been ages since this got put in place during the baymerge testing, do you remember the reasoning?


Dimasw99:
I would then increase the power rating of the engine vent and large vent too or else they would be weaker then normal vents.
line 87 and 100

Glloyd:
I don't remember why this was done, but Dimas is right that the other values should be adjusted accordingly.

Glloyd:
Looks good 👍

PR #995, opened 2021-06-24
Feature - PvP Ship Combat + Combat Bugfixes

CPU-Blanc:
Bit of a big one, so I'll try to break it down.
This an overhaul of the current ship combat system to allow for future ship-to-ship pvp combat. Also includes many bugfixes to the current PvE system, UI tweaks and general QoL improvements.

Included in this PR...

### Bug Fixes

- Fixes weapons not charging once power has been lost/regained
- Fixes various nested for loops, causing damage to potentially duplicate
- Fixes various issues causing weapons to become unresponsive and not work, or fire multiple times
- Fixes a code oversight that allowed missile weapons to fire without checking for ammo when fired from the console
- Fixes fallback connect for the weapons console not working, due to not populating the global list it attempts to pull from
- Fixes general weapon status updates to fail and not show on the UI
- Fixes all ship components being targeted on combat start, causing the combat system to get confused
- Fixes that _really_ annoying off-centre 'Fire' button in the combat UI. My OCD.
- Fixes duplicate damage call-outs, caused by remembering damage call-outs per weapon and not in one place
- Fixes general jank in the Fire proc, causing damage to happen or not happen depending on how the game feels

### Tweaks

- Tweaked the security level alerts: Verbiage has been changed station -> ship, downgrade sound has been changed from scaring the shit out of you, to a gentle ding.
- Comms announcement when missile fire has been intercepted by the AI's point defence.
- Ship name & classification are now their own vars, allowing cleaner comm messages or UI prompts.

### Features

- Various combat UI tweaks: Now showing the AI ship's health and shields in bars, component integrity, clear weapon status updates, general housekeeping
- New UI to handle pvp content: Rather basic for now, but functionality is the focus on this PR. Plans to expand on this later
- PvP capability: Now any ship of type `obj/effect/overmap/ship/combat` with the correct vars & landmarks added can engage in combat with either the Nerva, or pirates. Currently teleport beacons and shields have not been split on a per-ship basis, so boarding parties and shield disruptors will not work against ships other than the Nerva

Some PvP features are placeholders or just ideas that can be tweaked depending on community feedback. Currently both ships will lock in place much like PvE content. Intercepted ships will also have to restableize their engines in order to escape. Currently the time this takes is dependant on the difference in ship mass between the 2 ships, with smaller ships having a quicker time to retreat. This amount this is by is _very_ placeholer-ish, but currently have a min/max of 5/15 minutes with 10 minutes as the baseline.
Ships that have left PvP (both attacking and defending) then have a 10 minute cooldown in which they can no longer engage other player ships. This is to help prevent combat locking another vessel. (Pirates can still engage or be engaged though!)

Notes on adding new PvP ships:

- New PvP ships will require ship projectile landmarks with a matching shipid to their assigned ship (`/obj/effect/urist/projectile_landmark/ship`)
- The lists `target_x_bounds(x1,x2)`, `target_y_bounds(y1,y2)` must be present. This is a bounding box for weapon fire to land in
- If new frequencies are ever made, they must be added to the `announcement_channel` list on the overmap ship. `public`, `private`, and `technical` keys are used for feedback.

Future features to expand on:
- Add per-ship shield checks: To allow for pirate shield disruptors (or pvp shield disturptors??), and to allow for shield stats to be shared between ships in combat.
- Add per-ship teleport beacons: To allow pirate boarders on non-nerva ships.
- Possible ability for other combat ships to have their own telecom networks and frequencies: This would allow for combat feedback to those ships. Base work for this has been done, making a possible future transition easier.

I'm sure I've probably made some mistakes somewhere, I've been staring at the same code a few days while battling with nanoUI/html. Let me know if there's any issues or general feedback!

### DIFF
--- a/code/datums/security_state.dm
+++ b/code/datums/security_state.dm
@@ -70,6 +70,12 @@
 		high_security_level = severe_security_level
 
 /decl/security_state/Initialize()
+	var/regex/reg = new/regex("(VESSEL_TYPE)")
+	for(var/decl/security_level/default/SL in all_security_levels)
+		if(SL.down_description)
+			SL.down_description = replacetext(SL.down_description, reg, GLOB.using_map.vessel_type_name)
+		if(SL.up_description)
+			SL.up_description = replacetext(SL.up_description, reg, GLOB.using_map.vessel_type_name)
 	// Finally switch up to the default starting security level.
 	current_security_level.switching_up_to()
 	. = ..()
@@ -210,7 +216,7 @@
 	overlay_alarm = "alarm_green"
 	overlay_status_display = "status_display_green"
 
-	down_description = "All threats to the ship have passed. Security may not have weapons visible, privacy laws are once again fully enforced."
+	down_description = "All threats to the VESSEL_TYPE have passed. Security may not have weapons visible, privacy laws are once again fully enforced."
 
 /decl/security_level/default/code_blue
 	name = "code blue"
@@ -226,7 +232,7 @@
 	overlay_alarm = "alarm_blue"
 	overlay_status_display = "status_display_blue"
 
-	up_description = "The bridge has received reliable information about possible hostile activity on the ship. Security staff may have weapons visible, random searches are permitted."
+	up_description = "Reliable information has been received about possible hostile activity on the VESSEL_TYPE. Security staff may have weapons visible, random searches are permitted."
 	down_description = "The immediate threat has passed. Security may no longer have weapons drawn at all times, but may continue to have them visible. Random searches are still allowed."
 
 /decl/security_level/default/code_red
@@ -243,8 +249,8 @@
 
 	psionic_control_level = PSI_IMPLANT_DISABLED
 
-	up_description = "There is an immediate serious threat to the ship. Security may have weapons unholstered at all times. Random searches are allowed and advised."
-	down_description = "The self-destruct mechanism has been deactivated, there is still however an immediate serious threat to the ship. Security may have weapons unholstered at all times, random searches are allowed and advised."
+	up_description = "There is an immediate serious threat to the VESSEL_TYPE. Security may have weapons unholstered at all times. Random searches are allowed and advised."
+	down_description = "The self-destruct mechanism has been deactivated, there is still however an immediate serious threat to the VESSEL_TYPE. Security may have weapons unholstered at all times, random searches are allowed and advised."
 
 /decl/security_level/default/code_delta
 	name = "code delta"
--- a/code/datums/security_state.dm
+++ b/code/datums/security_state.dm
@@ -70,12 +70,11 @@
 		high_security_level = severe_security_level
 
 /decl/security_state/Initialize()
-	var/regex/reg = new/regex("(VESSEL_NAME)")
 	for(var/decl/security_level/default/SL in all_security_levels)
 		if(SL.down_description)
-			SL.down_description = replacetext(SL.down_description, reg, GLOB.using_map.full_name)
+			SL.down_description = replacetext(SL.down_description, "VESSEL_NAME", GLOB.using_map.full_name)
 		if(SL.up_description)
-			SL.up_description = replacetext(SL.up_description, reg, GLOB.using_map.full_name)
+			SL.up_description = replacetext(SL.up_description, "VESSEL_NAME", GLOB.using_map.full_name)
 	// Finally switch up to the default starting security level.
 	current_security_level.switching_up_to()
 	. = ..()
--- a/code/datums/security_state.dm
+++ b/code/datums/security_state.dm
@@ -179,7 +179,7 @@
 	icon = 'icons/misc/security_state.dmi'
 
 	var/static/datum/announcement/priority/security/security_announcement_up = new(do_log = 0, do_newscast = 1, new_sound = sound('sound/misc/notice1.ogg'))
-	var/static/datum/announcement/priority/security/security_announcement_down = new(do_log = 0, do_newscast = 1, new_sound = sound('sound/misc/notice1.ogg'))
+	var/static/datum/announcement/priority/security/security_announcement_down = new(do_log = 0, do_newscast = 1, new_sound = sound('sound/misc/notice2.ogg'))
 
 /decl/security_level/default/switching_up_to()
 	if(up_description)
@@ -210,7 +210,7 @@
 	overlay_alarm = "alarm_green"
 	overlay_status_display = "status_display_green"
 
-	down_description = "All threats to the station have passed. Security may not have weapons visible, privacy laws are once again fully enforced."
+	down_description = "All threats to the ship have passed. Security may not have weapons visible, privacy laws are once again fully enforced."
 
 /decl/security_level/default/code_blue
 	name = "code blue"
@@ -226,7 +226,7 @@
 	overlay_alarm = "alarm_blue"
 	overlay_status_display = "status_display_blue"
 
-	up_description = "The station has received reliable information about possible hostile activity on the station. Security staff may have weapons visible, random searches are permitted."
+	up_description = "The bridge has received reliable information about possible hostile activity on the ship. Security staff may have weapons visible, random searches are permitted."
 	down_description = "The immediate threat has passed. Security may no longer have weapons drawn at all times, but may continue to have them visible. Random searches are still allowed."
 
 /decl/security_level/default/code_red
@@ -243,8 +243,8 @@
 
 	psionic_control_level = PSI_IMPLANT_DISABLED
 
-	up_description = "There is an immediate serious threat to the station. Security may have weapons unholstered at all times. Random searches are allowed and advised."
-	down_description = "The self-destruct mechanism has been deactivated, there is still however an immediate serious threat to the station. Security may have weapons unholstered at all times, random searches are allowed and advised."
+	up_description = "There is an immediate serious threat to the ship. Security may have weapons unholstered at all times. Random searches are allowed and advised."
+	down_description = "The self-destruct mechanism has been deactivated, there is still however an immediate serious threat to the ship. Security may have weapons unholstered at all times, random searches are allowed and advised."
 
 /decl/security_level/default/code_delta
 	name = "code delta"
--- a/code/datums/security_state.dm
+++ b/code/datums/security_state.dm
@@ -70,12 +70,12 @@
 		high_security_level = severe_security_level
 
 /decl/security_state/Initialize()
-	var/regex/reg = new/regex("(VESSEL_TYPE)")
+	var/regex/reg = new/regex("(VESSEL_NAME)")
 	for(var/decl/security_level/default/SL in all_security_levels)
 		if(SL.down_description)
-			SL.down_description = replacetext(SL.down_description, reg, GLOB.using_map.vessel_type_name)
+			SL.down_description = replacetext(SL.down_description, reg, GLOB.using_map.full_name)
 		if(SL.up_description)
-			SL.up_description = replacetext(SL.up_description, reg, GLOB.using_map.vessel_type_name)
+			SL.up_description = replacetext(SL.up_description, reg, GLOB.using_map.full_name)
 	// Finally switch up to the default starting security level.
 	current_security_level.switching_up_to()
 	. = ..()
@@ -216,7 +216,7 @@
 	overlay_alarm = "alarm_green"
 	overlay_status_display = "status_display_green"
 
-	down_description = "All threats to the VESSEL_TYPE have passed. Security may not have weapons visible, privacy laws are once again fully enforced."
+	down_description = "All threats to the VESSEL_NAME have passed. Security may not have weapons visible, privacy laws are once again fully enforced."
 
 /decl/security_level/default/code_blue
 	name = "code blue"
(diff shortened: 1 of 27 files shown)


code/datums/security_state.dm
```diff
@@ -210,7 +210,7 @@
 	overlay_alarm = "alarm_green"
 	overlay_status_display = "status_display_green"
 
-	down_description = "All threats to the station have passed. Security may not have weapons visible, privacy laws are once again fully enforced."
+	down_description = "All threats to the ship have passed. Security may not have weapons visible, privacy laws are once again fully enforced."
```
martinlyra, line 213: This is a change that might not sit well with the plans to host Glloydstation on odd days. Since the alert will refer the station as a ship instead. How about we make another addition to the map datum, so we could use `[USING_MAP.vessel_type_name]` in place of station or ship instead?

Unfortunately. These variables expect those strings to be constants. So we would need to put them in functions or use regex to initialize those strings proper.

code/modules/urist/modules/shipbattles/overmapships/components/advanced_components.dm
```diff
@@ -15,11 +15,11 @@
 	var/counterboarding = FALSE
 
 /datum/shipcomponents/teleporter/DoActivate()
-	if(mastership.boarding && !counterboarding)
+	if((mastership.boarding && !counterboarding) || mastership.target_ship.shipid != "nerva")
```
martinlyra, line 18: Again, we could use `USING_MAP` here instead of using a constant. Or it is going to be a pain to debug why shields won't work on ships that is not Nerva, for example.

code/modules/urist/modules/shipbattles/overmapships/components/advanced_components.dm
```diff
@@ -134,7 +134,7 @@
 	if(last_activation > world.time)
 		return
 
-	else
+	else if(mastership.target_ship.shipid == "nerva")	//TODO: Bind shield gens per combat ship. Check is temp to stop weirdness
```
martinlyra, line 137: Same as before

code/modules/urist/modules/shipbattles/overmapships/components/components.dm
```diff
@@ -5,15 +5,17 @@
...
 	var/last_activation = null
 
 /datum/shipcomponents/proc/BlowUp()
 //	qdel(src)
 	mastership.health -= 100
 	broken = TRUE
+	targeted = FALSE
 	name = "destroyed [initial(name)]"
 	if(mastership.health <= 0)
+		mastership.health = 0
```
martinlyra, line 18: Is this necessary?

code/datums/security_state.dm
CPU-Blanc: Good shout and idea! Added in 7998729

code/modules/urist/modules/shipbattles/overmapships/components/advanced_components.dm
CPU-Blanc: Shouldn't make debugging harder as area checks later on currently use `USING_MAP` regardless. Was just a temporary constant to stop pirates fighting another ship from effecting the Nerva. Shouldn't be an issue as I don't believe we have any other maps that can do space combat, but just in case, patched in d7fc3d7

code/modules/urist/modules/shipbattles/overmapships/components/advanced_components.dm
CPU-Blanc: Patched in d7fc3d7

code/modules/urist/modules/shipbattles/overmapships/components/components.dm
CPU-Blanc: Whoops! This was an artefact left from previous attempts to get the nanoUI to not display negative pirate health on destruction. This was fixed further down the line and I forgot to remove this.
Patched in d5c3d44

CPU-Blanc:
**Update:**
As per the previous suggestion, map datums now have a `vessel_type_name` variable used to change the security alert verbiage.

Glloyd:
That security alert is certainly an outlier, most things use the using_map.name nowadays.

CPU-Blanc:
I'm happy either way. If you'd rather use `using_map.full_name` (for better grammar vs just .name), I can revert the changes to the map datums and use that instead, just let me know.

Glloyd:
Yeah, that's var I was thinking of. I think that is the better solution, vs. adding a new var that would need to be added for every map. It reads better too, and is more consistent with other alert messages.

Glloyd (changes requested):
I haven't looked through it completely yet, but these are a few things I noticed at a first glance

code/datums/security_state.dm
```diff
@@ -70,6 +70,12 @@
 		high_security_level = severe_security_level
 
 /decl/security_state/Initialize()
+	var/regex/reg = new/regex("(VESSEL_NAME)")
```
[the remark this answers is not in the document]
CPU-Blanc: Regex is needed due to the descriptors being constant expressions,  it cannot take a var directly in the string without erroring on compile. If I'm doing something stupid than let me know, but for me, adding the var directly into the string doesn't work on build

code/modules/urist/modules/shipbattles/combatcomputer.dm
```diff
@@ -93,32 +93,69 @@
...
+		var/integrity = round(max((OM.health / OM.maxHealth) * 100, 0), 0.01)
+		var/maxshields = 0
 
 		for(var/datum/shipcomponents/C in OM.components)
 			var/status
 			if(C.broken)
 				status = "Broken"
-			else if(!C.broken)
+			else if(istype(C, /datum/shipcomponents/shield))
+				maxshields = C:strength
```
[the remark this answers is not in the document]
CPU-Blanc: Happy to change it, but I'm a little confused if it's necessary as we're already explicitly checking the type in the line above, so it would be impossible to call C:strength unless it is of type /datum/shipcomponents/shield regardless?

code/modules/urist/modules/shipbattles/overmapships/components/advanced_components.dm
```diff
@@ -134,7 +134,7 @@
 	if(last_activation > world.time)
 		return
 
-	else
+	else if(mastership.target_ship.shipid == lowertext(GLOB.using_map.name))	//TODO: Bind shield gens per combat ship. Check is temp to stop weirdness
```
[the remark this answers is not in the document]
CPU-Blanc: Must admit, I completely missed that var. Makes things much nicer. Will be patched in a future commit along with the other changes. Thanks!

code/modules/urist/modules/shipbattles/combatcomputer.dm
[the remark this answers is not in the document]
Glloyd: It's not strictly necessary to make it work per se, but it's far better for maintaining a long-term project such as this one. Using the : operator, if the strength or overcharged vars ever changed or were removed, this function would error out, and it wouldn't be caught by the compiler as it would be using the . operator. Because SS13 codebases are these big long-term projects with multiple contributors, it's best not to do something that can potentially introduce hard to track errors down the line when there's the . operator instead (even the BYOND docs recommend against using the : operator for the same reason)

code/modules/urist/modules/shipbattles/overmapships/components/advanced_components.dm
```diff
@@ -32,30 +32,30 @@
 	boarding_failure_chance = 0 //Zero this var out.
 
 	if(!bypass_shields)
-		for(var/obj/machinery/power/shield_generator/S in SSmachines.machinery) //Calculate our failure chance.
+		for(var/obj/machinery/power/shield_generator/S in SSmachines.machinery) //Calculate our failure chance. //TODO: Bind shield gens per combat ship
 			if(S.z in GLOB.using_map.station_levels)
 				if(S.running == SHIELD_RUNNING)
 					boarding_failure_chance += 25 // four shield generators to TOTALLY block boarding.
```
Naxodile, line 38:
```suggestion
		for(var/obj/machinery/power/shield_generator/S in SSmachines.machinery) //Calculate our failure chance.
			if(S.z in mastership.target_ship.map_z)
				if(S.running == SHIELD_RUNNING)
					boarding_failure_chance += 25 // four shield generators to TOTALLY block boarding.
```
Can be applied to all the other TODOs

maps/away/bearcat/bearcat.dm
```diff
@@ -16,7 +16,7 @@
 	)
 
 /obj/effect/overmap/ship/bearcat
-	name = "light freighter"
+	classification = "Light Freighter"
```
Naxodile, line 19: Capitalizing these makes them proper nouns, getting you "Bearcat, Light Freighter". Depending on what you want,
```suggestion
	classification = "light freighter"
```
or
```suggestion
	classification = "\improper Light Freighter"
```

Naxodile (changes requested):
Haven't tested this in game yet, just a couple things I noticed.

code/datums/security_state.dm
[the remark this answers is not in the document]
CPU-Blanc: Fixed in 916e5f6ceb786fb1d569467e3980e98358915e24

code/modules/urist/modules/shipbattles/combatcomputer.dm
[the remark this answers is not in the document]
CPU-Blanc: Ahh that makes sense! Thanks for the explanation! Switched to proper type in all files in 9b46ce7fe1b09c41eea235645a4927e7ecc690c9

code/modules/urist/modules/shipbattles/ship_overmaps.dm
```diff
@@ -122,4 +155,104 @@
...
+		autoannounce("<b>[T.ship_name] has exceeded weapons range - Exiting combat.</b>", "public")
+
+	if(shipid == lowertext(GLOB.using_map.name))	//If the Nerva is involved, put the alert level back where it was
+		var/decl/security_state/security_state = decls_repository.get_decl(GLOB.using_map.security_state)
+		security_state.set_security_level(security_state.stored_security_level)
+
+	T.leave_pvp_combat(!fled)	//Calls the other ship to leave. Won't loop back as target was cleared.
+
+/obj/effect/overmap/ship/combat/proc/autoannounce(var/message, var/channel)	//Moved all combat announcements to call this proc instead. In future, other player ships might have their own frequencies
+	if(!message || !channel)	return										//Stops any player ships without their own freq using the Nerva's, which would be wierd.
```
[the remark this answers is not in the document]
CPU-Blanc: Changed in 9b46ce7fe1b09c41eea235645a4927e7ecc690c9

code/modules/urist/modules/shipbattles/combatcomputer.dm
```diff
@@ -93,32 +93,69 @@
...
+				if(C:overcharged)	//If shields are overcharged, let's display that.
```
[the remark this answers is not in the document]
CPU-Blanc: Fixed in 9b46ce7fe1b09c41eea235645a4927e7ecc690c9

code/modules/urist/modules/shipbattles/combatcomputer.dm
```diff
@@ -93,32 +93,69 @@
...
+		data["status"] = 2
+		data["targetname"] = OM.ship_name
+		data["classification"] = OM.classification
+		data["target_flee_timer"] = OM.flee_timer
+		data["target_can_escape"] = OM.can_escape
+		data["self_flee_type"] = homeship.fleeing
+		data["self_flee_timer"] = homeship.flee_timer
+		data["self_can_escape"] = homeship.can_escape
+
+	else if(homeship.contacts:len)
```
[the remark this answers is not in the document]
CPU-Blanc: Whoops. Patched in 9b46ce7fe1b09c41eea235645a4927e7ecc690c9

code/datums/security_state.dm
```diff
@@ -70,6 +70,12 @@
...
+	for(var/decl/security_level/default/SL in all_security_levels)
+		if(SL.down_description)
+			SL.down_description = replacetext(SL.down_description, reg, GLOB.using_map.full_name)
+		if(SL.up_description)
+			SL.up_description = replacetext(SL.up_description, reg, GLOB.using_map.full_name)
```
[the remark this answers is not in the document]
CPU-Blanc: My bad, used to the old regex way. Patched in 916e5f6ceb786fb1d569467e3980e98358915e24

code/modules/urist/modules/shipbattles/overmapships/components/advanced_components.dm
CPU-Blanc: Ah! I was not aware this var existed! This helps a lot later down the road! Thanks a lot
Removed all TODO's and temporary `using_map` constraints in d8e9c2aa30c7b9268b2b1db6998bf43ce1472aa0

maps/away/bearcat/bearcat.dm
CPU-Blanc: Went with the first option and added the upper case helper to the nanoUI to compensate in 9b46ce7fe1b09c41eea235645a4927e7ecc690c9 
Thanks!

Glloyd:
I will try to look through this again later tonight. There's a lot of great stuff in here, so the sooner we get it in the better.

CPU-Blanc:
Just did a quick refactoring of the various statuses on the weapons. Seemed odd and messy to me to have 4 different vars dancing about when we can just use a single bitwise var to store various states.

Glloyd:
The UI looks great, and testing doesn't show any issues. I haven't tested PvP yet, but the code looks fine, so I think I'll just merge this now to get the fixes in, and then do further testing of PvP once we enable it for another ship (likely the Noctis)

Glloyd (approved):
Looks good from my end. unless @Naxodile has any other issues to raise, I'll merge.

Glloyd:
Nothing from Nax in 24 hours, so I'm just going to call this done and merge.

PR #1005, opened 2021-07-04
Bugfix - Fixed mined rocks not generating resources on new turf

CPU-Blanc:
Fixes an issue where when a new asteroid turf is created below a mined rock, it does not generate anything in the resources list, making the big drill unusable on those spaces.

There's probably a _much_ better way of doing this, but this works and my knowledge on map generation is somewhat limited.


Glloyd:
I also don't know too much about the mineral generation, but this works, and looks good to me, so 👍

PR #1006, opened 2021-07-08
Various NPC trader tweaks/bugfixes

CPU-Blanc:
Fixes a few issues currently present in NPC trading:

- Fixes not being able to sell items in your right hand, due to a missing `else`.
- Fixes `get_trade_value()` simply returning the global worth and not the trader's worth due to missing returns, resulting in _very_ high sale value, and often far above the buy value, creating an infinite money pit.
- Adjusts NPCs to correctly calculate selling multiple items at once; as once a sale is made, the price changes, and we need to apply the new price to each following item to better simulate loss of demand/the same price decrease as if sold individually.
- Removes a possible exploit by selling items in both hands at the same time, causing them both to be sold at the same rate, then sold back to the NPC for profit.

As a side-effect of this PR, most prices selling to NPCs have been nerfed compared to before. This again was due to the selling price being incorrectly tied to SS13's default worth procs, rather than the individual trader's.

### DIFF
--- a/code/modules/urist/modules/newtrading/NPC/NPC_types/complex_traders.dm
+++ b/code/modules/urist/modules/newtrading/NPC/NPC_types/complex_traders.dm
@@ -41,7 +41,7 @@
 	trade_categories_by_name = list("atmospherics","tools","bulky","medical")
 
 	sell_modifier = 0.8
-	price_increase = 1.1
+	price_modifier = 0.1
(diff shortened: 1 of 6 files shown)


Naxodile (approved):
Seems fine, worst case scenario Nerva fails its next interest payment.

Glloyd:
Looks good.

PR #1007, opened 2021-07-10
Gun in hand fixes and additions

Cheedows:
All credits for the sprites go to Dongwaiver, Erin and Bimmer. All this does is add the in-hand sprites for these guns to the game (we have already ported all the item sprites already and they've been in use for years, for whatever reason the in-hands also made alongside them never got ported in). Along with this because of the scope of all the guns covered (practically the vast majority of common ballistic pistols, rifles, and shotguns you'd find in-game normally) there were a number of bug and tag weirdness that also got resolved (item-states being mismatched causing incorrect inhand sprites, some items already previously referencing incorrect .dmi sprites so they had no inhands period) so things should be much more consistent moving forward.


Glloyd:
Thanks for this Cheeds, it looks good. There's some other weirdness with the boarding pistol that I noticed, but I'll take care of that myself.

PR #1010, opened 2021-07-11
Reloadable Energy Weapons - Mostly Done Project

Cheedows:
WIP Project to add the framework/functionality for energy weapons to be reloadable. As of now it is entirely possible and ready to go, however has massive balance and gameplay changes so additional features and checks need to be added to keep energy weapons still distinct and different from ballistic ones.  

Things to do and add/change:
- Implementing a check that restricts the size of cell that can be put into the energy gun, ideally something that is the default size of the cell it spawns with. Otherwise you would be able to slap in infinite energy cells and have infinite charge laser carbines which would be problematic.
- Change the reloading process to be distinct and slower compared to normal ballistic reloads. This would keep it more as a new convenience option to incentivize it's economical use for away missions etc. rather than speed reloads on the fly in combat. This would help retain the different feel and functionality of energy weapons compared to ballistic ones
- Any additional balance suggestions are welcome


Cheedows:
As of now there have been a couple of checks implemented now and some other features from above:
- Cell restrictions are now put in. It checks the actual original charge size of the original cell and compares it with the new one. This way you can never have a higher ammo capacity that the original. Error handling put in that puts a message to the user if they try to put in a bigger cell. As of now it will also accept smaller cell sizes, this can be changed in the check if needed
- Do_after restriction put in for reloads, tacticool reloads removed. Right now after testing it's set to 5 seconds the user has to be in place to put in the new cell (I handwaved it as configuring wires or some shit). With this, energy weapons now have a cell cover, opened and closed with a screwdriver. The cover has to be open before you can take out the existing cell or load in a new one. If the cover is open, you can't shoot the gun either. This effectively puts in another restriction to reloading/shooting, and 2 screwdriver actions on the fly needed for a mid-combat reload. This should be enough to keep them slow and ideally only when you have a break in cover.
- With all of the above, relevant messaging put in place, examine text changed to show if the cell cover is open/closed and if there is a cell loaded in or not

Cheedows:
Have now also disabled self charging weapons from being reloaded, as in the AEG's case especially with a mini nuclear reactor it doesn't make a lot of sense to be able to take out the battery. Additionally prevents the potential from taking out self charging batteries from AEGs and putting it into other guns.

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+				
+				if(!do_after(user, reload_time, src))
+					user.visible_message("[usr] stops reconfiguring the wires in [src].","<span class='warning'>You stop reconfiguring the wires in [src].</span>")
+					return
+				if(!user.unEquip(AM, src))
+					return
+				power_supply = AM
+				user.visible_message("[user] inserts [AM] into [src].", "<span class='notice'>You insert [AM] and hot wire it into [src].</span>")
+				playsound(loc, mag_insert_sound, 50, 1)
+				AM.update_icon()
```
Naxodile, line 137: Why are we updating the icon of an invisible icon?

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+			else
+				user.visible_message("<span class='warning'>The cell size is too big for the [src]. It must be [max_shots*charge_cost] Wh or smaller.</span>")
```
Naxodile, line 139:
```suggestion
				to_chat(user,"<span class='warning'>The cell size is too big for the [src]. It must be [max_shots*charge_cost] Wh or smaller.</span>")
```

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+				return
+		else
+			user.visible_message("<span class='warning'>The cell cover is closed. Use a screwdriver to open it.</span>")
```
Naxodile, line 142:
```suggestion
			to_chat(user,"<span class='warning'>The cell cover is closed. Use a screwdriver to open it.</span>")
```

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+			return
+	update_icon()
```
Naxodile, line 144: Probably meant to be in the actual cell code, hitting lasers with random things shouldn't need to update them.

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+	if(self_recharge)
+		return
+	//only let's you load in power cells
+	if(istype(A, /obj/item/weapon/cell))
+		. = TRUE
+		var/obj/item/weapon/cell/AM = A
+
+		if(hatch_open)
+			if(power_supply)
+				to_chat(user, "<span class='warning'>[src] already has a power cell loaded.</span>") //already a magazine here
```
Naxodile, line 123: magazine cells

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+	if(self_recharge)
+		return
+	if(hatch_open)
+		if(power_supply)
+			user.put_in_hands(power_supply)
+			user.visible_message("[user] removes [power_supply] from [src].", "<span class='notice'>You disconnect the wires and remove [power_supply] from [src].</span>")
+			playsound(loc, mag_remove_sound, 50, 1)
+			power_supply.update_icon()
+			power_supply = null
+			if(modifystate)
```
Naxodile, line 158: This sort of thing should be covered in on_update_icon()

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+				icon_state = "[modifystate][0]"
+			else
+				icon_state = "[initial(icon_state)][0]"
+		else
+			to_chat(user, "<span class='warning'>[src] is empty.</span>")
+	else
+		user.visible_message("<span class='warning'>The cell cover is closed. Use a screwdriver to open it.</span>")
+		return
+	update_icon()
```
Naxodile, line 167: Same as previous comment.

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+
+//to trigger loading cell
+/obj/item/weapon/gun/energy/attackby(var/obj/item/A as obj, mob/user as mob)
+	if(self_recharge)
```
Naxodile, line 171: Prevents all parents from having their attackby cases. Just invert and combine with the screwdriver check.

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+	if(istype(A, /obj/item/weapon/cell))
+		. = TRUE
+		var/obj/item/weapon/cell/AM = A
+
+		if(hatch_open)
+			if(power_supply)
+				to_chat(user, "<span class='warning'>[src] already has a power cell loaded.</span>") //already a magazine here
+				return
+			if(AM.maxcharge <= (max_shots*charge_cost))
+				current_user = user
```
Naxodile, line 126: Unused var.

code/modules/projectiles/guns/energy.dm
```diff
@@ -24,6 +24,15 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
 	var/recharge_time = 4
 	var/charge_tick = 0
 
+	var/current_user
```
Naxodile, line 27: Unused and not required.

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+
+	if(!..())
+		return
+	
+	if(hatch_open)
+		to_chat(user, "<span class='warning'>[src] has its cell cover still open!</span>")
+		return 0
+	return 1
+
+//Attempts to load A into src, depending on the type of thing being loaded and the load_method
```
Naxodile, line 111: Inaccurate comment.

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +98,114 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+				AM.update_icon()
+			else
+				user.visible_message("<span class='warning'>The cell size is too big for the [src]. It must be [max_shots*charge_cost] Wh or smaller.</span>")
+				return
+		else
+			user.visible_message("<span class='warning'>The cell cover is closed. Use a screwdriver to open it.</span>")
+			return
+	update_icon()
+
+//attempts to unload src. If allow_dump is set to 0, the speedloader unloading method will be disabled
```
Naxodile, line 146: As above

Naxodile (changes requested):

code/modules/projectiles/guns/energy.dm
Cheedows: Removed the redundant icon update

code/modules/projectiles/guns/energy.dm
Cheedows: Changed the comment to be more clear it is for cells

code/modules/projectiles/guns/energy.dm
Cheedows: Removed check and included it into screwdriver check

code/modules/projectiles/guns/energy.dm
Cheedows: Removed unused user variable

code/modules/projectiles/guns/energy.dm
Cheedows: Removed unused user variable

code/modules/projectiles/guns/energy.dm
Cheedows: Changed comment to be a bit more clear that it is for loading power cells

code/modules/projectiles/guns/energy.dm
Cheedows: Changed comment to be a bit more clear that it is for unloading power cells

code/modules/projectiles/guns/energy.dm
Cheedows: Removed this with update_icon() instead as it's included already

code/modules/projectiles/guns/energy.dm
Cheedows: Removed this and put it above under the clause where the power supply is removed, rather than updating icon every time unload_ammo is called

code/modules/projectiles/guns/energy.dm
```diff
@@ -89,13 +96,104 @@ GLOBAL_LIST_INIT(registered_cyborg_weapons, list())
...
+	return 1
+
+//To load a new power cell into the energy gun, if item A is a cell type and the gun is not self-recharging
+/obj/item/weapon/gun/energy/proc/load_ammo(var/obj/item/A, mob/user)
+	if(self_recharge)
+		return
+	//only let's you load in power cells
+	if(istype(A, /obj/item/weapon/cell))
+		. = TRUE
+		var/obj/item/weapon/cell/AM = A
```
CPU-Blanc, line 116:
```suggestion
/obj/item/weapon/gun/energy/proc/load_ammo(var/obj/item/weapon/cell/AM, mob/user)
	if(self_recharge)
		return
	//only let's you load in power cells
	if(istype(AM))
		. = TRUE

```
Might be me being overly picky and sorry if this is out of line, but this reassignment seems redundant rather than just checking the type defined?

Cheedows: Thanks that makes sense and tested it, I implemented this in the latest commit.

Cheedows:
Is there anything else that needs to be changed or added? Not really much else that needs to be WIP I think, with the last commit sounds for some of the actions have also been added so it should be mostly polished at this point but let me know.

Glloyd:
Pinging @Naxodile because he's been taking the lead on this one

Glloyd:
Have you heard at all from Nax? I've been super busy lately, but I haven't seen him around at all. If he doesn't swoop in to give this the all clear I'll try to give it a review later this week.

Naxodile (approved):
Looks all good.
One tiny thing you can do if you have the time, toss in some 'notice' span tags on the visible_message calls without any styling.

Glloyd:
Thanks for the review Nax! I'll get this in for tonight then.

PR #1012, opened 2021-07-12
Allows magazines to reload other magazines of the same calibre

Cheedows:
Small change that let's you reload magazines with other magazines of the same calibre.

### DIFF
--- a/code/modules/projectiles/ammunition.dm
+++ b/code/modules/projectiles/ammunition.dm
@@ -165,7 +165,7 @@
 				AM.stored_ammo -= C //should probably go inside an ammo_magazine proc, but I guess less proc calls this way...
 				count++
 		if(count)
-			user.visible_message("[user] reloads [src].", "<span class='notice'>You load [count] round\s into [src].</span>")
+			to_chat(user,"<span class='notice'>You transfer [count] round\s into [src].</span>")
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/ammunition.dm
```diff
@@ -147,6 +147,28 @@
...
+		for(var/obj/item/ammo_casing/C in AM.stored_ammo)
+			if(stored_ammo.len >= max_ammo)
+				break
+			if(C.caliber == caliber)
+				C.forceMove(src)
+				stored_ammo.Add(C)
+				AM.stored_ammo -= C //should probably go inside an ammo_magazine proc, but I guess less proc calls this way...
+				count++
+		if(count)
+			user.visible_message("[user] reloads [src].", "<span class='notice'>You load [count] round\s into [src].</span>")
```
Naxodile, line 168:
```suggestion
			to_chat(user,"<span class='notice'>You transfer [count] round\s into [src].</span>")
```

Naxodile (approved):
Having a simple add_ammo proc on magazines would be nice for this stuff, won't bully you on it though.

Glloyd:
Neat idea. Thank you Cheeds, our new guncode messiah.

PR #1013, opened 2021-07-12
HI-2521 SMG magsize halved, mags added to autolathe

ggermanicus:
They're still very strong, but like, less. Magazines are now 15 rounds from 30, and to compensate Cargo can now make more for you.


Glloyd:
taking a look at damage for the various calibres is probably something else we should do

PR #1015, opened 2021-07-13
Removes Unathi regen

ggermanicus:
Well, comments it out. 

Unathi are still Tougher Than You, but will no longer walk everything off with enough time and steak dinners, bringing them closer to humans health-wise.

I accept my fate.

### DIFF
--- a/code/modules/species/station/lizard.dm
+++ b/code/modules/species/station/lizard.dm
@@ -73,13 +73,13 @@
 		)
 	breathing_sound = 'sound/voice/lizard.ogg'
 
-	base_auras = list(
+	/*base_auras = list(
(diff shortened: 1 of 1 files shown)


Dimasw99:
Killing an Unathi is actually fairly simple and all what you have to do is chose your weaponry to what you need. Any burn damage dealing weapons like lasers wreck Unathis in two shots making them hit the floor and be easy pickings

Glloyd:
It's been a week and I haven't heard any good arguments against this. There's a good argument in favour of reworking the regen system (which has been talked about for a long time) but barring someone taking up the initiative on that, this is good for now.

PR #1016, opened 2021-07-14
adds drop pouches and holobadges to loadouts

ggermanicus:
...or it will when the mystery runtime is fixed


ggermanicus:
runtime fixed/nonexistent in the first place, bug fixed. should be good

PR #1018, opened 2021-07-18
Akimbo Guns on the Nerva: John Woo Edition

Cheedows:
Slightly modified version of the old version of /tg/'s dual wielded guns for whoever is familiar. Basically how they had it was that if you had appropriate guns in both of your hands, a single click would fire them both at the same time. This slightly modified version instead has it so that having pistol sized guns in both of your hands would fire normally on a single click, but would swap hands to your other gun and let you fire again. Basically, alternative fire akimbo guns more like the movies rather than firing them at the same time.

Features and Balance:
- To prevent crazy shit like dual wielding 2 e-guns or tasers and destroying someone, for now dual wielding is restricted to ballistic guns that are normal sized or smaller (basically most things that you would consider "pistol sized"). Both guns have to meet this requirement to dual wield.
- Otherwise, you will fire your gun as you normally would
- As of now when you fire guns akimbo, both guns will suffer an accuracy modifier that is the same as you would by one-handing a 2-handed gun (like single handing a rifle), although without the shot dispersion (with shot dispersion in testing you basically wouldn't be able to hit anything whatsoever as every shot would be off trajectory on an angle. Probably realistic to real life, but that's no fun)
- Effectively, you are trading off accuracy and convenience (reloading 2 guns is harder, you need 2 guns in the first place for this to work, and reloading a single gun is going to be a lot faster with a free hand) for faster rate of fire and access to 2 guns worth of magazine capacity
- Considering the relative rarity of ballistic pistols, the cost of having 2 at once, the accuracy malus fairly significant already added on, trade offs in reload speed, and the fact that most pistols average between 7-10 shots and 15 on the high end (so you'd effectively end up having 20-30 rounds in total, all semi-auto/single shot) it still wouldn't be as good as a lone SMG, but hopefully will end up providing some fun options to people. Open to any further balance suggestions/changes that I missed.

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -62,6 +62,7 @@
 	var/silenced = 0
 	var/accuracy = 0   //accuracy is measured in tiles. +1 accuracy means that everything is effectively one tile closer for the purpose of miss chance, -1 means the opposite. launchers are not supported, at the moment.
 	var/accuracy_power = 5  //increase of to-hit chance per 1 point of accuracy
+	var/can_dual_wield = 0
 	var/dual_wield_penalty = 4 // by default dual wielding reduces accuracy by 25%
 	var/bulk = 0			//how unwieldy this weapon for its size, affects accuracy when fired without aiming
 	var/last_handled		//time when hand gun's in became active, for purposes of aiming bonuses
@@ -149,6 +150,17 @@
 	for(var/obj/O in contents)
 		O.emp_act(severity)
 
+/obj/item/weapon/gun/proc/check_dual_wield(atom/A, mob/living/user, params)
+	var/obj/item/weapon/gun/G = user.get_inactive_hand()
+	if(can_dual_wield && G.can_dual_wield)
+		if(w_class <= ITEM_SIZE_NORMAL && G.w_class <= ITEM_SIZE_NORMAL)
+			Fire(A,user,params, dual_wield=1)
+			user.swap_hand()
+		else
+			Fire(A,user,params) //Otherwise, fire normally.
+	else
+		Fire(A,user,params)
+
 /obj/item/weapon/gun/afterattack(atom/A, mob/living/user, adjacent, params)
 	if(adjacent) return //A is adjacent, is the user, or is on the user's person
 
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/gun.dm
```diff
@@ -334,6 +351,10 @@
 	if(one_hand_penalty >= 4 && !held_twohanded)
 		acc_mod -= one_hand_penalty/2
 		disp_mod += one_hand_penalty*0.5 //dispersion per point of two-handedness
+	
+	if(dual_wield)
+		acc_mod -= dual_wield_penalty/2
+		to_chat(user, "<span class='warning'>Works!</span>")
```
Glloyd, line 357: you left in your debug text

Glloyd (changes requested):
haven't had a chance to fully review this yet, but you might need to adjust the w_class used. item size normal is w_class 3, all pistols (minus the silenced ones) are 2. 3 includes SMGs and the sawn off shotgun. I'm not sure whether this is intended or not.

Glloyd:
Did some further testing on this. I feel like some small amount of dispersion might be necessary. Maybe baymiss has just broken my brain, but it doesn't seem like you miss all that much more than you normally do. I didn't test firing the SMGs full auto with this, but you can fire akimbo, so if that's not an intentional thing, it'll need to be changed (and if it is intentional, that's a whole other balance issue)

Cheedows:
Definitely not intentional, I'll change the w_class to 2, didn't realize pistols were small not normal. I'll add some dispersion too then, maybe a smaller degree than the normal penalty for 1 handed a 2 handed gun, since in my testing with that penalty you basically couldn't hit a single thing because all the bullets would disperse on a different angle than the target.

Glloyd:
Yeah, even just a tiny tiny amount of dispersion would make it feel better while shooting, and make the accuracy tradeoff slightly more relevant. You could even add a small amount of dispersion and very slightly lower the accuracy penalty to compensate.

code/modules/projectiles/gun.dm
```diff
@@ -148,6 +150,17 @@
 	for(var/obj/O in contents)
 		O.emp_act(severity)
 
+/obj/item/weapon/gun/proc/check_dual_wield(atom/A, mob/living/user, params)
+	var/obj/item/weapon/gun/G = user.get_inactive_hand()
+	if(can_dual_wield && G.can_dual_wield)
+		if(w_class <= ITEM_SIZE_NORMAL && G.w_class <= ITEM_SIZE_NORMAL)
```
Naxodile, line 156:
```suggestion
		if(w_class <= ITEM_SIZE_SMALL && G.w_class <= ITEM_SIZE_SMALL)
```

code/modules/projectiles/gun.dm
Naxodile, line 357: May as well keep with the weird math
```suggestion
		disp_mod += dual_wield_penalty*0.5
```

Naxodile (approved):
Changes made, @Glloyd ?

Glloyd:
Seems like it addresses the issues I had back then, we'll give it a try.

PR #1025, opened 2021-08-13
[Feature] Bureaucracy comes to the Nerva

CPU-Blanc:
Bit of a big one with several features, with the general aim to bring a little more bureaucracy to the ICS Nerva, as well as a few tweaks.

### Fixed/Tweaked Behaviour:

- Email client login behaviour has been changed. Previously it would _always_ attempt to log into an account matching the account associated with your ID, making it impossible to log into a different account regardless of what you enter (would throw wrong password). Now if all fields are left blank, it will default to using your ID login. If the fields are entered however, it'll prioritise those.
- Added some missing job departments to command roles as well as missing head positions vars. This change actually makes the SO the head of Civilian, and the Quartermaster the head of Cargo, and now have access to their relevant department account login information.
- Tweaked account creation timestamp. It was the only transaction timestamp to use named months and not numbered. This has been corrected to match all other transactions.
- Payment Controller has been tweaked slightly with a fallback to stop odd behaviour breaking the payment loop completely.
- Removed Senior Scientist from the payroll. This brings them in line with other NT scientists, seeing as NT pay them and not the Nerva.
- Payroll now emails the crew their paycheck, so people actually know how much they're being paid (or if they didn't at all!). Includes dumb sponsor slogans to carry on the dumb gag. NT Scientists will still get an email.

### Added Features:

- ICS Nerva logo has been added to paperwork (Thanks to @Laniakeaaaa for the logo work!). Can be added to paperwork by using `[logo]` when the main map is the ICS Nerva, or alternatively `[nervalogo]`
- 2 new computers: The Fines Terminal, and Account Management Console.

**Account Management Console:**
Designed to solidify the Second Officer's role as the HR of the ship; this console places managing finances under the SO. It grants them the ability to overview the payroll, suspend crew pay, edit crew pay, create new accounts (both email and bank accounts, adding players to the payroll), fine crew, pay bonuses to crew, freeze crew accounts, as well as oversee department accounts' transaction history.
Unfortunately, I was about 75% through coding this before I noticed the old accounts DB console. However, that console seemed somewhat broken with money magically being spawned in with little accountability. All transactions through this version cannot go negative, and must come from an actual source; no unlimited funds. All actions also trigger an email to the effected account, and are logged internally. Happy to hear any suggestions for further balancing if needed!

**Fines Terminal:**
A small terminal located in security, allowing officers to fine crew members without taking their ID to the nearest ATM and running off with the cash. The target crew member's ID is simply scanned into the console, where officers may issue a fine of up to 1000 Th. Command personnel may bypass this limit. Funds are taken from the target account, and deposited into the ICS Nerva primary account. All fines are logged internally and create the appropriate transaction histories.

Both of these add some fun forms to sign for people who love paperwork :D

Before it is asked why the accounts management has not just been made a nano module but a dedicated computer, this is the current issue with the SO. ID changes (their main role) are easily done through just opening the program in any command console and performing it there, making it _more_ of an effort to get the SO to carry it out, than quickly doing it yourself. This way, you actually have to go to the SO's office to make changes, hopefully giving the role to the SO proper.
Happy for any feedback or changes wanted!


Glloyd:
Sorry Blanc, I've been very busy lately. I'll try to review these two over the weekend

code/controllers/subsystems/payment.dm
```diff
@@ -3,6 +3,7 @@ SUBSYSTEM_DEF(payment_controller)
 	wait = 15 SECONDS
 	var/timerbuffer = 60 MINUTES //buffer for time check
 	var/list/brand = list("Cheesie Honkers", "Men At Arms Tobacco", "Lucky Stars Cigarettes", "Carcinoma Angels Cigarettes", "Robust Coffee", "Getmore Chocolate Corp products", "4no Raisins", "SweatMax products", "Space Cola", "Space Mountain Wind", "Dr.Gibb", "ClothingLord 9000 Clothes", "NanoTrasen Personal Computers", "Rougelady Chewing Tobacco", "Robust Softdrinks products") //this is just for the jokes
+	var/list/slogan = list("Now with 18% less carcinogens!", "Lawsuit free since 2579!", "Leading doctors only moderately advise against our products!", "Now occasionally without illegal narcotics!", "Now complying with 94% of trading standards laws!", "Condemned by the Space Pope!", "Larva-tested, pupa-approved", "For external use only", "Fun for the whole family except grandma and grandpa", "Beats a hard kick in the face", "The proud result of prison labor!", "No refunds", "For the sophisticated shut-in", "Tell your parents it's educational!")
```
Glloyd, line 6: I love these so much

Glloyd:
Haven't looked through the code fully yet, but I've started doing some testing. As per your comments, the old account DB is on glloydstation, and it is super broken and exploitable, and really has always been. This seems much better. 

I do question making this a separate computer though. I get your reasoning, but I really don't like the idea of adding in two separate snowflake computers that are stuck in the SO's office and need circuitboards to be replaced. I'm not sure what the best solution is though, outside of limiting the software to SO access and putting it on the SO's computer's by default I guess. That solves the issue of it not being accessible to the captain or the FO if needed, and not being easily lost if a bomb goes off, but you still might have the issue that you brought up

CPU-Blanc:
Glad to hear it's at least somewhat passable in terms of balance, that was my main concern!

Just a quick note, that it's 1 snowflake computer in the SO's office; the fines terminal lives in the brig.

I do understand it's not desirable to have it as a fixed console, however as far as I'm aware, both captain and FO have `access_hop` on their IDs, and please do correct me if I'm mistaken, but I believe all nano software thingies can be downloaded from any console so long as their ID has access to it. Which would be worse than the issue with ID modifications, as at least that requires a card reader and someone's ID!

As far as I see it, we have 3 options currently:

1. We make it a nano software thingy, which would be able to be downloaded at any console so long as they have `access_hop` (correct me if I'm wrong). This would once again completely invalidate the need for a SO. Adjusting pay would be as easy as downloading the program, and clicking a few buttons. No need to ask the SO to do it, and more effort to do so.
2. We make it a nano software thingy but fudge in some sort of check to see if their ID also contains captain/FO access and block it if so. This would keep the software not bound to one console/circuit board, and would keep the SO in a job. Only issue is then this limits the program to _just_ the SO, which is equally as bad.
3. It's an independent console in the SO office. This presents the issues you listed above in terms of circuit boards etc (I actually haven't make it a board yet, so if we go with this option I can append that to this PR).

It's a tricky one. But I personally believe that so long as the captain or FO can simply download the software to any console, the SO will find themselves once more completely devoid of a point of existing, much like currently. As all of their duties can _easily_ be performed by anyone else, and it takes more effort and time to ask the SO to do it. ID changes? Captain/FO can quickly do these at any console they're at. Away missions? QM's joint responsibility and actually doesn't require anything special. FO often runs these too. Civilian? Captain/FO can also manage this department, if they even need it as it's a department that isn't really complex. The same goes for service.
There's currently no role the SO neatly can fit into, and such this was the general idea of this PR; to hand the role of HR/payroll to them and justify their position a little more (as well as adding some neat features!).
Would love to hear your thoughts though, as clearly it's not an easy solution

Glloyd:
Yeah, there's no perfect solution. I'm generally averse to dedicated computers, and try to avoid them wherever possible. It sounds like the fine computer is something that would be better served as a nano module, with the accounts one I don't know. The thing about that is that if the SO dies, or if there isn't an SO at all, the FO and captain need to use the SO's office to do something that they generally should be capable of doing elsewhere in a pinch. The only real solution to that is to have it as a nano-module, or to have a second computer in the bridge or the captain's office, which kind of defeats the purpose. I really think the nano module might be the best way, but then just explicitly make it the SO's job and install it by default on their computer. Our command players are generally good at that sort of thing. I dunno, @ggermanicus thoughts as the resident captain?

CPU-Blanc:
To be honest, it hadn't crossed my mind to put the fine computer as a nano module, sounds like a solid plan for me! Mobile ticketing of people with tablets could be interesting. It'll almost be like some sort of traffic warden! I'll had to do a bit of reworking to accommodate that, but it shouldn't be an issue.

The main issue I can think of with making them nano modules is currently all action logs and references are stored locally on the machine, and match the paperwork reference. Moving to a nano module would mean I'd have to scrap or re-think the terminal ID part as well as make some sort of global logs list and global var containing the current fine number. Not a big issue but something I'd need to do regardless. I haven't looked too deeply into nano software code yet, but hopefully there's some sort of global var for the program that isn't instanced I can use, rather than making some sort of new global var... 

I'm all for hearing more input on the Accounts Console! I completely get what you're saying, but to me and personal experience playing SO, in the actual game you can't rely on people just to simply not do it if it's not their role, especially when it's as simple as a couple of button presses. Currently at the moment it's hard to justify the SO's existence, other than courtesy from other command players.


As I was writing this, I just had another idea! Perhaps, like with telecoms, set a password to enter the accounts program? The SO's locker could spawn with the code as well as perhaps inside one of the captain's lockers or safes as a backup? Would at least require a little more effort than simply downloading the software?

Again, happy for thoughts! End of the day, I'll roll with whatever you judge to be best, this is just my personal opinion.

ggermanicus:
Opinion as requested:
Blanc is completely correct that it will always be easier for FO/Cap to do things themselves than ask SO, and that SO is not in a great place in terms of responsibilities as a result - making this software would just mean the easiest action is to DL it to the bridge computers and then continue running the entire ship from the Bridge with ease. Blanc knows this pain as the Bridge center console spawns with ID computer by default (and not any of the ones in the FO/SO office) and so I constantly wind up forgetting she exists and changing the ID myself... A code solution like this is probably the least painful and most rando-proof.

Also, things blowing up is good - this is SS13 and the map is made to be broken, this computer doesn't really sound like a valuable target even if it were blown up, and if someone is so DESPERATELY in need of a paycut/raise that even the computer being bombed out of existence isn't dissuasion enough, I can always just take their ID and use an ATM. Absolutely NEEDING to do this elsewhere in a pinch is resolvable by just putting the board in R&D/tech storage (works okay with helm computers, which is far more valuable).

Glloyd:
I will defer to you two on this one then. Add a circuitboard in for the accounts console (and toss one in secure tech storage for good measure), and turn the fines console into a nano module of some sort. I'll do a full review/tests once that's all in.

CPU-Blanc:
Sorry for the delay! Been a really busy day for me, and I've never worked with nano modules before. Turns out, quite a lot had to be re-written for it to function!

Account manager now has a circuitboard in secure storage, and the fines terminal is now a nano module that starts on the security preset consoles.

Slight changes to how it works as a result:

- New GLOB list to accommodate the fact the modules will now exists pretty much anywhere. Logs and fine number are now synced over all devices.
- No longer 'scans' an ID on attackby (as it isn't possible without adding some sort of passthrough proc for modular computers, which seems a little overkill). It now uses the card reader module to scan from an ID.
- No longer spawns paper at the location. The module now uses the nano printer hardware to print. No printer, no forms.
- As a result of the above, a new option has been added to the log viewer to re-print a fine. This is in case you issued a fine on a device without a printer but still want a paper copy (PDA or simple tablet for example!). Or in case Ian ate the paper copy or something.
- Due to its requirement for a card reader, I've added one to the security console preset. Just makes sense!

Sorry if there's any new bugs introduced, this has basically been an almost entire rework and I've never touched nano module code. Upon touching it, I've decided I hate it.
Initial testing by myself seems to work as intended, but I've yet to go over it with a fine comb.

Let me know if anything else needs changing @Glloyd!

Glloyd:
I did test this before I vanished for a month and I don't remember running into any issues. I say we just merge it and deal with the consequences lmao.

PR #1026, opened 2021-08-13
[Feature] Alert Controlled Gun Safes + Map Tweaks

CPU-Blanc:
A few features and tweaks to hopefully curb command sec play.

### Fixes/Tweaks:

- PvE ship combat now auto-triggers red alert. Will resume previous alert level once combat has ended. This brings it in line with the current PvP setup.
- SO and Blueshield closets have been tweaked. Thigh holster and spare headsets added to the SO's closet to bring it more in line with their role. The Deckard has been removed from the BG locker due to requests. BG is often close to the captain, so armoury access is very easy to obtain if needed. Can be added back in if needed.
- Stamps have returned! For whatever reason, only the captain and Senior Scientists had stamps... This adds stamps for the FO, SO and CoS, and adds all stamps back into the map. Each head's stamp is now in their office, as well as cargo now get their stamp and a denied stamp in the mailing office. Yay paperwork!
- SO office tweak. Brings it more in line with a paperwork focused office, adding a photocopier, shredder and filing cabinet. Losing the redundant ID and PDA boxes (they spawn in the locker anyway!)
- Gun cabinet removed and replaced with a charging station in command storage. In line with the theme of this PR, the gun cabinet has been removed and replaced with a charging station for those pesky robots. Beep boop.

### New Features:

- Alert Controlled gun safes. The main focus of this PR. This is an attempt to curb some command tenancy to play security, removing the gun cabinet in command storage (which _all_ command with access to the bridge had access to!), and replacing it with small safes in the SO, CMO, CE, and QM's offices. CoS, FO and BG have not been given one due to their easy access to the armoury.

These safes have 2 locks; one user lock which opens so long as an authorised ID is present, and an automatic lock. This automatic lock disengages once the ship reaches red alert. Once both the automatic lock and user lock are disengaged, the safe can be opened and the arms accessed. Once the ship drops from red alert, the safes will attempt to re-arm the automatic lock; these will wait for both the user lock to be re-engaged and _all_ stored weapons to be returned to the safe. This is to stop command from taking the weapons during red alert and just keeping them. Safes waiting for weapons to be returned will not only display this on the Use of Force Authorisation program (so the captain & CoS can keep tabs), but they will beep annoyingly every minute; pestering the user to return the gun! Once the auto-lock re-arms, access to the contents will once again be lost until red alert is called.

Alternatively, these safes can be emagged, allowing traitors access to their juicy contents. This will also stop the auto-lock from working in the future and remove the safe from the tracking software.

Once again, open to any suggestions or changes!


code/modules/urist/modules/shipbattles/ship_overmaps.dm
```diff
@@ -58,9 +58,10 @@
 			else
 				M.DoActivate()
 
-//	var/decl/security_state/security_state = decls_repository.get_decl(GLOB.using_map.security_state)
-//	security_state.stored_security_level = security_state.current_security_level
-//	security_state.set_security_level(security_state.high_security_level)
+	if(src == GLOB.using_map.overmap_ship)
```
Glloyd, line 61: You're a hero for realizing all my unfinished ideas. Thank you!

Glloyd:
Looks good to me, although I haven't tested it too much. Fingers crossed it doesn't conflict with #1037, but I'll deal with it if it does.

PR #1027, opened 2021-08-14
Bugfixing again

CPU-Blanc:
Just a few bugfixes that have been annoying me.

- Emergency lighting now actually works. Was caused by a var comparing itself to... itself. Then exiting if it equated itself.. lolwut
- Fixed double glass airlock construction. Caused by a incomplete build path. Airlock repairs are go!
- Allowed the crew to abandon the ship via escape pods under red alert or higher. Seemed to just be completely missing and impossible to do without admin intervention. Because a crew transfer doesn't seem quite fitting if half the ship is missing and floating into the endless void...

### DIFF
--- a/code/game/objects/structures/door_assembly.dm
+++ b/code/game/objects/structures/door_assembly.dm
@@ -56,7 +56,7 @@
 	dir = EAST
 	var/width = 1
 	airlock_type = /obj/machinery/door/airlock/multi_tile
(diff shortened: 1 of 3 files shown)


Glloyd:
The double airlock was something else I'd thought I'd fixed but only did locally. Thanks again Blanc!

PR #1029, opened 2021-08-16
Fixed Torpedo Construction

CPU-Blanc:
Fixes an issue where warheads could not be placed on an empty torpedo casing. Caused by empty casing starting with an ammo value, despite having no warhead. This also fixes the potential for empty casings to be loaded into the torpedo launcher and fired for 0 damage.


Glloyd:
Thanks Blanc. I'd fixed this locally, but I've been so busy the last couple weeks I haven't pushed it.

PR #1030, opened 2021-08-19
[Feature] Adds character slots as an option for AI holograms

CPU-Blanc:
Small PR that adds the ability to load a character's icons from a player's character slots, and use it as the AI's hologram.
The hologram will use any job or loadout gear as displayed on the character creation preview (including show/hide gear toggles). This will allow AI players to essentially create their own custom AI holograms if they so wish.

Due to how this works, a new helper has also been added: `getFullIcon()`. As `getFlatIcon()` is limited to returning an icon with a single dir; this proc returns a flat icon containing all 4 dirs, allowing the holograms to turn much like the default ones do.

Thank you to Bealdor for the idea!


Glloyd:
Looks good to me! Neat idea.

PR #1031, opened 2021-08-19
Holopad bugfixes

CPU-Blanc:
Fixes a few issues currently present with holopads plus a small tweak

- Tweaks the holocall hologram to no longer pull from the calling character's image in crew records, but from the current look of the player. Logically, this makes more sense
- Fixes a bug where voices and emotes are transmitted while waiting for a call to be answered
- Should address a random and rare bug where holopads would continue to transmit even after a call has ended. As I'm completely unable to replicate this despite repeated tests (though have experienced at least twice), a workaround fix has been introduced. During these occurrences, the calling holopad had a targetpad set, but the targetpad was unaware of that call and had no sourcepad set. This workaround checks on attempted transmission if the receiving pad is actually expecting it, otherwise terminates the call and resets the calling pad.

I'm not too happy about using a workaround, but as I'm completely unable to replicate the bug, it is the best I can do. If anyone is able to spot where the root cause is, I'll gladly update this PR to reflect that.

### DIFF
--- a/code/game/machinery/hologram.dm
+++ b/code/game/machinery/hologram.dm
@@ -201,9 +201,15 @@ For the other part of the code, check silicon say.dm. Particularly robot talk.*/
 	var/name_used = M.GetVoice()
 	var/message = get_hear_message(name_used, text, verb, speaking)
 	if(targetpad) //If this is the pad you're making the call from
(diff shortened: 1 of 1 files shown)


CPU-Blanc:
Closing this because the general holocall code is spaghetti and I've decided to just do a refactor instead.

PR #1033, opened 2021-08-21
Adds Cards Against the Galaxy to the games vending machine

ggermanicus:
Now you can actually play it without an admin spawning it.


Glloyd:
Weird that this wasn't already there

PR #1035, opened 2021-08-25
Adds a faction tab to the contract database

CPU-Blanc:
As requested c:

Faction tab that displays all factions in the sector, with a detailed view displaying standings, reputation and the faction description.
Standing text is somewhat a placeholder but can always be changed later on, should the faction subsystem get expanded.


Glloyd:
Goddamn, your UIs look so much better than anything I could do. Thank you for this!

PR #1036, opened 2021-08-28
Fixes shield bypass bug

CPU-Blanc:
Fixes a bug where all ship weapons bypass hostile shields.

This was changed to `TRUE` in #1020 . Am I missing some sort of reason for this @Glloyd or is it an oversight?
Because right now, **all** weapons except some torpedo warheads pass through shields :L


Glloyd:
Absolutely not intentional, I have no idea why I did that.

PR #1038, opened 2021-09-05
Pirates Balance Changes

CPU-Blanc:
This is gonna cause me pain later on as this **will** conflict with other bits I'm working on, but this needs to go in. I'll deal with the pain later.

PR tries to address some issues currently present in ship combat

- **Combat Computer now has a circuitboard:** Starts in secure tech storage along with the helm console circuitboard. New checks have been added on construction to prevent multiple computers linking to the same overmap ship.
- **Shield Disruptor nerfed:** This is brutal currently, nuking the shields up to 50% every 2 minutes. Even on a 8000kw input shield gen, the recharge rate can not even remotely handle it, making it impossible to keep shields even remotely up without using some powergaming min-max engine setup.
- **Pirate defenders cannot teleport onto the Nerva:** They're meant to be defenders after all. Stops ghosts curb stomping Nerva especially on low pop rounds. Hopefully this'll satisfy both those who hate player pirates, and those who love it.

Happy to discuss any issues, but these are come big ones causing headaches currently.


code/modules/urist/structures/boardingstructures.dm
```diff
@@ -30,13 +30,17 @@
 /obj/structure/boarding/shipportal/proc/teleport(atom/movable/M as mob|obj)
 	if(istype(M, /obj/effect)) //sparks don't teleport
 		return
-	else
-		var/tele_x = GLOB.using_map.overmap_ship.evac_x
-		var/tele_y = GLOB.using_map.overmap_ship.evac_y
-		var/tele_z = GLOB.using_map.overmap_ship.evac_z
+	if(istype(M, /mob/living))
```
Glloyd, line 33: make this controlled by a var so we can change this if we want to for events

Glloyd (changes requested):

Glloyd:
I don't see any other issues with the rest, thanks for this Blanc!

code/modules/urist/structures/boardingstructures.dm
CPU-Blanc: Didn't think of that! Boolean `block_pirate_teleport` has been added!

PR #1044, opened 2021-12-21
Holopads 'R' Us

CPU-Blanc:
The current holopad code has been driving me nuts with various bugs, oversights and general mess. This is a (hopefully better) refactor of the code, as well as some added features.
Changes are as follows:

- Holopad power consumption is now properly handled, and scales with each additional hologram as well as depending on normal or long ranged versions
- Code refactor should address an odd bug that would occasionally happen where holopads would continue to transmit voice, despite the call being ended. A fallback sanity check has been added.
- Holopad range limits have been tweaked to require _both_ pads to be within each other's range. Previously a long-range pad could call a normal pad even if the normal pad would not be able to dial back, this way, long-range pads can only dial other long-range pads if the standard range has been exceeded.
- Languages are now preserved through holocalls. Previously all languages were magically translated by holopads, this is no longer the case as a proc has been added to allow the pad to 'speak'.
- Emote behaviour has been changed to make more sense. Now, not _all_ emotes are broadcast to the other side. **Visible** emotes require the other party to be able to actually see the source; either their 'eye'/camera must be viewing the source (as with a standard holocall), or if the source is projecting a hologram of themselves. For example; in a standard holocall where person A has dialled person B while person C is close by A (A is projecting a hologram), A can see all visible emotes B performs while C cannot, while B can see all visible emotes only A performs (hologram) and not C. Anyone else in the room with B would also be able to see A's emotes. All **audible** emotes are transmitted regardless. This means you will no longer get visible emotes for random stuff the player cannot actually see (looking at you, sudden acceleration messages!).
- Added conference calls. This is an extra option that appears while starting a holocall. If selected, both holopads work in a phonecall loudspeaker sort of ways, where **everyone** in a 2 tile radius (default) of each holopad is transmitted to the other side in their relative locations. AI's can still use the pads while this mode is active, however currently they won't be able to hear or speak to the other end of the call.

Below are a couple of screenshots highlighting the conference feature. The location of the mobs relative to the pad would match either side (as well as direction), just in this example the mobs had moved between shots. For an added bonus, animals and simplemobs are also captured for fun! Be careful if using this on the Trajan however, as long range pads use more power, and so do additional projected holograms!
image
image

Sorry if there's any issues, this code has been sitting 70% done for a while and I've only recently come back to it, but as far as I've tested, it all seems to work as intended.

### DIFF
--- a/code/game/machinery/hologram.dm
+++ b/code/game/machinery/hologram.dm
@@ -19,13 +19,20 @@ Possible to do for anyone motivated enough:
 	Itegrate EMP effect to disable the unit.
 */
 
+/*
+Revised by CPU_Blanc.
+-Multiple hologram users at the same time added (conference calls)
+-Power usage reworked to scale with additional holograms
+-Languages are now preserved
+-Both holopads must be within range of eachother's max range to connect
+-Visible emotes now only pass through the pad if the source is visible to the user (Either as a hologram, or if the player's eye is in that location)
+*/
+
 
 /*
  * Holopad
  */
 
-#define HOLOPAD_PASSIVE_POWER_USAGE 1
-#define HOLOGRAM_POWER_USAGE 2
 #define RANGE_BASED 4
 #define AREA_BASED 6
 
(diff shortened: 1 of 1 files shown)


Glloyd:
As mentioned on discord, I was really sick with COVID, but I want to get back to this and SS13, I'll try to give this a look through this week!

Glloyd:
I'm sorry, I just haven't had the time to look through this one lately. Perhaps if @Naxodile has the time he can take a quick look through, otherwise I might not get to this until next week.

Naxodile (approved):
Looks good to me. I didn't find any bugs in testing besides the unrelated issue that holograms don't copy their subject's transforms. Might look a bit weird in some conference calls, but that's pretty rare.

Glloyd:
Thanks Nax! Been super busy lately, so I really appreciate this.

PR #1046, opened 2022-02-07
Latejoins announced on common by default

ggermanicus:
This makes the game a little more alive-feeling instead of being surprised by new random underlings.


Glloyd:
Thanks, this has been bothering me.

CPU-Blanc:
I don't meant to nit-pick or anything...

But wouldn't it have made sense to simply change line 115 from
`AnnounceArrivalSimple(character.real_name, rank, join_message, get_announcement_frequency(job))`
to `AnnounceArrivalSimple(character.real_name, rank, join_message, "Common")`
rather than commenting out the contents of the `get_announcement_frequency()` proc? Otherwise we're going through the hassle of calling a different proc and passing a datum argument just to return a single string? Seems a little wonk to me XD

PR #1047, opened 2022-02-09
Bugfixes and steal objective modernization

ggermanicus:
Steal objective targets are now suited to the Nerva. Account card added as a steal objective as well. Captain gets a spare jumpsuit in their locker to facilitate the theft of it.
Medal/account lockbox now properly protects its contents
Account card links to money in fewer lines
Message that broadcasted to everyone in sight when you messed up changing a cell now only tells you that you messed up
Changed the worst of the writing in job blurbs
NT labcoat and scientist jumpsuit works now
Errant pixel on telebaton that haunted me removed
and Wizard no longer spawns with an extra useless box.


code/game/gamemodes/objective.dm
```diff
@@ -385,17 +385,18 @@ datum/objective/steal
 	var/target_name
 
 	var/global/possible_items[] = list(
-		"the captain's antique laser gun" = /obj/item/weapon/gun/energy/captain,
+		"the captain's antique revolver" = /obj/item/weapon/gun/projectile/revolver/coltsaa,
```
Glloyd, line 388: I'm still generally trying to keep compatibility with glloydstation. So in this case, I think it makes more sense (in general) to add a file somewhere in the nerva folder and overwrite these definitions there. We've been moving towards having most things initialized on a per-map basis anyways, so it makes more sense for these to be done like that too.

Glloyd:
Everything else looks good, so if you can define objectives for Nerva in a separate file, we can get this merged.

Glloyd:
I meant that specific list, not the whole file...

That's my bad, I should've been more clear on that, sorry. My line of thinking was just to overwrite that possible_items list var on a per-map basis, not to delete the original file and duplicate the code between the maps.

Glloyd:
Looks good! Thanks Grant.

PR #1053, opened 2022-03-02
Bugfixes + Pirate balance changes

CPU-Blanc:
This PR was originally meant to be a simple balance pass, but I ended up falling down a rabbit hole of bugfixes along the way...

### Changes:
**Balance Changes**
- Pirate Ion weapon cooldowns increased by 4 seconds
- Pirate shield regen rate slowed by 2 seconds
- Introduction of `recovery_debt` to mimic Nerva shields. Pirate shields will now need to recharge a certain amount before they block projectiles again (similar to shield generators' 'Recovering' stage). Amount regenerated is not lost, shields are simply not effective until this threshold has passed. Set the amount required using `recovery_threashold` on each shield module. Eyeballed these numbers for now to aim for at least 1 regen cycle without shields coming online
- Slight buff to the light laser cannon. Taking it from an anti-shield DPS of 13.33 to 14.66
- Shieldwall generators now evenly split the energy load between both units, rather than picking a unit at random; sometimes causing extremely asymmetrical loads and shutting a single unit down while the other is untouched.
- Buff to the bridge shieldwall max draw, see below in bugfixes for reason

**Bugfixes/Tweaks**
- Fixed a bug present in pirate ship module activation causing all modules to not activate once a single module was destroyed. This was caused by a pesky `return` in the for loops during module activation rather then skipping the broken module with `continue`
- Fixed a runtime caused by passing a turf to the shieldwall's `New()` proc rather than its expected generator object type, causing multiple issues with load down the line and null variable calls.
- Tweaked shieldwall generators to include their default power usage into the powernet calculation rather than after, allowing stored power to actually reach their maximum rather than max - idle usage
- As a result of fixing the incorrect `New()` call, shield loads are now properly working. Due to the large distance between the bridge's shieldwall generators, there is not enough max draw to facilitate the initial generation of the shields. As a result, the max power draw for these units was increased to compensate. Alternative solution is a decrease of the generation cost, though this would be global. Hopefully this won't be too much of an issue as this'll simply put more strain on the powernet

Any issues or ideas for further changes, let me know!

### DIFF
--- a/code/modules/urist/modules/shipbattles/overmapships/components/components.dm
+++ b/code/modules/urist/modules/shipbattles/overmapships/components/components.dm
@@ -32,21 +32,30 @@
 	var/recharging = 0 //are we waiting for the next recharge delay?
 	var/recharge_delay = 5 SECONDS //how long do we wait between recharges
 	var/overcharged = FALSE //only for stations, we stop torpedos from doing hull damage. they can still hurt components though
+	var/recovery_threashold = 0	//The amount of total recharge needed to recover the shields once offline. Ie a value of 100 requires 100 shield strength to be recharged before the shields recover and block attacks
+	var/recovery_debt = 0 //Amount of recharge remaining before shields are online
 
 /datum/shipcomponents/shield/DoActivate()
 	if(!broken && !recharging)
-		if(mastership.shields <= strength)
-			mastership.shields += recharge_rate
-			if(mastership.shields >= strength)
-				mastership.shields = strength
+		if(recovery_debt)
+			if(recovery_debt >= recharge_rate)
+				recovery_debt -= recharge_rate
+			else
+				mastership.shields = min(recovery_threashold + (recharge_rate - recovery_debt),strength)	//Amount required to recharge + anything left over
+				recovery_debt = 0
+		else if(mastership.shields < strength)
+			mastership.shields = min(mastership.shields + recharge_rate, strength)
 
-			recharging = 1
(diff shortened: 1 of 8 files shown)


Glloyd:
There's a few minor quibbles I have with some of the balancing stuff, I'll run through it tomorrow and give more feedback, but this looks good besides that.

code/modules/urist/modules/shipbattles/overmapships/components/components.dm
```diff
@@ -59,56 +68,64 @@
 	strength = 750
 	health = 200
 	recharge_rate = 75
-	recharge_delay = 10 SECONDS
+	recharge_delay = 12 SECONDS
```
Glloyd, line 71: I did lower the recharge rate on this lately, and with the buff to the starting laser I think also changing the recharge_delay might be a bit much. I'd go with one or the other, either the laser buff or the delay change. If you stick with the delay change, remove it for everything below freighter, because those ones are either for event ships or for alien ships/stations which are their own special beasts.

Glloyd (changes requested):
Finally had time to go through this. As I said, most of this is good, I just have some quibbles with the recharge rates. When I first looked at this I thought I had more issues but it's really only the one thing that stood out to me this time around.

CPU-Blanc:
Not a problem! Reversed the nerf to shield recharge rates

Glloyd:
Thanks! We'll see how it plays and we can always nerf them in the future if need be.

PR #1055, opened 2022-03-07
Changes spare PDAs to spawn off

CPU-Blanc:
Small PR. Changes boxes of spare PDAs to spawn them offline rather than on. A box of 5 online PDAs actually uses a stupid amount of power and adds necessary drain on APCs which can be critical during lowpop with no engineers. See the SO's office APC during roundstart :P


Glloyd:
This is a weird one, good find.

PR #1057, opened 2022-03-14
Nerva-specific sechud icons

ggermanicus:
Accommodates every job I can reasonably think of currently, although alt-title HUD changes don't work and haven't worked in the past. Bay has moved arrest HUDs and antag HUDs to other places now.

For actual new things, this:
Gives the FO a 'new' (torch pilot icon recolored) icon instead of HoP 2.0.
Fixes bodyguard and science sechud icons.


Glloyd:
Thanks Grant!

PR #1063, opened 2022-03-22
Multi-Z Pathfinding

CPU-Blanc:
This is a bit of a followup to the addition of navbeacons on the Nerva.
Multi-Z pathfinding! Because without it, Beepsky is too stupid to understand stairs :c

**Changes**

- Added railings to pathfinding obstacles. Beepsky will no longer try to ram through a solid railing, then question his life choices.
- Added bolted check to airlock pathfinding. Beepsky (and other NPCs), are now able to recognise bolted airlocks as obstacles and attempt to path around it! _Clever girl_
- Added `AStarWithZ()` proc. This is the meat of this PR. This a new proc which utilises the currently existing pathfinding code to allow for z level changes to be taken into account. To accommodate this, a new global list has been created, `GLOB.zlevel_transitions`. This is indexed by Z level (string), and returns a list of turfs or stairs that are able to facilitate z-level changes from that z level. At the moment, this is only set to include stairs (and thus open spaces above them), but future expansions could include ladders?

This proc isn't necessarily the most efficient or clever way to handle z transformations, but it does at least allow for dynamic navigation to different floors, something currently impossible. If someone big-brained wants to come along and write some super nerd script to handle all those nodes in a 3 dimensional space, be my guest!
This currently works by stringing along the already available `AStar()` pathfinding proc. The NPC will first identify what z levels it needs to cross to get to its target z level by working its way upwards, then downwards. Then on a level-by-level basis, it'll calculate a valid path to each set of stairs that brings it closer to its target. Finally, it'll insert and append the navigation of it's current turf to the first stair, and from the last stair to its final destination. The end result is a string of paths to the first stairs, between all the other stairs, and then its final destination.

At the moment, the only AI that calls this proc are types of `/mob/living/bot`, but it could easily be expanded to other mobs that follow players.

Again, if someone else has a more elegant and efficient solution then by all means correct or disregard this code!

### DIFF
--- a/code/modules/mob/living/bot/bot.dm
+++ b/code/modules/mob/living/bot/bot.dm
@@ -302,7 +302,9 @@
 /mob/living/bot/proc/startPatrol()
 	var/turf/T = getPatrolTurf()
 	if(T)
-		patrol_path = AStar(get_turf(loc), T, /turf/proc/CardinalTurfsWithAccess, /turf/proc/Distance, 0, max_patrol_dist, id = botcard, exclude = obstacle)
+		busy = 1	// Multi-z can take a bit. This'll make sure it isn't called twice while it runs
+		patrol_path = AStarWithZ(get_turf(loc), T, /turf/proc/CardinalTurfsWithAccess, /turf/proc/Distance, 0, max_patrol_dist, id = botcard, exclude = obstacle)
+		busy = 0
 		if(!patrol_path)
 			patrol_path = list()
 		obstacle = null
@@ -334,7 +336,7 @@
 	return
 
 /mob/living/bot/proc/calcTargetPath()
-	target_path = AStar(get_turf(loc), get_turf(target), /turf/proc/CardinalTurfsWithAccess, /turf/proc/Distance, 0, max_target_dist, id = botcard, exclude = obstacle)
+	target_path = AStarWithZ(get_turf(loc), get_turf(target), /turf/proc/CardinalTurfsWithAccess, /turf/proc/Distance, 0, max_target_dist, id = botcard, exclude = obstacle)
 	if(!target_path)
 		if(target && target.loc)
 			ignore_list |= target
@@ -346,7 +348,7 @@
 	if(!path.len)
 		return 0
 	var/turf/T = path[1]
-	if(get_turf(src) == T)
+	if(get_turf(src) == T || (istype(path[1], /turf/simulated/open) && get_turf(src) == path[2]))
 		path -= T
 		return makeStep(path)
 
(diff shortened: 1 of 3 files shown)


Glloyd:
> Again, if someone else has a more elegant and efficient solution then by all means correct or disregard this code!

@Naxodile Might be a good idea to have a second pair of eyes on this one in this regard.

scrdest:
Hi @Glloyd, you messed up your summoning circle, so here I am. A _tiny bit_ rusty on DM, so slap me if I'm having a senior moment, but I know my Astars.

This is waaaaaay overkill. You can easily get this functionality with existing Astar. It's not obvious, because this code uses callbacks.

Look at this line: https://github.com/UristMcStation/UristMcStation/blob/master/code/procs/AStar.dm#L96 (unfortunately, cannot highlight it in the diffs directly). 

This line calls an arbitrary proc that generates a list of neighbors for the current location. This proc is actually a _parameter_ to this implementation of Astar (3rd one, to be precise). To get a 3D Astar, all we need to do here is pass in a 3D-aware `adjacent()` proc implementation and a 3D-aware `dist()` proc for the 4th arg.

Hell, why stop there? If my ancient event mobs are still in here, we could pass in a pair of procs that returns adjacent locations in 4D space for teleporting mobs or something to stop them from hopping around at random, if we wanted to. This is a surprisingly nice, mostly generic Astar implementation!

I can take a stab at this later, but I'd need to set the whole BYOND toolchain up again first.

code/modules/mob/living/bot/bot.dm
```diff
@@ -302,7 +302,9 @@
 /mob/living/bot/proc/startPatrol()
 	var/turf/T = getPatrolTurf()
 	if(T)
-		patrol_path = AStar(get_turf(loc), T, /turf/proc/CardinalTurfsWithAccess, /turf/proc/Distance, 0, max_patrol_dist, id = botcard, exclude = obstacle)
+		busy = 1	// Multi-z can take a bit. This'll make sure it isn't called twice while it runs
+		patrol_path = AStarWithZ(get_turf(loc), T, /turf/proc/CardinalTurfsWithAccess, /turf/proc/Distance, 0, max_patrol_dist, id = botcard, exclude = obstacle)
```
scrdest, line 306: To illustrate the more elegant approach (mockup, but the distance procs should work):

```suggestion
/turf/proc/3dCardinalTurfsWithAccess(id_card)
    return all_adjacent_turfs_in_3d // somehow
    
/turf/proc/Euclidean3dDistance(turf/endloc)
   // equivalent to the current Distance proc metric, except for 3D
   var/euclid_dist =(src.x - t.x) * (src.x - t.x) + (src.y - t.y) * (src.y - t.y) + (src.z - t.z) * (src.z - t.z)
   euclid_dist *= (pathweight+t.pathweight)/2
   return euclid_dist
    
/turf/proc/Manhattan3dDistance(turf/endloc)
   // just to show off that we can use multiple metrics; AKA taxicab distance
   var/manhattan_dist =abs((src.x - t.x)) + abs((src.y - t.y)) + abs((src.z - t.z))
   manhattan_dist *= (pathweight+t.pathweight)/2
   return manhattan_dist
   
		patrol_path = AStar(get_turf(loc), T, /turf/proc/3dCardinalTurfsWithAccess, /turf/proc/Euclidean3dDistance, 0, max_patrol_dist, id = botcard, exclude = obstacle)
```

Replace the procs like this for anything you want to be 3D-capable.

CPU-Blanc:
I knew someone big-brained would show up and point out a _much_ more elegant solution!
Thanks for the pointers @scrdest ! Somehow completely overlooked the adjacent proc within AStar.

For the sake of keeping things clean, I'll be closing this PR and opening a new one utilising the methods kindly posted above. Thanks for the input again!

PR #1062, opened 2022-03-22
ICS Nerva: The Remappening (And bugfixing)

CPU-Blanc:
Oh boy. This is a biggie.
This originated as some idle chat with players during deadpop, and turned into a big of a rework. Some controversial changes in here I'm sure, but happy to revert or edit any changes. Thought it'd at least be an idea to put these designs out there, whether they're taken up or not!

### Main Map Changes

**Mailing Office**

The mailing office has been moved south to allow for a public lobby, with a public use orders console. This was requested by cargo players so other players can drop by and put their orders in, rather than shouting them over the radio like uncultured heathens. A side-effect is the cargo locker room has been moved to the lower deck to replace the unused storage room.
Screenshot from 2022-03-22 01:22:10

**Bar/Kitchen**

The bar and kitchen have been merged into a single area that serves both food and drinks. Currently, the cafeteria is woefully underused and a bit of a waste of space; players only ever seem to use the bar. This will hopefully give the chef a bit more to do, instead of splitting the possible 'customers' up. The area that was once the cafeteria has been replaced by the holodeck. Nice and central, and no longer tucked away to be forgotten. Being outside sec gives the holo-court a little more plausibility, as well as being closer to medical for those out-of-hand boxing matches. _It is also in a more central location to emag and spawn carps and cause utter chaos_. A result of this is the kitchen has had a slight extension (as well as all holodeck maps), so now they have a serving hatch to bake doughnuts and throw at security.
Screenshot from 2022-03-22 01:07:06

**AI Upload**

As the holodeck has been moved, after discussion, the most logical thing we could think to take its place was AI Upload. It may seem like a bit of an odd move, however, the current AI Upload is _far_ too close to command, and _far_ too easy for them to reset; making subverting the AI as an antag virtually impossible. AI gains a little extra security with thicker walls and the dispenser, antags hopefully get the ability to entertain the thought of AI subversion.
Screenshot from 2022-03-22 01:04:15

**Deck 4 Command**

[...]

PR #1064, opened 2022-03-22
More Bugfixing/Tweaks

CPU-Blanc:
**Changes**

- Adds detergent to the lavatory vending machine next to the washing machines. Washing machines will not work at all without them, and there's no way currently to get them on the Nerva.
- Adds combat channel use to ship combat. All combat feedback (shots missed, hit etc) now go to the combat channel to avoid command channel spam. (Will require the changes to telecoms introduced in #1062 )
- Fixed a missing weakref check in mobs' friend lists. This would cause space carp summoned by an emagged holodeck to attack the emagger, which was not the intention. I thought we were friends?
- Fixed (see: fudged) odd NPC behaviour on the NT station. Currently if you're an antag (faction changed) and you visit the NT station; you will be mobbed, followed, and subsequently nuzzled to death. This patch forces all NPC that spawn on the NT station to set their stance to `HOSTILE_STANCE_TIRED`, which prevents their targeting loops from happening. This is the best I could come up with, without forcing global changes that might have unwanted knock-on effects. NT NPCs are now tree-hugging pacifists.
- Removed the mysterious second beach program from the holodeck program list, which was so secretive it forgot to exist.

### DIFF
--- a/code/modules/urist/gamemodes/scom/civilian_mobs.dm
+++ b/code/modules/urist/gamemodes/scom/civilian_mobs.dm
@@ -45,11 +45,6 @@
 	icon_living = "gunman"
 	icon_dead = "gunman_dead"
 
-/mob/living/simple_animal/hostile/scom/civ/combat/police/Initialize()
-	. = ..()
-	var/area/A = get_area(src)
-	if(A?.type == /area/spacestations/nanotrasenspace)
-		stance = HOSTILE_STANCE_TIRED //Prevents any targetting/agression for NPC's on the NT station
 
 /mob/living/simple_animal/hostile/scom/civ/combat/mil
 	name = "soldier"
--- a/code/modules/urist/gamemodes/scom/civilian_mobs.dm
+++ b/code/modules/urist/gamemodes/scom/civilian_mobs.dm
@@ -45,6 +45,12 @@
 	icon_living = "gunman"
 	icon_dead = "gunman_dead"
 
+/mob/living/simple_animal/hostile/scom/civ/combat/police/Initialize()
+	. = ..()
+	var/area/A = get_area(src)
+	if(A?.type == /area/spacestations/nanotrasenspace)
+		stance = HOSTILE_STANCE_TIRED //Prevents any targetting/agression for NPC's on the NT station
+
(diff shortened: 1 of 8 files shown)


Glloyd:
You will need to add some way for NPCs to change their stance when attacked, because at the moment the faction system only really applies to the crew of the Nerva, or anyone with faction = "neutral"

code/modules/urist/modules/newtrading/NPC/NPC_types/colonist.dm
```diff
@@ -115,6 +115,12 @@
 		speak += "I saw a mudcrab the other day. Vile creatures they are." //the chances of this being heard are so small, but I had to put it in.
 	..()
 
+/mob/living/simple_animal/hostile/npc/colonist/Initialize()
+	. = ..()
+	var/area/A = get_area(src)
+	if(A?.type == /area/spacestations/nanotrasenspace)
```
Glloyd, line 121: I really don't like this specific area check. That aside, I think this can be addressed in other ways along with some other NPC hostility issues. I'd say revert this chunk here for now, and I'll merge the PR and we can deal with it in the future.

Glloyd (changes requested):
If you just revert that bit I can merge this, and we can deal with it in a different way in the future.

code/modules/urist/gamemodes/scom/civilian_mobs.dm
```diff
@@ -45,6 +45,12 @@
 	icon_living = "gunman"
 	icon_dead = "gunman_dead"
 
+/mob/living/simple_animal/hostile/scom/civ/combat/police/Initialize()
+	. = ..()
+	var/area/A = get_area(src)
+	if(A?.type == /area/spacestations/nanotrasenspace)
```
Glloyd, line 51: Same for this actually. We shouldn't be checking specific areas like this, especially in this way, which will cause compile errors on glloydstation and any map that doesn't load that awaymap.

CPU-Blanc:
No worries, was a bit of a fudge fix. Reverted the NT station checks. I guess I'll have to dive into faction targeting code later down the line

PR #1065, opened 2022-03-23
Multi-z path-finding support + Bot path-finding tweaks/fixes

CPU-Blanc:
And here it is! This time done properly. This PR adds multi-z path finding support to the `AStar()` proc as well as addresses some issues currently present in bot pathfinding.

**Changes:**
- New multi-z aware adjacent and distance procs for use with `AStar()`. Currently only implemented in the bots, but could very easily be extended to any other mobs that navigates via `AStar()`
- Adds additional awareness to existing path-finding; such as bolted, powerless, and welded airlocks. AI's will try to avoid these airlocks now and navigate around them
- Tweaked the pathweight of turfs with airlocks on them. Bots will now _slightly_ prefer routes without airlock tiles. This was to combat bots using what are _technically_ shorter routes, that were actually longer due to airlock opening times.
- Set `max_frustration` to the most common bots. This'll allow them to 'give up' and start path-finding again if they get stuck
- Added frustration checks to patrol navigation. Currently bots only get frustrated when navigating to a target. However, there is an issue if bots get a target while on patrol and thus move off of the existing patrol route; once they lose interest in their target, they'll 'dumb' navigate back to the last point on the patrol path by calling `step_towards()`. This leads to many bots simply ramming walls over and over again, and getting trapped in maintenance. Now they'll get frustrated on each failed move to the patrol path, and recalculate using `AStar()` when they're tired of rubbing their metallic cheeks against maintenance.
- Made path-finding aware of railings. Previously they just.. didn't understand them and would walk into them and get stuck
- Dissallowed bot path-finding from using lift airlocks unless they are open. Now that path-finding is z-level aware, bots had the tendency to hurl themselves down the elevator shaft as a shortcut to the bottom deck. As amusing as this is, this could cause the elevator to get stuck as the airlock would be left open, as well as being a massive OSHA violation...

A couple examples of the path-finding changes in action: 

Beepsky working around a wall by taking the stairs either side

https://user-images.githubusercontent.com/54823378/159814095-8b23d30b-9444-49fd-b1c2-e8d139c9bd65.mp4

Beepsky avoiding a bolted airlock and path-finding around

https://user-images.githubusercontent.com/54823378/159814321-09aedaf5-f52b-4689-a32d-2198525fbc02.mp4

Once again, big thanks to @scrdest for the pointers in my last attempt. I wouldn't have noticed this much more (hopefully) elegant implementation without them!

There's also the option in future to potentially path-find using ladders and open space (stalker space carp, anyone?), though currently this is probably overkill and would require code on the mob side of things to handle activating the ladder etc

Any issues, as always, let me know!

### DIFF
--- a/code/game/machinery/doors/door.dm
+++ b/code/game/machinery/doors/door.dm
@@ -103,8 +103,6 @@
 	set_density(0)
 	update_nearby_tiles()
 	QDEL_NULL(dummy)
-	var/turf/T = get_turf(src)
-	T.pathweight = initial(T.pathweight)
 	. = ..()
 
 /obj/machinery/door/Process()
--- a/code/game/machinery/doors/door.dm
+++ b/code/game/machinery/doors/door.dm
@@ -103,6 +103,8 @@
 	set_density(0)
 	update_nearby_tiles()
 	QDEL_NULL(dummy)
+	var/turf/T = get_turf(src)
+	T.pathweight = initial(T.pathweight)
 	. = ..()
 
 /obj/machinery/door/Process()
--- a/code/modules/mob/living/bot/bot.dm
+++ b/code/modules/mob/living/bot/bot.dm
@@ -305,7 +305,7 @@
 /mob/living/bot/proc/startPatrol()
 	var/turf/T = getPatrolTurf()
 	if(T)
-		patrol_path = AStar(get_turf(loc), T, /turf/proc/CardinalTurfsWithAccessWithZ, /turf/proc/Manhatten3dDistance, 0, max_patrol_dist, id = botcard, exclude = obstacle)
+		patrol_path = AStar(get_turf(loc), T, /turf/proc/CardinalTurfsWithAccessWithZ, /turf/proc/Manhattan3dDistance, 0, max_patrol_dist, id = botcard, exclude = obstacle)
 		if(!patrol_path)
 			patrol_path = list()
 		obstacle = null
@@ -493,15 +493,23 @@
 
 /turf/proc/Euclidean3dDistance(turf/t)
 	var/euclid_dist = sqrt(Square(src.x - t.x) + Square(src.y - t.y) + Square(src.z - t.z))
-	var/currentPathweight = ((locate(/obj/machinery/door) in src) && pathweight == 1) ? 2 : pathweight
-	var/targetPathweight = ((locate(/obj/machinery/door) in t) && t.pathweight == 1) ? 2 : t.pathweight
+	var/currentPathweight = src.pathweight
+	var/targetPathweight = t.pathweight
+	if(locate(/obj/machinery/door) in src)
+		currentPathweight += 1
+	if(locate(/obj/machinery/door) in t)
+		targetPathweight += 1
 	return euclid_dist * ((currentPathweight+targetPathweight)/2)
 
-/turf/proc/Manhatten3dDistance(turf/t)
-	var/manhatten_dist = abs(src.x - t.x) + abs(src.y - t.y) + abs(src.z - t.z)
-	var/currentPathweight = ((locate(/obj/machinery/door) in src) && src.pathweight == 1) ? 2 : src.pathweight
-	var/targetPathweight = ((locate(/obj/machinery/door) in t) && t.pathweight == 1) ? 2 : t.pathweight
-	return manhatten_dist * ((currentPathweight+targetPathweight)/2)
+/turf/proc/Manhattan3dDistance(turf/t)
+	var/manhattan_dist = abs(src.x - t.x) + abs(src.y - t.y) + abs(src.z - t.z)
+	var/currentPathweight = src.pathweight
+	var/targetPathweight = t.pathweight
+	if(locate(/obj/machinery/door) in src)
+		currentPathweight += 1
+	if(locate(/obj/machinery/door) in t)
+		targetPathweight += 1
+	return manhattan_dist * ((currentPathweight+targetPathweight)/2)
 
 //NORTH, SOUTH, EAST, WEST, NORTHEAST, NORTHWEST, SOUTHEAST, SOUTHWEST + Traversal up and down stairs
 /turf/proc/AllDirTurfsWithAccessWithZ(var/obj/item/weapon/card/id/ID)
--- a/code/modules/mob/living/bot/bot.dm
+++ b/code/modules/mob/living/bot/bot.dm
@@ -444,12 +444,15 @@
 		else if(DirBlockedWithAccess(xStep,adir&(NORTH|SOUTH),ID))
 			return 1
 
+		var/blocked = FALSE
 		for(var/obj/O in yStep)	//Both paths free? Check them for objects. So long as one isn't blocked, we can move
 			if(O.density && !istype(O, /obj/machinery/door) && !(O.atom_flags & ATOM_FLAG_CHECKS_BORDER))
-				for(var/obj/OB in xStep)
-					if(OB.density && !istype(OB, /obj/machinery/door) && !(OB.atom_flags & ATOM_FLAG_CHECKS_BORDER))
-						return 1
-				return 0
+				blocked = TRUE
+				break
+		if(blocked)
+			for(var/obj/O in xStep)
+				if(O.density && !istype(O, /obj/machinery/door) && !(O.atom_flags & ATOM_FLAG_CHECKS_BORDER))
+					return 1
 		return 0
 
 	if(DirBlockedWithAccess(A,adir, ID))
--- a/code/modules/mob/living/bot/bot.dm
+++ b/code/modules/mob/living/bot/bot.dm
@@ -296,13 +296,16 @@
 	return 1
 
 /mob/living/bot/proc/handlePatrol()
-	makeStep(patrol_path)
+	if(makeStep(patrol_path) && frustration)
+		frustration = 0
+	else if(max_frustration)
+		++frustration
 	return
 
 /mob/living/bot/proc/startPatrol()
 	var/turf/T = getPatrolTurf()
 	if(T)
-		patrol_path = AStar(get_turf(loc), T, /turf/proc/CardinalTurfsWithAccess, /turf/proc/Distance, 0, max_patrol_dist, id = botcard, exclude = obstacle)
+		patrol_path = AStar(get_turf(loc), T, /turf/proc/CardinalTurfsWithAccessWithZ, /turf/proc/Euclidean3dDistance, 0, max_patrol_dist, id = botcard, exclude = obstacle)
 		if(!patrol_path)
 			patrol_path = list()
 		obstacle = null
@@ -334,7 +337,7 @@
 	return
 
 /mob/living/bot/proc/calcTargetPath()
-	target_path = AStar(get_turf(loc), get_turf(target), /turf/proc/CardinalTurfsWithAccess, /turf/proc/Distance, 0, max_target_dist, id = botcard, exclude = obstacle)
+	target_path = AStar(get_turf(loc), get_turf(target), /turf/proc/CardinalTurfsWithAccessWithZ, /turf/proc/Euclidean3dDistance, 0, max_target_dist, id = botcard, exclude = obstacle)
 	if(!target_path)
 		if(target && target.loc)
 			ignore_list |= target
@@ -355,7 +358,7 @@
 /mob/living/bot/proc/resetTarget()
 	target = null
 	target_path = list()
-	frustration = 0
+	//frustration = 0
 	obstacle = null
 
 /mob/living/bot/proc/turn_on()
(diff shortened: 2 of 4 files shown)


code/modules/mob/living/bot/bot.dm
```diff
@@ -435,11 +438,63 @@
 		if(D.dir == dir)		return 1
 
 	for(var/obj/machinery/door/D in loc)
+		if(loc.pathweight == 1)
+			loc.pathweight = 2	//Slight bias in pathfinding away from airlocks
```
scrdest, line 442: Err, if I'm reading this right, you're *updating* the pathweight in-place. We _probably_ shouldn't mess with the map variables or Weird Badness might happen.

Would be much cleaner to modify the Distance proc used - instead of reading the pathweight off of the turf directly, read it into a var, put in type-check for doors and add/multiply the base cost there.

code/game/machinery/doors/door.dm
```diff
@@ -103,6 +103,8 @@
 	set_density(0)
 	update_nearby_tiles()
 	QDEL_NULL(dummy)
+	var/turf/T = get_turf(src)
+	T.pathweight = initial(T.pathweight)
```
scrdest, line 107: re: other comment about messing with pathweight - changing the Distance proc also means you don't have to do stuff like this. 

There's probably cases where initial() will do something invalid here (e.g. turfs loaded on runtime using spawners - initial() will set them to the _compile-time_ value, not the value the spawner created them with).

code/modules/mob/living/bot/bot.dm
CPU-Blanc: No excuses for this one, slipped by me. Meant to return to it later and completely forgot! I'll incorporate it into the distance proc

code/modules/mob/living/bot/bot.dm
```diff
@@ -435,11 +438,63 @@
...
+		if(dir == D.dir)		return 1
 	return 0
+
+
+//Multi-Z AStar Procs
+
+/turf/proc/Euclidean3dDistance(turf/t)
+	var/euclid_dist = sqrt(Square(src.x - t.x)+Square(src.y - t.y)+Square(src.z - t.z))
+	euclid_dist *= (pathweight+t.pathweight)/2
+	return euclid_dist
```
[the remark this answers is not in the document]
CPU-Blanc: More importantly, I just noticed we're only returning _cardinal_ turfs with the adjacent proc anyway, which defeats the whole point of using Euclid... Whoops.

Testing how the bot handles passing all dirs to adjacent with Euclid. If it's wonky, I'll change it to Manhattan + cardinal.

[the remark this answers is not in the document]
scrdest: It shouldn't be _that_ big of a deal to use Euclid if the adjacents are cardinal - for cardinally adjacent turfs, Euclid == Manhattan. 

The difference only pops up for diagonals (`Euclid((1,1), (0,0)) == sqrt(2)` while `Manhattan((1,1), (0,0)) == 2)` - so Manhattan is biased towards straight lines where possible.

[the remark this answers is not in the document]
scrdest: Also, in case it comes up later - spessmobs that have full diagonal motion in 3D (e.g. could work for spesscarp) are most accurately measured in yet another way: `max(abs(x2-x1), abs(y2-y1), abs(z2-z1))` (Chebyshev, or chessboard, distance; works for 2D diagonals if you drop the Zs). 

In practice, though, this doesn't matter for bots unless you specifically want them to home in on a target with ruthless, mechanical precision using the absolute shortest path possible - good for e.g. precision-guided missiles, bad for animal mobs.

CPU-Blanc:
Phew. That was a bit of a rabbit hole. What started as simple curiosity to see how well the current code would handle diagonals, turned into a complete painful experience due to byond's wonk.

The main issue is windows. If you're moving diagonally North-East, if there's a window on the tile above facing east, there **must** be another window facing south on the same tile, otherwise the mob will get trapped there, even if the tile right is completely free. Issue is highlighted here. Green arrow is a valid path, red arrow would get stuck. Diagonal code has now been adjusted to check that the entering dir to that tile is blocked if it's trying to navigate through it and the exit dir is also blocked.
Screenshot from 2022-03-25 01:25:24

And with the new code:

https://user-images.githubusercontent.com/54823378/160161022-0e6354e5-d98e-42c6-9346-b8718f9fc52b.mp4

I've decided to have the patrol pathfinding stick to cardinals; bots cutting corners on a regular patrol just looks... weird and hasty. However, I also thought it made sense for them to cut corners and take diagonals when tracking a target, as there's more urgency and it _is_ generally faster. Beepsky can now make a beeline for you!

code/modules/mob/living/bot/bot.dm
```diff
@@ -456,17 +485,45 @@
...
-	var/euclid_dist = sqrt(Square(src.x - t.x)+Square(src.y - t.y)+Square(src.z - t.z))
-	euclid_dist *= (pathweight+t.pathweight)/2
-	return euclid_dist
+	var/euclid_dist = sqrt(Square(src.x - t.x) + Square(src.y - t.y) + Square(src.z - t.z))
+	var/currentPathweight = ((locate(/obj/machinery/door) in src) && pathweight == 1) ? 2 : pathweight
+	var/targetPathweight = ((locate(/obj/machinery/door) in t) && t.pathweight == 1) ? 2 : t.pathweight
+	return euclid_dist * ((currentPathweight+targetPathweight)/2)
+
+/turf/proc/Manhatten3dDistance(turf/t)
+	var/manhatten_dist = abs(src.x - t.x) + abs(src.y - t.y) + abs(src.z - t.z)
```
scrdest, line 501: Typo: `manhatten` -> `manhattan`

code/modules/mob/living/bot/bot.dm
```diff
@@ -441,5 +475,84 @@
...
+		if(dir & D.dir)			return 1
 	return 0
+
+
+//Multi-Z AStar Procs
+
+/turf/proc/Euclidean3dDistance(turf/t)
+	var/euclid_dist = sqrt(Square(src.x - t.x) + Square(src.y - t.y) + Square(src.z - t.z))
+	var/currentPathweight = ((locate(/obj/machinery/door) in src) && pathweight == 1) ? 2 : pathweight
+	var/targetPathweight = ((locate(/obj/machinery/door) in t) && t.pathweight == 1) ? 2 : t.pathweight
```
scrdest, line 497: Might be nicer to add a flat 'door traversal cost' if there's a door in the turf rather than using a ternary expression.
It's easier to think about and tune up and it will work with turfs with higher pathweights.

For example, imagine a 'sticky floor' turf that slows down robots massively (pathweight e.g. 5) - our preferences in terms of speed would be: Normal > Normal+Door > Sticky > Sticky+Door. Using current logic, Sticky == Sticky+Door.

code/modules/mob/living/bot/bot.dm
```diff
@@ -441,5 +475,84 @@
...
+	return euclid_dist * ((currentPathweight+targetPathweight)/2)
+
+/turf/proc/Manhatten3dDistance(turf/t)
+	var/manhatten_dist = abs(src.x - t.x) + abs(src.y - t.y) + abs(src.z - t.z)
```
scrdest, line 501: Typo: `manhatten` -> `manhattan`

scrdest (changes requested):

code/modules/mob/living/bot/bot.dm
CPU-Blanc: Whoops. Blame dyslexia. Corrected.

code/modules/mob/living/bot/bot.dm
```diff
@@ -441,5 +475,92 @@
...
+	for(var/obj/A in lower)
+		if(!A.density)		continue
+		if(istype(A,/obj/structure/window) || istype(A,/obj/structure/railing))
+			return (A.dir == dir)
+		else				return 1
+	var/turf/exit = get_step(upper, dir)
+	if(exit.density)		return 1
+	for(var/obj/B in exit)
+		if(B.density)		return 1
+	return 0
```
scrdest, line 566: I think this code is mostly unreachable. Either we return 1 in L561 because it's not a window or railing, or we return a boolean in L560 if it is. 

We'll only ever get to L562 if there are absolutely no objs in lower.

code/modules/mob/living/bot/bot.dm
```diff
@@ -296,13 +296,16 @@
 	return 1
 
 /mob/living/bot/proc/handlePatrol()
-	makeStep(patrol_path)
+	if(makeStep(patrol_path) && frustration)
+		frustration = 0
+	else if(max_frustration)
+		++frustration
```
scrdest, line 302: Shouldn't this make sure we don't go over the max, or are we just delegating that to some other system elsewhere?

scrdest (changes requested):
Aside from the specific notes below - a lot of this would be more readable with some extra whitespace.

code/modules/mob/living/bot/bot.dm
```diff
@@ -405,14 +408,45 @@
...
+			if(DirBlockedWithAccess(pStep,rdir&(EAST|WEST),ID) || DirBlockedWithAccess(pStep,adir&(NORTH|SOUTH),ID) || pStep.density)
+				return 1
+			else
+				for(var/obj/O in pStep)	//Check for dense objects in the other path
+					if(O.density && !istype(O, /obj/machinery/door) && !(O.atom_flags & ATOM_FLAG_CHECKS_BORDER))
+						return 1
+				return 0
+		else if(DirBlockedWithAccess(iStep,adir&(EAST|WEST),ID))	//Adjacent tile exiting dir is blocked but entering isn't. Mobs will get stuck here, even if the other path is free!
+			return 1
+		if(DirBlockedWithAccess(pStep,rdir&(EAST|WEST),ID))	//Same checks, but for the other path
```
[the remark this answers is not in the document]
CPU-Blanc: I couldn't really find a clean way of doing this with a proc define mainly due to the requirements of needing 3 turfs and their directional relation to the start and end turfs. Added some whitespace and trimmed up some unnecessary duplicate checks instead!

code/modules/mob/living/bot/bot.dm
```diff
@@ -405,14 +408,45 @@
...
-		if(!LinkBlockedWithAccess(A,pStep,ID) && !LinkBlockedWithAccess(pStep,B,ID))
-			return 0
-		return 1
+		for(var/obj/O in B)
+			if(O.density && !istype(O, /obj/machinery/door) && !(O.atom_flags & ATOM_FLAG_CHECKS_BORDER))
+				return 1
+
+		//This mess is mainly to combat getting stuck in glass windows due to weird diagonal step_towards behaviour. If you're going northeast for example: if there's a window facing east on the tile above, there MUST be another one facing south or the east facing window will block the move, regardless if the tile to the right of the starting point is completely clear.
+		var/turf/iStep = get_step(A,adir&(NORTH|SOUTH))
+		var/turf/pStep = get_step(A,adir&(EAST|WEST))
```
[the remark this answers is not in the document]
CPU-Blanc: Agreed. Changed to y+xStep with comments

code/modules/mob/living/bot/bot.dm
```diff
@@ -441,5 +475,92 @@
...
+				L.Add(T)
+		else
+			var/obj/structure/stairs/S = locate(/obj/structure/stairs) in src
+			if(S?.dir == d && !LinkBlockedAbove(src,GetAbove(S),d))
+				L.Add(get_step(GetAbove(S), d))
+
+	return L
+
+//Checks for directional blockages at the base of the stairs leading up, open space above, and that nothing is blocking the exit point. Returns 1 on fail
+/proc/LinkBlockedAbove(var/turf/lower, var/turf/simulated/open/upper, var/dir)
```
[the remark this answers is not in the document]
CPU-Blanc: Stairs require open space above them in order to use them; some pesky assistant could come along and place a tile above some stairs for example and block it off.

code/modules/mob/living/bot/bot.dm
CPU-Blanc: Wow. No excuse for this, no idea how that return slipped past me! Corrected and added a reverse direction window check to upper rather than a flat density check.

code/modules/mob/living/bot/bot.dm
CPU-Blanc: This is a mirror of the frustration code used for target pathfinding on L277. I believe the idea is if max_frustration is not set a value (ie you don't want any frustration logic to apply to a bot), then frustration is never counted in the first place and so doesn't trigger frustration checks elsewhere in the code.

max_frustration is handled on L242 and L252 for target pathfinding and patrol pathfinding respectively, with frustration being reset upon a single successful step taken by the bot (L275 & L300)

code/modules/mob/living/bot/bot.dm
```diff
@@ -400,19 +403,54 @@
...
+				return 0
+		else if(DirBlockedWithAccess(xStep,adir&(NORTH|SOUTH),ID))
+			return 1
+
+		for(var/obj/O in yStep)	//Both paths free? Check them for objects. So long as one isn't blocked, we can move
+			if(O.density && !istype(O, /obj/machinery/door) && !(O.atom_flags & ATOM_FLAG_CHECKS_BORDER))
+				for(var/obj/OB in xStep)
+					if(OB.density && !istype(OB, /obj/machinery/door) && !(OB.atom_flags & ATOM_FLAG_CHECKS_BORDER))
+						return 1
+				return 0
```
scrdest, line 452: I can't quite follow why the two for-loops are nested here. If it's a cardinal move, X/Y steps shouldn't influence each other.  Either we move horizontally and don't care about the vertical blockers, or we move vertically and we don't care about the horizontal ones.

If this is a diagonal move, surely we'd only need to check a single turf in a single loop as xStep & yStep refer to the same turf, right? Or is the idea that if we move diagonally, we need to have at least one valid step in both cardinal directions to avoid phasing through corners?

In either case, it doesn't seem like we ever need to compare O and OB here, so it probably should be two separate for-loops over yStep & xStep. This would reduce the complexity of this logic from `O(N^2)` to `O(2N) == O(N)` (or less abstractly - stop this proc from eating the server hardware raw just because someone put a big pile of items on both tiles).

scrdest (changes requested):
Sorry, one more concern I spotted only now - thanks to the var renames, so thank you for applying these!

code/modules/mob/living/bot/bot.dm
CPU-Blanc: This is only triggered in the case of diagonal moves (within the if statement for diagonals on L412)
`xStep` and `yStep` are not the same turf, they're the in-between adjacent turfs on each axis, as byond diagonal movement considers a diagonal move to mean passing through 4 turfs, not 2. This is here to validate that at least one of the steps is clear of objects so the move can complete.

Although they are _technically_ nested, in this case the second `for` is acting more like a `break`. The idea is _either_ `xStep` or `yStep` must be clear of dense objects; if they both have dense objects, the diagonal move will fail.

If there's no dense objects in `yStep`, the second nested `for` loop on L449 is never triggered (`xStep` is never checked). If there is a dense object, the first loop would pause while the second loop checks for blockers in `xStep`. Whether there's blockers or not, that second loop will **always** call return (L451+L452), so the first loop is never ran again, again kinda working like a `break` would. The whole thing works like it's a giant AND check (if loop one is true, check the other loop. If that loop is also true, return 1, else 0)

The idea was to save an additional variable and check, and functions in the same way as below:
```dm
var/blocked = FALSE
for(var/obj/O in yStep)
    if(O.density && !istype(O, /obj/machinery/door) && !(O.atom_flags & ATOM_FLAG_CHECKS_BORDER))
        blocked = TRUE
        break
if(blocked)
    for(var/obj/O in xStep)
        if(O.density && !istype(O, /obj/machinery/door) && !(O.atom_flags & ATOM_FLAG_CHECKS_BORDER))
            return 1
return 0
```

I can understand it doesn't look as obvious as  the above, however, I thought it was a more elegant way of handling it. Happy to switch it over though it readability is a concern!

scrdest: Ah, yes, I see now, the fact that the second for-loop is gated behind an if and there's ALWAYS a return in that if means that the worst-case performance is still O(n+m), so that's fine.

Still, I'd prefer option 2 for clarity; variables are _cheap_

CPU-Blanc: Not a problem! Switched to the above example for readability.

scrdest (approved):
LGTM!

Glloyd:
Looks good to me to! Thanks for doing the heavy lifting on this @scrdest

PR #1066, opened 2022-03-24
The Great Gunfuckening

Razberyl:
Cell swapping has been in for a bit, but even the best device cell RnD could make was worse than the default device cells that the guns started with in the majority of cases, meaning there was pretty much no point in building extra cells to reload with.

That changes now.

RnD's device cells have been boosted to 100Wh and 150Wh, and a new cell was added with 200Wh capacity, making cell swapping actually viable now!

In addition, _cell sizes for energy guns that can be cell swapped have been standardized across the board_, and their charge costs have been adjusted accordingly to keep their max shots the same. I'm sure I've missed some somewhere*, but even the Urist guns have been given a pass, _and fixed_, since they currently can't be fired due to their ludicrous charge costs vs the size of their cells.

(As a side note, there's a bug with how cell-swap code determines max cell size and how the temperature gun changes its charge costs based on what temperature it's set to, but I don't know nearly enough about coding to fix that.)

All of these guns have been tested and confirmed to work with the new values:
dreamseeker_M9jn7fddGO

*Pulse weapons were intentionally left alone, partially due to having incredibly wonky max_shots, partially because we're not going to see them in a normal round anyway, and partially because them needing a car battery to fire is fine.

### DIFF
--- a/code/modules/clothing/chameleon.dm
+++ b/code/modules/clothing/chameleon.dm
@@ -283,7 +283,7 @@
 	fire_sound = 'sound/weapons/gunshot/gunshot_pistol.ogg'
 	projectile_type = /obj/item/projectile/chameleon
 	charge_meter = 0
-	charge_cost = 20 //uses next to no power, since it's just holograms
+	charge_cost = 2 //uses next to no power, since it's just holograms - ACTUALLY uses next to no power now!
 	max_shots = 50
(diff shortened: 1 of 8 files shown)


Glloyd:
That failing check isn't on you, this is good to go.

PR #919, opened 2020-12-11
Bug Fixing - #907, #909, #908

Laniakeaaaa:
### Fixes:

**Fixes disappearing sprite on empty pistol with holdout:**
#909 

**Gives the mimic it's own sprite to avoid it being invisible.**
#907 

**Moves White Dragon vendor into main urist folder to machinery, changes icons to support broken, panel, etc. Adds vend reply and removes delay.**

#908

PR #1067, opened 2022-03-27
Fixes comb & urist pistol icons

Naxodile:
Fixes the urist-specific silenced pistol icon changed in #919 
Fixes the urist comb by having it be a recolor of the normal comb


Glloyd:
Thanks Nax!

PR #1074, opened 2022-04-17
Surgical & medical borg merge + surgery fixes/tweaks

CPU-Blanc:
- Merges the crisis and surgery borg modules together, due to both being crippled without the functionality of the other in terms of giving actual medical aid
- Fixes the organ gripper not being able to be used during surgery steps, due to a check for the organ in their active hand
- Removes duplicate crowbars over all borg modules
- Fixes null entries in wound lists in external organs as well as GC runtimes caused by wounds. Caused by not removing wounds from the wound list before calling `qdel` leaving references
- Fixed/tweaked various verbiage during surgery: Duplicate a's, awkward phrasing etc
- Changed wounds created by scalpels and surgical tools description to be "clean surgical incision". "Large ugly ripped cut" does not make sense for a incision site unless using ghetto tools.
- Added clamped check to wound `heal_damage()` proc. This prevents things like Bicaridine from closing clamped incision sites (surgical wounds already do not auto-heal, so logically this makes sense to me). Admin rejuv will still close the site and heal as intended.
- Removed surgical wound 'damage' from total damage/health tally. Surgical wounds would already not show up as damage to the limb _unless_ another wound is later inflicted on the limb to cause more damage. Happy to reverse this if necessary, but as the damage wasn't tallied on surgery steps, I assumed this was intentional.

### DIFF
--- a/code/_global_vars/lists/flavor.dm
+++ b/code/_global_vars/lists/flavor.dm
@@ -1,6 +1,6 @@
 // Used by robots and robot preferences.
 GLOBAL_LIST_INIT(robot_module_types, list(
-	"Standard", "Engineering", "Surgeon",  "Crisis",
+	"Standard", "Engineering", "Medical",
 	"Miner",    "Janitor",     "Service",  "Clerical", "Security",
 	"Research"
 )) // This shouldn't be a static list. Am I the only one who cares about extendability around here?
--- a/code/global.dm
+++ b/code/global.dm
@@ -147,7 +147,7 @@ var/global/list/alphabet_uppercase = list("A","B","C","D","E","F","G","H","I","J
 
 // Used by robots and robot preferences.
 var/list/robot_module_types = list(
-	"Standard", "Engineering", "Surgeon",  "Crisis",
+	"Standard", "Engineering", "Medical",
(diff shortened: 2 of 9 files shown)


Glloyd:
This mostly looks good to me, although I will defer to @Naxodile for opinions on the bicard/wound changes, because Nax knows medical much better than me.

Naxodile (approved):
Looks good. The wound desc being set should probably be handled in createwound (with that surgical arg it already gets!), but that's only a nitpick.

CPU-Blanc:
The reason it wasn't handled in `createwound` is because although it has a surgical var, it isn't aware of what tool was used. Ghetto surgery means you can cut someone open with a glass shard for example and that would pass the surgical var; however to me it wouldn't make sense to have the glass shard create a clean incision site. That's why I put a type check on L128.

Happy to move it to the `createwound` proc is desired, though I personally am not fond of ghetto surgery giving clean cuts.

If that's really desired, we could perhaps do something like pass surgical = 1 for ghetto surgery, and surgical = 2 for proper and handle it that way?

Glloyd:
The way it's done seems fine to me right now, and it's always something we can change if we want to add those checks down the line.

PR #1075, opened 2022-04-17
adds qm headset to qm locker

ggermanicus:
that's literally it. since it's a glloydstation locker it only had a glloydstation non-head headset.


Glloyd:
Does this locker still have the old headset in it though? If so it might be best to just add a Nerva subtype that has the Nerva headset instead.

PR #1079, opened 2022-05-18
Bugfixing again

### LINKED ISSUE #1068: Selling items via the cargo shuttle is broken
#### Description of issue

Selling items via the cargo shuttle does not work. I'm more adding this as a reminder for myself when I have more free time, although if anyone is currently working on anything related to trading this might be something you want to hit along the way.

#### Difference between expected and actual behavior

We don't get money, I expect to get money.

#### Steps to reproduce 

Put an item in a crate, either from the trade_items list or not, and send it on the cargo shuttle. You only get the crate money.

#### Specific information for locating

It's failing in the supply subsystem find_item_value() proc, it probably needs to be rewritten like @CPU-Blanc did for the NPC trading logic.

#### Length of time in which bug has been known to occur

Unknown, but seemingly for a while. It worked when I first added it, but at some point it broke.

#### Issue bingo

Please check whatever applies. More checkboxes checked increase your chances of the issue being looked at sooner.

- [X] [Couldn't find an existing issue about this](https://github.com/Baystation12/Baystation12/issues)

CPU-Blanc:
Several bundled bugfixes:

- Fixed a edge case where someone double clicks on a holopad and starts a call, sending 2 requests at the same time and confusing the system. Call cooldown check has been moved further up
- Fixed bot pathfinding to avoid closed blastdoors. Bots try to open all doors with bump, but blastdoors cannot be opened by normal means. They will now check for and avoid closed ones
- Fixed selling items on the NanoTrasen Station shuttle (Closes #1068). Compiles a list and bundles all common items together, then calculates the value of multiple sales taking into account a drop of value caused by selling an individual item. Currently it's set to drop the price by 4% on each sale, this compounds! Can be changed by setting `price_modifier`. Materials currently aren't effected by price drops because they're basically worthless as-is and if they weren't, it'd simply encourage mining to actually become a thing /shrug
- Fixed an issue where readying as AI at roundstart always sent you back to the lobby unless you got selected for antag. This was a really strange one, and a bit of a chicken vs egg situation. This was caused by AI slots being limited by the amount of inactive AI cores present. AI cores would spawn on the roundstart hook depending if the gamemode allowed AIs or not. The issue is, this hook is too late as it is called _after_ jobs have been given out by the jobs subcontroller. However, the previous hook is too early as the gamemode hasn't been set in `SSticker` so it cannot check if AI's are allowed. The solution was to add 2 new hooks: `hook/gamemode_selected` and `hook/gamemode_start_failed`, both of which are passed a `/datum/game_mode` parameter. The `gamemode_selected` hook is called in `SSticker` once a gamemode has been selected via votes, but _before_ any jobs/roles have been assigned. This allows a core to spawn if the mode allows it, and for the job limit for AIs to be properly taken into account. `gamemode_start_failed` is called if the gamemode then fails to start for whatever reason (not enough players, no antags etc). This is currently used to despawn the AI cores to allow for the next gamemode presented to handle spawning again.
- Fixed roundstart AI spawn logic. On fixing the roundstart AI problem, another bug presented itself; roundstart AIs were spawning ontop of inactive AI cores. This was caused by the way `new_player` handled `AIize()`. It used the old method of spawning a new AI mob at the AI landmark, rather than using an inactive core. New logic has been added to use the inactive core first, and if none are available, use the old method as a fallback. As this is called after roles have been handed out, the only situation this will happen is on gamemodes where the AI core wasn't allowed to spawn, but there is an antag role to override the available core check (such as Malf AI)
- Fixed money cannon 0Th infinite round exploit. Previously, you could set the cannon to fire rounds of 0Th and just... spam the  map. Now it requires a minimum of 1
- Fixed combat channel. Somehow during the staging of my last PR to add the combat channel, I missed these changes. Whoops. Actually adds the combat channel to the overmap ship and corrects a missing call to it. Also (a bit jankily) adds the combat frequency to the AI integrated headset so that they are able to access it. It is jank, but this was the best way I could think of without having any hard references for paths that are Nerva only subtypes and thus not guaranteed to be included.
- Added T-scanner to drones. Because it is a pain pulling up every single tile to try and find a broken cable.
- Fixed the wiki and discord buttons from giving scuffed URLs
- Added a _bunch_ of missing config options + corrected default logic. While doing this PR, I noticed several config options were completely missing from `config.txt` and `game_options.txt`. These have been added in and commented/uncommented to match our current settings. Several options were also unchangeable via `config.txt`. This was caused by the variable being initialised as 1, with the config option setting the var to 1 if it was uncommented-- no way to set it to 0. These options have been set to init to 0, and have been uncommented to set to 1 where they were previously by happenstance. The only option that will change from all of this is `ORGANS_CAN_DECAY`. This was set to true, however was placed in `config.txt`, but this option is only read in `game_options.txt` where it has now been moved.
- Config `OBJECTIVES_DISABLED` option & variable renamed to `OBJECTIVES_ENABLED`. Because.. that's actually what it does... Determines which objectives types are enabled, not disabled

As usual, give me a good yelling if I've done anything stupid or anything needs changing (namely that AI radio). Always happy to switch things over!

### DIFF
--- a/code/controllers/subsystems/supply.dm
+++ b/code/controllers/subsystems/supply.dm
@@ -186,11 +186,12 @@ SUBSYSTEM_DEF(supply)
 						continue
 
 					//Sell anything else that isn't unique or needs special handling
-					if(A.type in to_sell)
-						to_sell[A.type]["count"]++
-						qdel(A)	//We already have an object reference to use for values, duplicate items are safe to qdel
-					else
-						to_sell[A.type] = list("obj" = A, "count" = 1)
+					if(GLOB.using_map.using_new_cargo)
+						if(A.type in to_sell)
+							to_sell[A.type]["count"]++
+							qdel(A)	//We already have an object reference to use for values, duplicate items are safe to qdel
+						else
+							to_sell[A.type] = list("obj" = A, "count" = 1)
 
 			else
 				qdel(AM)
@@ -297,24 +298,15 @@ SUBSYSTEM_DEF(supply)
 
 	var/sell_modifier = 1
 
-	if(GLOB.using_map.using_new_cargo)
-		if(istype(object, /obj/item/stack/material))	//Materials aren't effected by reduced value/compounded prices. Apply any needed modifiers and sell as-is
-			var/obj/item/stack/material/M = object
-			sell_modifier = GLOB.using_map.new_cargo_inflation * 0.25
-			if(use_reinf_material)
-				return round(round(count * 0.5) * M.reinf_material.sale_price * sell_modifier)
-			return round(count * M.material.sale_price * sell_modifier)
-		
-		else if(GLOB.using_map.trading_faction?.reputation > 50)
-			sell_modifier = (((GLOB.using_map.trading_faction.reputation - 50) / 100) / 2) + 1	//0.5% ~ 25% bonus
+	if(istype(object, /obj/item/stack/material))	//Materials aren't effected by reduced value/compounded prices. Apply any needed modifiers and sell as-is
+		var/obj/item/stack/material/M = object
+		sell_modifier = GLOB.using_map.using_new_cargo ? (GLOB.using_map.new_cargo_inflation * 0.25) : M.matter_multiplier
+		if(use_reinf_material)
+			return round(round(count * 0.5) * M.reinf_material.sale_price * sell_modifier)
+		return round(count * M.material.sale_price * sell_modifier)
 
-	else
-		if(istype(object, /obj/item/stack/material))
-			var/obj/item/stack/material/M = object
-			sell_modifier = M.matter_multiplier
-			if(use_reinf_material)
-				return round(round(count * 0.5) * M.reinf_material.sale_price * sell_modifier)
-			return round(count * M.material.sale_price * sell_modifier)
+	if(GLOB.using_map.trading_faction?.reputation > 50)
+		sell_modifier = (((GLOB.using_map.trading_faction.reputation - 50) / 100) / 2) + 1	//0.5% ~ 25% bonus
 
 	//try and find it via the global controller
 	var/datum/trade_item/T = SStrade_controller.trade_items_by_type[object.type]
@@ -339,7 +331,7 @@ SUBSYSTEM_DEF(supply)
 	var/total_value = newPrice	//Price changes AFTER an obj is sold. Let's skip the first trade then.
 	count--
 	while(count)
-		newPrice = (newPrice * (1-src.price_modifier))	//Calculate what the new price would be with the devalue of selling individually
+		newPrice = newPrice * (1-src.price_modifier)	//Calculate what the new price would be with the devalue of selling individually
 		total_value += newPrice
 		count--
 	return round(total_value)
--- a/code/controllers/subsystems/supply.dm
+++ b/code/controllers/subsystems/supply.dm
@@ -33,6 +33,10 @@ SUBSYSTEM_DEF(supply)
 	//virus dishes uniqueness
 	var/list/sold_virus_strains = list()
 
+	//Price changes
+	var/price_modifier = 0.04	//% price goes down from each sale of a material to the supply station: 4% decrease in value for each item sold
+	var/sold_items = list()	//Items sold to the supply station	//TODO: Generate an inventory of trade_item datums for the station and simulate supply/demand proper
+
 /datum/controller/subsystem/supply/Initialize()
 	. = ..()
 	ordernum = rand(1,9000)
@@ -110,6 +114,8 @@ SUBSYSTEM_DEF(supply)
 
 /datum/controller/subsystem/supply/proc/sell()
 	var/list/material_count = list()
+	var/list/to_sell = list()
+	var/list/crates = list()
 
 	for(var/area/subarea in shuttle.shuttle_area)
 		for(var/atom/movable/AM in subarea)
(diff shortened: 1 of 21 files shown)


CPU-Blanc:
Hey Glloyd. As far as I can tell, everything is behaving as usual. Using the debug verbs to open the config datum shows that all values I changed to init to 0 are set to 1 as they were previously. For example:
Screenshot from 2022-05-19 00:59:44

The only one that's changed is organ decay as previously mentioned

If there's any specific instances you can remember being broken, I'd be happy to test and verify them specifically!

Glloyd:
Sorry, went away for the long weekend and I've been away all week. I don't remember any specific vars that weren't working, this was like 7 years ago maybe, it's been a while, haha.

code/controllers/subsystems/supply.dm
```diff
@@ -191,12 +182,42 @@ SUBSYSTEM_DEF(supply)
 							if(!(dish.virus2.uniqueID in sold_virus_strains))
 								add_points_from_source(5, "virology_dishes")
 								sold_virus_strains += dish.virus2.uniqueID
+						qdel(dish)
+						continue
+
+					//Sell anything else that isn't unique or needs special handling
+					if(A.type in to_sell)
```
Glloyd, line 189: this should probably be under using_map.using_new_cargo

code/controllers/subsystems/supply.dm
```diff
@@ -191,12 +182,42 @@ SUBSYSTEM_DEF(supply)
...
+						to_sell[A.type] = list("obj" = A, "count" = 1)
+
+			else
+				qdel(AM)
+
+	var/payout = 0
 
-			qdel(AM)
+	//Sell all items in bulk and qdel them
+	for(var/atom_type in to_sell)
```
Glloyd, line 201: along with this

code/controllers/subsystems/supply.dm
```diff
@@ -264,16 +285,61 @@ SUBSYSTEM_DEF(supply)
...
+			sell_modifier = GLOB.using_map.new_cargo_inflation * 0.25
+			if(use_reinf_material)
+				return round(round(count * 0.5) * M.reinf_material.sale_price * sell_modifier)
+			return round(count * M.material.sale_price * sell_modifier)
+		
+		else if(GLOB.using_map.trading_faction?.reputation > 50)
+			sell_modifier = (((GLOB.using_map.trading_faction.reputation - 50) / 100) / 2) + 1	//0.5% ~ 25% bonus
+
+	else
+		if(istype(object, /obj/item/stack/material))
```
Glloyd, line 312: this is already handled on line 152, is it not?

code/controllers/subsystems/supply.dm
```diff
@@ -264,16 +285,61 @@ SUBSYSTEM_DEF(supply)
...
+/datum/controller/subsystem/supply/proc/find_item_value(var/obj/object, var/count = 1, var/use_reinf_material = FALSE) //here we get the value of the items being traded
 	if(!object)
 		return 0
 
-	//this uses the default SS13 item_worth procs so its a good fallback
-	. = get_value(object)
+	var/sell_modifier = 1
 
-	var/datum/trade_item/T
+	if(GLOB.using_map.using_new_cargo)
```
Glloyd, line 300: If you make the above changes, payout() will only be called if the map is using_new_cargo, which is how it should be, so you could remove the check here, or put it at the top of the proc instead

Glloyd (changes requested):
Just a couple little things related to cargo. I'm trying to keep compatibility with glloydstation still, so I've marked out a few changes that'll need to be made to do so. Looks good besides that though!

CPU-Blanc:
Hey Glloyd! I must admit, my knowledge of the old trading system is a little lacking. Is it just items I've put under the "others" sort of bit that the old system shouldn't process? ie, stacks, manifests, surveys & virus dishes for the old system and nothing else?

code/controllers/subsystems/supply.dm
CPU-Blanc: L152 calls `find_item_value()` for each material present in a stack and adds the resulting value to the list, it doesn't actually handle how much each stack/material is worth. The idea was that if we wanted to change how materials are sold/calculated in the future, all of that is under the `find_item_value` proc and L152 just handles the splitting (like if we wanted to have compounded sales effect materials too later down the road)

Glloyd:
Yup, just those and crates. The whole additional trading system falls under the "new cargo" umbrella.

code/controllers/subsystems/supply.dm
CPU-Blanc: Resolved in 28d597c3e46e3f3174ea47be272505b37db6c774
Check isn't needed, as the list won't be populated so nothing to iterate through

code/controllers/subsystems/supply.dm
CPU-Blanc: Re-wrote this part to clean it up a little. Check has been moved to within the type check for stacks. It's still required there as stacks are sold in both the old and new system but at least this way it's not constantly being checked when it's unnecessary.

Glloyd: To clarify, material stacks are sold slightly differently under the old system, and the actual "points" are added separately from trading points on line 210. I haven't looked through your changes fully yet, but doing it this way might end up with materials being sold twice on Glloydstation.

CPU-Blanc: I'm rather confused. Comparing the old file to the new changes, nothing should be any different. In the old file material stacks were tallied https://github.com/UristMcStation/UristMcStation/blob/a8182098b0e1dad3c201e47b16b996e1c5a67689/code/controllers/subsystems/supply.dm#L142 & https://github.com/UristMcStation/UristMcStation/blob/a8182098b0e1dad3c201e47b16b996e1c5a67689/code/controllers/subsystems/supply.dm#L173

In fact, looking at this code, materials _were_ actually being sold twice if you were using the new cargo system due to a lack of a `continue` or `else` on L150 or L164 respectively.

Both of these lines tallied the value using slightly different methods --one using cargo inflation and the other the material multiplier, which has been carried over in the new code on L303-- and added to the same list, `material_count` which was then processed at the end (L197 on the old, L210 new). The method of calculating the value has simply been moved outside of the sell loop and into `find_item_value()` for ease of future changes. Again, on the old file both new cargo and old added the materials to the same list which was then processed in the same way without a further new cargo check. I'm really struggling to see how this new code functions any differently unless I'm being incredibly dumb?

scrdest (commented):
Nothing urgent; worth noting though, the renames might cause some maintenance pain down the line.

code/controllers/subsystems/supply.dm
```diff
@@ -264,16 +286,52 @@ SUBSYSTEM_DEF(supply)
...
-	//this uses the default SS13 item_worth procs so its a good fallback
-	. = get_value(object)
+	var/sell_modifier = 1
 
-	var/datum/trade_item/T
+	if(istype(object, /obj/item/stack/material))	//Materials aren't effected by reduced value/compounded prices. Apply any needed modifiers and sell as-is
+		var/obj/item/stack/material/M = object
+		sell_modifier = GLOB.using_map.using_new_cargo ? (GLOB.using_map.new_cargo_inflation * 0.25) : M.matter_multiplier
+		if(use_reinf_material)
+			return round(round(count * 0.5) * M.reinf_material.sale_price * sell_modifier)
```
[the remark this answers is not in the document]
CPU-Blanc: This is a carry over from the old code https://github.com/UristMcStation/UristMcStation/blob/a8182098b0e1dad3c201e47b16b996e1c5a67689/code/controllers/subsystems/supply.dm#L178

I _believe_ if I remember correctly that reinforcing a material is done on a 2:1 ratio so this is here so value isn't just fabricated from nothing?

code/controllers/subsystems/supply.dm
```diff
@@ -264,16 +286,52 @@ SUBSYSTEM_DEF(supply)
...
+		return round(count * M.material.sale_price * sell_modifier)
+
+	if(GLOB.using_map.trading_faction?.reputation > 50)
+		sell_modifier = (((GLOB.using_map.trading_faction.reputation - 50) / 100) / 2) + 1	//0.5% ~ 25% bonus
```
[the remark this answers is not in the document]
CPU-Blanc: Again as above, this was simply carried over from the old code but corrected as it only had a base of +0.75. Meaning the modifier would actually be less than 1 if you were over 50 rep until like 75 rep. So you'd be selling for a loss compared to if you sat at 49 rep
https://github.com/UristMcStation/UristMcStation/blob/a8182098b0e1dad3c201e47b16b996e1c5a67689/code/controllers/subsystems/supply.dm#L158

I personally think the `if` check would still be required as I _believe_ faction standing with NT starts at 30 with others hovering around the 10-25 range with no great way to increase that currently. With the diminishing sale value added in this PR on top, that might be a bit too much of a nerf to cargo in its current state

code/controllers/subsystems/supply.dm
```diff
@@ -121,67 +127,52 @@ SUBSYSTEM_DEF(supply)
...
-
-							add_points_from_source(addvalue, "trade")
+					//Calculate sales for each material
+					if(istype(A, /obj/item/stack/material))
+						var/obj/item/stack/material/S = A
+						var/amount = S.get_amount()
+						if(S.material.sale_price > 0)
+							material_count[S.material.display_name] += find_item_value(S, amount)
+						if(S.reinf_material?.sale_price > 0)
+							material_count[S.reinf_material.display_name] += find_item_value(S, amount, TRUE)
```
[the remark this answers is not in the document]
CPU-Blanc: Good catch. I'll get on that

CPU-Blanc:
Happy to revert the config + variable rename; however I stand by that in its current state it's stupid and only serves to confuse given it does exactly the opposite of its name, with little to no documentation on its usage without seeing where the variable is utilised in the code and working it out from there.

code/controllers/subsystems/supply.dm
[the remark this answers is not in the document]
CPU-Blanc: Actually, after having a quick look, there might be an issue with using `name` as opposed to `display_name`. That is that currently the keys of `material_count[]` is passed to `add_points_from_source()` as the source. This could cause weird and duplicate entries on the cargo statistics screen and could even be completely missed in the supply nano module `print_summary()` proc which iterates through `point_source_descriptions[]` which in turn uses global material display names (not names).

I could of course work around this by indexing by the material name and using a nested list to contain the display_name to pass back to the `add_points_from_source()` proc but that seems rather excessive? How likely would using `display_name` lead to causing an issue? It's also the variable the current system is using

Glloyd:
After some testing, I think this is okay as-is. I'd like to get it in for today regardless, as I'll need a bit of time to mess with the serverside config to get the new changes implemented. Thanks again for this Blanc! You're a legend.

PR #1081, opened 2022-06-05
Noctis mapping fixes

Naxodile:
Railing color fixed, medicine cabinet moved, fluff addition/tweaks. Just some minor things I noticed when viewing it.


Glloyd:
Looks good Nax! Vak also suggested a gas sensor for the fuel tank if you want to add that here or in a future PR.

Glloyd:
I'll merge this for the sake of getting it in, and we can add the gas sensor in the future.

PR #1084, opened 2022-06-09
New Whip

CaraS212:
Adds a new whip for the captain with nautical flavor.

also makes goat guards no longer spriteless


Glloyd (approved):
we did it!

PR #1089, opened 2022-11-06
Bugfixing again

CPU-Blanc:
Fixes a couple of bugs raised by AseaHeru in Discord

- New lighttube boxes made from cardboard would not take any items. This was caused by the `make_exact_fit()` proc that'd assume the storage item would already have contents and set `max_storage_space` to match the count of contents. As these new boxes had no contents, this was being set to 0. Added a check for contents under the `make_exact_fit()` proc and set default values for the boxes manually to match what pre-spawned boxes have.
- Fistful of shotgun shells were remaining with 0 shells once inserted into an ammo holder. This makes sense for magazines but not a fistful of shells. This was caused by a missing `qdel()` call/check in the `ammo_magazine/attackby()` proc. To facilitate future possible ammo expansion and to avoid strict typecasting, I've refactored the fistful of shotgun shells to be under a new subtype of ammo_magazine, `ammo_magazine/bundle`. Any future fistful of x ammo types can go here and the correct qdel checks and icon updates will apply

Gonna tag @Cheedows here as he worked on the ammo PR in case he had any input or concerns.

As always, let me know if there's any issues, it's been a hot while since I last touched and DM code.

### DIFF
--- a/code/game/objects/items/weapons/storage/boxes.dm
+++ b/code/game/objects/items/weapons/storage/boxes.dm
@@ -411,9 +411,8 @@
 
 /obj/item/weapon/storage/box/lights/bulbs/empty
 	startswith = null
-
-/obj/item/weapon/storage/box/lights/bulbs/empty
-	startswith = null
+	max_w_class = ITEM_SIZE_SMALL
+	max_storage_space = 25
 
 /obj/item/weapon/storage/box/lights/tubes
 	name = "box of replacement tubes"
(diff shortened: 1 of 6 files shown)


Glloyd:
> it's been a hot while since I last touched and DM code.

Me too, but I actually started messing with it again this past week. Now that I'm around more often I want to get back to working on Urist.

Glloyd (approved):
like I said, it's been a while, but this all looks good to me!

PR #1091, opened 2023-01-09
[WIP] GOAI initial obstacle handlers pass

CPU-Blanc:
The first initial pass at adding obstacle handlers to the new GOAI.
Still needs a lot of work and cleaning up, but draft PR for better co-ordination.

Todo:

- [ ] Code cleanup
- [ ] Finish airlock hack handler (fuck wire datums)
- [ ] Add additional completed/navigation checks for directional windoors
- [ ] Add console obstacle handling (climb on)
- [ ] Add table obstacle handling
- [ ] Finish generic screw handler
- [ ] Finish generic pry handler
- [ ] Debug and fix the AI having a bit of a door fetish
- [ ] Verify handlers all work as intended
~~Split/move handlers into a new `mob_commander` subtype for `/mob/living/carbon/human`???~~
- [ ] Implement an interaction datum for atoms to return available actions for the calling pawn type 

As per usual, if I've missed stuff or done something dumb, just yell. I'm sure I have somewhere.


code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/obstacle_handlers/airlocks.dm
```diff
@@ -0,0 +1,287 @@
...
+			src.brain.DropMemory(MEM_OBSTRUCTION)
+		tracker.SetFailed()
+		return
+
+	//If the mob can see the airlock, check if its bolt lights are on
+	if(obstruction in src.brain.perceptions[SENSE_SIGHT_CURR])
+		if(obstruction.locked && obstruction.lights)
+			tracker.SetFailed()
+			src.brain.SetMemory(MEM_OBJ_LOCKED(obstruction), TRUE, 5 MINUTES)
+			if(src.brain.GetMemoryValue(MEM_OBSTRUCTION, null) == obstruction)
```
scrdest, line 170: Not specifically directed at you, but this made me realize - we'll probably need to refactor this to store a weakref and resolve it, or the GC will get mad at us.

code/modules/urist/modules/GOAI/integrations/blocker_definitions/machinery.dm
```diff
@@ -2,6 +2,12 @@
 
 // MACHINERY
 
+/obj/machinery
+	blocker_gen_enabled = TRUE
+
```
scrdest, line 7: I've gone through these, I don't think we can afford to default to generating here. A whole lot of machinery should be non-blocking - wires, buttons and such.

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/obstacle_handlers/generic.dm
```diff
@@ -0,0 +1,60 @@
...
+
+/datum/goai/mob_commander/proc/HandleScrew(var/datum/ActionTracker/tracker, var/obj/target)
+	log_debug("Screw task for [target]")
+	tracker.SetDone()
+	return
+	//TODO
+
+/datum/goai/mob_commander/proc/HandleGenericBreak(var/datum/ActionTracker/tracker, var/obj/target, var/method)
+	log_debug("Break task for [target]")
+	var/mob/pawn = src.GetPawn()
```
scrdest, line 17: Can we rename this to mob_pawn? I'd like to keep plain 'pawn' to be an atom by convention to make it explicit what logic is supported for all objects vs just mobs vs just humanoids.

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/obstacle_handlers/airlocks.dm
```diff
@@ -0,0 +1,287 @@
...
+			tracker.SetFailed()
+			src.brain.SetMemory(obstruction.locked ? MEM_OBJ_LOCKED(obstruction) : MEM_OBJ_NOPOWER(obstruction), TRUE, 5 MINUTES)
+			if(src.brain.GetMemoryValue(MEM_OBSTRUCTION, null) == obstruction)
+				src.brain.DropMemory(MEM_OBSTRUCTION)
+			return
+		if(!obstruction.allowed(pawn))
+			//No access to the airlock. Add it memory and fail the tracker. Still poke it though for that sweet sweet denied buzz
+			obstruction.attack_hand(pawn)
+			tracker.SetFailed()
+			src.brain.SetMemory(MEM_OBJ_NOACCESS(obstruction), TRUE) //Come back to this and change the ttl if changing IDs/access becomes a thing
```
scrdest, line 197: Should probably still use MEM_TIME_LONGTERM rather than the default infinity

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/obstacle_handlers/generic.dm
```diff
@@ -0,0 +1,60 @@
...
+		return
+
+	//Cooldown check
+	var/last_action = tracker.BBGet("LastAttack", null)
+	if(last_action && (world.time < last_action + DEFAULT_ATTACK_COOLDOWN))
+		return
+
+	//See what's in the mob's hands
+	var/obj/item/AH = pawn.get_active_hand()
+	var/obj/item/IH = pawn.get_inactive_hand()
```
scrdest, line 46: Just curious, does that work for the generic /mob type? I would have thought only humanoids would support this!

code/modules/urist/modules/GOAI/integrations/blocker_definitions/machinery.dm
CPU-Blanc: The default `GenerateGenericFullBlocker()` will only do it's thing (by default) if the `/obj/machinery` is dense. As it only gets called when Astar checks it and it's then cached to `blocker_gen_enabled = FALSE`, would this really be significant overhead in return for some definition shortcuts, as it'll only run once _if_ Astar pokes it? Idea was for things that shouldn't generate, we could just override it by default?
Happy to reverse this though if you think it will be!

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/obstacle_handlers/generic.dm
CPU-Blanc: Absolutely!

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/obstacle_handlers/airlocks.dm
CPU-Blanc: At the moment we don't have plans to handle the pawn's access changing, so I don't see why we shouldn't store it forever? The access won't change, so it'll always fail here.

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/obstacle_handlers/generic.dm
CPU-Blanc: It actually does!
Definition is from `/mob/proc/get_inactive_hand()`

There are some cases I could see being an issue in the future though, so might it be an idea at some point to make a new `mob_commander` subtype designed to be used with humans/humanoids only, and have them run different handlers to the parent type? Either that or the handlers get expanded to typecast the pawn itself?

code/modules/urist/modules/GOAI/integrations/blocker_definitions/machinery.dm
scrdest: Ah, you're right. I think I need to work on some more docs for this, the complexity of this project is getting overwhelming.

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/obstacle_handlers/generic.dm
scrdest: I want to keep AI polymorphous w.r.t. the pawns, the pawn type shouldn't leak through to the AI. Type-casting is good enough for now.

Although now that you mention it - we could make a submodule class for the AI that gets instantiated based on the pawn class and handles the type-castey bits.

So, for example, instead of directly adding an action, we'd go through the associate that would return e.g. mob-flavored/object-flavored/humanoid-flavored, etc. variants of each Action and/or provide a unified API to do things for all different types where possible. 

Option 2 is similar but inverted - rather than creating an ACL class based on the pawn type, we inject a set of Interface objects into atoms one way or another and just use these in handlers. This is effectively the same pattern as the Cover/Blocker stuff.

scrdest:
Also, w.r.t. door fetishism - the old dev-doors had their costs changed, you might want to look at that. They seem to work much better (in the GOAI repo's test maps) with the new costs.

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/obstacle_handlers/airlocks.dm
scrdest: I guess it would be nicer to leave the door (pun not intended) open for a periodic recheck. Not a big deal though.

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/actions/waypoint.dm
```diff
@@ -91,13 +91,13 @@
...
 			if(last_link_blocked)
-				OBSTACLEHUNT_DEBUG_LOG("[owner]: LINK BETWEEN [previous] & [pathitem] OBSTRUCTED")
+				OBSTACLEHUNT_DEBUG_LOG("[owner]: LINK BETWEEN [previous] ([previous.x],[previous.y]) & [pathitem] ([pathitem.x],[pathitem.y]) OBSTRUCTED")
 				// find the obstacle
 				var/atom/obstruction = null
 
 				if(!obstruction)
 					for(var/atom/potential_obstruction_curr in pathitem.contents)
-						var/datum/directional_blocker/blocker = potential_obstruction_curr?.directional_blocker
+						var/datum/directional_blocker/blocker = potential_obstruction_curr.GetBlockerData(TRUE, TRUE)
```
scrdest, line 100: As a rule, I'd recommend the null-conditional `?.` here to deal with nulls.

code/modules/urist/modules/GOAI/integrations/blocker_definitions/_shared.dm
```diff
@@ -4,12 +4,79 @@
...
+
+/proc/GenerateDynamicDirBlocker(var/owner, var/invert_dir = FALSE)
+	var/atom/atom_owner = owner
+	var/datum/directional_blocker/blocker = null
 
-		else
-			// Caching to speed up queries
-			atom_owner.cover_gen_enabled = FALSE
+	if(atom_owner)
+		var/dir_to_block = invert_dir ? dir2opposite(atom_owner.dir) : atom_owner
```
scrdest, line 43: I think this is an oversight. This currently causes runtimes.
```suggestion
		var/dir_to_block = invert_dir ? dir2opposite(atom_owner.dir) : atom_owner.dir
```

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/actions/waypoint.dm
CPU-Blanc: Is`?.` necessary here? Doesn't the way DM iterate through lists like this ignore nulls because we're iterating by type `/atom`? If it's not of type `/atom`, it's not iterated over surely?

code/modules/urist/modules/GOAI/integrations/blocker_definitions/_shared.dm
CPU-Blanc: WHOOPS. Yep, that's totally a mistake on my side. I quickly changed to code to allow for inverting directions and obviously didn't read over it through enough. My bad.

code/modules/urist/modules/GOAI/integrations/base_commander/mob_commander/actions/waypoint.dm
scrdest: Hmm... that's an interesting question and I don't know the answer to it. As a rule of thumb, I don't trust anything DM gives me as non-null until explicitly checked, but it might be unnecessary.

CPU-Blanc: From DM reference:
```If the loop var has a type, a hidden istype() call is included in the code. Only items of that type, or its descendants, are handled within the loop. That means null values and objects of unrelated types will be skipped.```

scrdest: Of-bloody-course it's an _implicit typecheck_, can't have any too-clever optimizations in this godforsaken language...

image

CPU-Blanc:
Closing to re-target scr's branch for easier development

PR #1093, opened 2023-01-24
Bugfixing again

CPU-Blanc:
Just a few small bugs I've noticed/been told about.

Fixes the following:

- `simple_animal`'s being able to attack while netted: Tweaked the code so that netted `simple_animal`'s count as restrained.
- Soulstone not recapturing shards: This relies on 'attacking' the shard with the soulstone. Attack calls are ignored if the item has no `force`.. Derp.
- Typo in the maintenance drone event
- Silicons receiving alarms for aways: Added a connected Z-level check before alerting the listener

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -491,3 +491,7 @@
 /mob/living/simple_animal/Destroy()
 	GLOB.simple_mob_list.Remove(src)
 	..()
+
+/mob/living/simple_animal/restrained()
+	if(buckled && istype(buckled, /obj/effect/energy_net/safari))
+		return 1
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -490,4 +490,8 @@
 
 /mob/living/simple_animal/Destroy()
 	GLOB.simple_mob_list.Remove(src)
-	..()
\ No newline at end of file
+	..()
+
+/mob/living/simple_animal/restrained()
+	if(buckled && istype(buckled, /obj/effect/energy_net/safari))
```
scrdest, line 496: Does this really need to be constrained to a subtype?

scrdest (approved):
One minor quibble only.

code/modules/mob/living/simple_animal/simple_animal.dm
CPU-Blanc: I wasn't sure if anything else buckles `simple_animal`s that warranted removing their ability to attack. ie, buckling a panther to a chair imo should only count as `PARTIALLY_BUCKLED`, rather than `FULLY_BUCKLED`. I believe it's the same for humans

PR #1096, opened 2023-01-28
Terran marine_space faction fix

Meunahlux:
fixes the marine_space simplemob having the Terran faction and therefore always being very intolerant and angry about the fact that people aren't wearing their confederation pins at any given moment. I did this by removing the faction= "Terran" being set at the root mob and therefore might not actually have done it right. But it should be good to go.

### DIFF
--- a/code/modules/urist/mob/humanhostiles.dm
+++ b/code/modules/urist/mob/humanhostiles.dm
@@ -254,7 +254,6 @@
 	faction = "terran"
 
 /mob/living/simple_animal/hostile/urist/terran/marine_space
-	faction = "terran"
 	name = "\improper Terran Confederacy Marine"
 	desc = "A Terran Confederacy Marine. This one is wearing a voidsuit."
 	ranged = 1


code/modules/urist/mob/humanhostiles.dm
```diff
@@ -254,7 +254,6 @@
 	faction = "terran"
 
 /mob/living/simple_animal/hostile/urist/terran/marine_space
-	faction = "terran"
```
scrdest, line 257: The removed line is an override, think of it as an extra layer on top of a pile that takes precedence above anything below it.

With this removed, we fall back to the parent class (`/mob/living/simple_animal/hostile/urist/terran`), which doesn't set it at all, so we keep going to the grandparent, `/mob/living/simple_animal/hostile/urist`, which sets it to "neutral".

"neutral" makes mobs, well, neutral to typical spessmen, so this should be good. There's some extra special sauce in the Factions subsystem, but by the looks of it it should work as intended too.

scrdest (approved):

PR #1097, opened 2023-01-31
feature: New Spoops

scrdest:
A bunch of spooky effects in preparation of bigger things to come (a Certain Gamemode getting reworked).

* Added a new Hallucination effect. With severe enough hallucinations, you will start seeing monsters. Featuring all new, animated sprites, ranging from a slightly goofy imp, through shadow people, to Discount Amigara Horrors.
* Refactored Haunters, the /obj/effect/shadow_wight clone. They now have their own file, a more maintainable/extendable framework, and a new variant that chases after the last mob to cross their path. These further come in classic Narnar ghost flavor and in randomized Hallucination monster variant from the earlier point.

Preview: 
HorrGreenImp
HorrSpriteVeiny1Thick


Glloyd (approved):
👍

PR #1102, opened 2023-02-14
ship combat rebalancing

Glloyd:
I am in a balancing hell of my own creation:

buffs ai ship hulls (might need to further buff them depending on how things go). buffs lactera shields very slightly. massively buffs AP weaponry, and autocannon ammo capacity. Differentiates ammo types better: HE's role is busting components, AP's is doing huge hull damage. Slightly nerfs the damage of bluespace torps but makes them accurate component busters. Might need to nerf them further idk. Also adds an HE torp as a middle ground between the bs torp and the AP torp.


Glloyd:
will probably do more tweaks to this in the future but i'll merge this for now and see how it goes

PR #1104, opened 2023-02-17
Rebalances traitor costs+ammo+adding antag items

ggermanicus:
Removes some ammo noob traps (any gun you can buy should let you buy ammo too, except railguns)
Raises the cost of some of the strongest firearms
Unifies cost of 357 and 44 and lowers them in total. They're much weaker than they were years ago and functionally identical. Revolvers themselves have also been fixed so the 44 no longer takes 357.
Syndicate toolbox now comes with a toolbelt
You can now buy cuffs/duct tape and metal foam grenades
Explosive implant cost largely reduced.
Decreases odds of ballistic pod buyers thinking they are getting guns

### DIFF
--- a/code/datums/uplink/ammunition.dm
+++ b/code/datums/uplink/ammunition.dm
@@ -6,11 +6,11 @@
 	category = /datum/uplink_category/ammunition
 
 /datum/uplink_item/item/ammo/c45m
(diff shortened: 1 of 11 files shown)


Glloyd (approved):
looks good

PR #1107, opened 2023-02-20
Bluespace Angel

CaraS212:
Adds Bluespace Anglels for all your badmin needs, like smiting prayer abuse.
-Adds 3 angels sprites/kinds.

current issue: they leave a bullet case behind after shooting despite using an energy projectile


code/modules/urist/mob/spoopymobs.dm
```diff
@@ -702,4 +702,65 @@
...
+	visible_message("<span class='danger'>The [src.name] wails and disappears!</span>")
+	playsound(src.loc, 'sound/effects/angel-reveal.ogg', 50, 1)
+	flick("forgotten_die", src)
+	sleep(4)
+	qdel(src)
+	return
+
+/obj/item/projectile/energy/holy
+	name = "holy fire"
+//	icon = 'icons/obj/projectiles.dmi'
```
Glloyd, line 743: can you move these sprites to icons\urist\items\effects.dmi? overwriting Bay icon files is something we try to avoid.

Glloyd (changes requested):

code/modules/urist/mob/spoopymobs.dm
CaraS212: copy that. closing this RP to pack it with the big planet one

PR #1113, opened 2023-02-28
shiptoship crate labels and gun sprite fixes

ggermanicus:
-Egun inhands update appropriately when changing mode and power cells.
-Guns that I saw used old wielded_item_states now use intended ones. Does not fix wield sprites updating, does fix SMGs looking like other SMGs.
-Ship-to-Ship weapons from cargo properly labeled.
-Ithaca shotgun now has a sprite for being unloaded. M3 Grease Gun looks like a machine pistol instead of a rifle inhand. BANG gun displays properly.
-A50 is very off center. It is now as on-center as it can be without respriting it.
-Crew pistol will hopefully display better, if not perfectly due to wield bug.


Glloyd (approved):

Glloyd:
not sure why mapdiffbot is hung up, but it looks all good

PR #1119, opened 2023-03-08
Fixes misaligned console, lights and additional intercom, additional crowbar added.

Laniakeaaaa:
- Fixes some small map errors in Cryo, as one of the lights was overlaying onto the requests console, which has been moved to the side.

- Removes additional intercom

- Adds extra crowbar to emergency locker (For power outages)

image


Glloyd (approved):
honestly thought there already was a crowbar in there.

PR #1126, opened 2023-03-17
adds functionality for mobs to make a sound when they aggro

Glloyd:
As in the title. Adds it to the forgotten. Also slightly nerfs the forgotten on abandoned_colony.dmm


scrdest (approved):
LGTM; not sure if it's really worth having a separate subclass rather than just override projectiletype on the editor level, but your call.

PR #1128, opened 2023-03-26
Raises the Emag's amount of uses before breaking.

Yantarii:
- Considerably increases the Emag's uses to 24, instead of 10.
- The intent behind increasing this is to make it worthwhile spending 32 TC on an item that runs out of uses incredibly quickly. This also gives breathing room to use the Emag more to mess with other systems, such as APCs, Alarms, etc. As you have more charges to use it on.

I forgot to change this during the rebalancing of https://github.com/UristMcStation/UristMcStation/pull/1115


Glloyd:
Imo it should be random, between 18 and 24. The number of remaining uses isn't visible to the user anyways, so it should be a little unpredictable about when it's going to short on you.

Yantarii:
Emag now starts at base value of 18, and will now potentially go up to 24 uses.

PR #1131, opened 2023-04-01
[In-Progress] Expanding Ammo Bundles to accept most bullet types.

Yantarii:
**Drafted until I finish up what's needed.**

Goal:

Allowing for most bullet types to placed into a bundle and used to fill magazines quicker for non-shell fed.
There are some minor balance implications behind this, I've made it currently that you can grab around the amount of ammo needed to fill a magazine, some bundles may have slightly more or less, but that is to account for weapons with a variety of magazines and subtypes, such as the 9MM, ACP, etc.

You won't be able to place 14MM and some other super powerful ammo into fistfuls at the moment, and 7.62 and 5.56 currently have a limit of 10 per hand, this can be changed if that's something people really want.

This also brings minor balance considerations and concerns to revolvers, currently the speedloader is still the fastest way to reload, as it should be, but if you click faster enough you could probably get something similar.

- [ ] Allow Magazines to accept bundles ammo.
- [ ] Balance Checking/Addressing
- [ ] Style


Yantarii:
another time, busy with stuff atm

PR #1138, opened 2023-04-19
Painting Decals

Yantarii:
Adds a fair few paintings as structures/decals that can be added to maps.

image


code/modules/urist/decals/urist.dm
```diff
@@ -45,3 +45,68 @@
...
+
+/obj/structure/sign/painting/stillife
+	name = "\improper still-life painting"
+	desc = "A painting that shows what appears to be fruit on a table in great detail."
+	icon_state = "still"
+
+/obj/structure/sign/painting/persistance
+	name = "\improper surreal painting"
+	desc = "A surreal painting focusing on clocks melting into the landscape, almost trippy to look at."
+	icon_state = "persistance"
```
scrdest, line 92: "Persistence", rather. It's bugging me a bit, but you'd have to fix up the DMI as well.

Yantarii: fixed

PR #1150, opened 2023-08-05
Raises TEG max output to 1500kW

CPU-Blanc:
Seeing as a super simple SM setup can get you 1,000-2,000kW... This 500kW limit is insanely dumb.
Raises that to 3,000kW per TEG before side-effects start to kick in. Should give some wriggle room for people who want to push more power without going tooooo ham

### DIFF
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -11,7 +11,7 @@
 	construct_state = /singleton/machine_construction/default/panel_closed
 	maximum_component_parts = list(/obj/item/stock_parts = 16)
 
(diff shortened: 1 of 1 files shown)


CPU-Blanc:
I've actually tweaked this and adjusted it to 1,500kW.
Although setups that start to push it will  exceed this, it'll only nuke APCs that are _directly_ connected to the TEGs.

So in reality, it just means going over 1,500kW will cause some efficiency loss, which is fine I guess?

Glloyd:
Seems good to me.

PR #1144, opened 2023-07-29
feature: Girlcats

scrdest:
An absurd hellish abomination of a cat with human ears. 
Low dmg, high windup, _fast_. 
Sprites by Ryanwraith w/ my own edits.

,-(_(E.w.)3

girlcat

Based on this classic:
image

EDIT: to clarify, this branch is based off of @ggermanicus 's Baymerge, that's why there's a brazillion commits here right now. Wasn't sure whether to PR straight here or to the Baymerge branch, since it's a new feature.

This code has a hard dependency on the new Bay AI and I CBA to rewrite it back to old AI code temporarily.

PR #1153, opened 2023-08-08
feature: Spawners

scrdest:
Adds simplemob spawners, 'cause Bay ones are broken. Ours are a simple atom/movable, so they're much easier to adminspawn and can be stacked on one tile.

The spawners come in two variants:

- Squad: one-off, spawn 4 mobs from a weighted random distribution. Comes in a ton of flavors, from NTIS to spoop through pirates.
- Endless: spawns a steady stream of mobs, currently by default per-area (rather than in a radius). Same flavors as above with minor exceptions where I didn't feel we'd need 'em currently.

They are invisible to non-observers and even for observers, they use the old S-COM icon, so they're like 4x4 px.
Invisibility is set at runtime, so these can be used for mapping.


Glloyd:
this will conflict with #1144 whenever that gets merged. You'll also need to adjust the code quality checks.

Glloyd:
i still need to familiarize myself more with the new AI stuff, but it looks good to me 👍

PR #1156, opened 2023-08-18
Adds client scanner result output preference + Tweaks

CPU-Blanc:
Bay being Bay have done things half-arsed and left things inconsistent. Currently, performing a scan with a gas analyser will make the scan result appear in a new window rather than posting to chat. But medical scanners will only ever post to chat... This PR rectifies this behaviour to make it consistent. Some players did not like this feature, while others did. So let's make it optional.

- Adds a new client preference `Display scanner results in window`. Scanner results that have the ability to be posted to a new window will be. If disabled, the old legacy behaviour takes effect and everything goes to chat. This defaults to on.
- Removed `html2pencode` conversion in the health and gas PDA scanner code. Not sure why this was ever there, nanoUI is html...
- Added new option to the scanner PDA program; `Notify on output` (I suck at naming things). Always seemed odd to me that PDA scanners required you to open the PDA to see the results. Enabling this option (default on for medical + gas scanners) will have the scan results post to the player's desired output (new window, or chat), rather than only being able to see the results in the program itself. With the exception of the old medical scan proc, that was hardcoded to output to chat for some dumb reason rather than simply returning the data...
- Slight tweaks to the atmospheric scanner proc. Without this, if the client pref is set to output to chat, it's... stupidly large and eats all of chat with huge formatting and unneeded context. This reverts it to post similar output to the chat as it used to before (volume info is new, but can be handy).
- Adds the beep boop noises to the medical scanner on PDAs

### DIFF
--- a/code/game/objects/items/devices/scanners/gas.dm
+++ b/code/game/objects/items/devices/scanners/gas.dm
@@ -17,14 +17,22 @@
 	if(!air_contents)
 		to_chat(user, SPAN_WARNING("Your [name] flashes a red light as it fails to analyze \the [A]."))
 		return
-	scan_data = atmosanalyzer_scan(A, air_contents)
-	show_menu(user)
 
-/proc/atmosanalyzer_scan(atom/target, datum/gas_mixture/mixture)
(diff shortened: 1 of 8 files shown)


CPU-Blanc:
... /var in proc parameters is now a code quality check fail? wtf, why?
Do I really need to change that?

PR #1166, opened 2023-08-29
[Bugfix] Air Control Consoles

CPU-Blanc:
Bay appears to have changed the way they handle air gas sensors; from utilising an assoc list, to instead using a single dedicated variable for both sensor name & tag. This caused essentially all gas sensors accessed via gas control consoles to be broken.
I have noticed however, that there are some remains of the old assoc list style (See ln:64-71 of `atmo_control.dm`), but the rest of the code + UI (see the settings page for the consoles) seem to be directed to only allowing a single sensor. Currently I've stuck with the 1 single sensor route, but I could edit the UI and such at a later date if we want the ability to have multiple.

I've also fixed a bunch of vents and sensors which weren't using the correct subtype, which meant they were on the incorrect frequencies and inaccessible to the control consoles. (Currently, essentially every output vent in atmospherics is unusable)

This fixes all thruster bay output vents + sensors, Nerva fuel tank output vent + sensor, Trajan + Hadrian external sensors, SM core sensor, all atmospherics sensors + output vents.

### DIFF
--- a/code/game/machinery/air_sensor.dm
+++ b/code/game/machinery/air_sensor.dm
@@ -88,7 +88,7 @@
 	frequency = ATMOS_ENGINE_FREQ
 
 /obj/machinery/air_sensor/dist
-	stock_part_presets = list(/singleton/stock_part_preset/radio/basic_transmitter/air_sensor/engine = 1)
+	stock_part_presets = list(/singleton/stock_part_preset/radio/basic_transmitter/air_sensor/dist = 1)
 
-/singleton/stock_part_preset/radio/basic_transmitter/air_sensor/engine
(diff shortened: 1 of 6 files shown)


Glloyd:
some failing map tests re: missing vents still

CPU-Blanc:
Ah. My guess is because the vents in the burn chambers are now correctly tank vents, so cannot be controlled via an air alarm. I'm guessing the tests don't count these (rightly so) and is upset because the thruster rooms don't have air vents in them (they never have).

I can try and map some in if you'd like, though that was a little out of scope of this PR.

Glloyd:
You can add them as an exclusion to the vent checks in that case, in nerva_unit_testing.dm

CPU-Blanc:
Done :+1:

PR #1175, opened 2023-09-05
Adds prep blast door button to bridge and removes storage/cycler access in prep

ggermanicus:
Every round I have to run down and open prep for people, how about I just make a button to do it and remove the access to the suits and cyclers? If we have a massive spike in suit thefts I'll put it back.

In a perfect world I could let the button toggle the cyclers open too but I am not pro enough.

The button is at the Bridge with the lockdown buttons.

### DIFF
--- a/maps/nerva/nerva-2.dmm
+++ b/maps/nerva/nerva-2.dmm
@@ -13460,7 +13460,7 @@
 	name = "Expedition Prep Blast Doors";
 	pixel_x = 5;
 	pixel_y = 5;
-	req_access = list("ACCESS_BRIDGE")
+	req_access = list("ACCESS_EXPEDITION")
 	},
 /obj/machinery/button/blast_door{
 	id_tag = "bridgeview";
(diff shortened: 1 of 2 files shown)


maps/nerva/nerva-2.dmm
```diff
@@ -13455,6 +13455,13 @@
 	pixel_y = -5;
 	req_access = list("ACCESS_BRIDGE")
 	},
+/obj/machinery/button/blast_door{
+	id_tag = "MainExpeditionPrep";
+	name = "Expedition Prep Blast Doors";
+	pixel_x = 5;
+	pixel_y = 5;
+	req_access = list("ACCESS_BRIDGE")
```
Glloyd, line 13463: should this not be expedition access

Glloyd (changes requested):

maps/nerva/nerva-2.dmm
ggermanicus: It could be, but all the other buttons there are bridge access, it's in the bridge, everyone likely to press the button will have bridge access and maybe not prep access. I can change it if you think I should.

Glloyd: Should probably be for consistency, the other buttons affect the bridge and this one doesn't. Plus not everyone with bridge access has expedition access, although most do.

ggermanicus: done

PR #1168, opened 2023-08-30
Bar + Library Z-Level Combination - Baymerge Edition

Yantarii:
Pull Description: Adds a section level of the bar, which is now merged with the Library to give use to both areas.
Moves Command Dorms into the old Library Posistion.

Redone for Baymerge, also tweaked a fair bit of the design, as I didn't like how it looked after not working on it for a while.
No issues w/ Leaky Pipes, Etc in Debug Verbs.

Images: 
jwMqUk9E6R

3hI14Ikfyi

SQDdPaCBEb

DMM View:

image

PR #1176, opened 2023-09-08
Nerva Bar/Library Revamp

Glloyd:
Rescues #1168 from Git hell.

Credit to @Yantarii 

Also does more things! Added more newscasters + status displays (ship still needs more newscasters imo), added some paint and floor decals, and remapped the area around the 3rd deck elevator.


Glloyd:
Since I was messing with the map anyways, I made a few minor changes as well. Added more newscasters + status displays (ship still needs more newscasters imo), added some paint and floor decals, and remapped the area around the 3rd deck elevator

PR #1188, opened 2023-09-10
Coin + Trinket Changes - Loadout

Yantarii:
Vaguely dependant on whether  #1183  is gonna go through

- Removes coin from loadout.
- Moves trinkets/jewelry to trinkets section.


Yantarii:
ah damn, i'll take a look

PR #1189, opened 2023-09-10
feature: DeepmaintRooms

scrdest:
Adds DeepmaintRooms, an alternate take on Deepmaint using WaveFunction Collapse implemented in a Rust DLL for mapgen.

deepmaint-showoff

[DeepmaintDemo1.webm](https://github.com/UristMcStation/UristMcStation/assets/4363639/f414acf6-990d-4c43-8829-c4e360eb4d00)

Currently only spawnable/accessible through adminbus measures.

### Features:

- A liminal horror take on Deepmaint - very rare, if any, face-eating monsters... but plenty of non-Euclidean geometries
- Fast mapgen; the Rust side completes in _miliseconds_ for a 90-tile map, the slight stutter occurs due to BYOND-side integrations
- The generated map is duplicated into four _nearly but not exactly_ identical variants. Variants are connected by invisible, silent, randomly-triggered teleporters, creating the illusion of subtle reality shifts.
- Reality shifts are the main threat in DeepmaintRooms - if you're unlucky, a shift may wall you in.
- Death in Deepmaint returns your dead body back to the station; in case of gibbing, spawns a gibspawner with your DNA instead.
- My best effort at neat modularisation of the added code; all integration can be turned off by unticking one file.

### Mapgen

The map is generated by a Rust DLL written by yours truly and implementing a WaveFunction Collapse algorithm. Hadn't pushed the full source code for the version used here, but it's a slight modification of this branch: https://github.com/scrdest/MorkovChain-maptiler/tree/feature/singlethreaded plus the `byond_fn` crate for integration into, well, BYOND.

The generator is codebase-agnostic and configurable; you simply point it at a JSON ruleset that defines:
- **arbitrary numerical keys** 
- (un-normalized, optionally-direction-specific) Markovian **transition weights** between them
-  **map size** (but see below) 
- optionally, an **adjacency class** (`octile` - all adjacents or `cardinal` - no diagonals, only straight-line neighbors; the former requires either specifying more rules or using undirected weights). 

Missing transition weights are interpreted as zeros.

The interpretation of the map into SS13 is done on the BYOND side, with the keys currently interpreted as pointing to a 3x3 'prefab' datum defining a 'patch' of tiles. This means that a size 30 in the mapgen ruleset becomes a 90x90 map in SS13, a size 20 map becomes a 60x60, and so on.

The generator can be reused with different rulesets for different applications; for example, mapping 1 mapgen tile to 1 SS13 turf with a particular undirected ruleset yields nice planets with plausible-looking coastlines.

### TODOs:

- [x] Remove debug code
- [x] Tooling to teleport people into Deepmaint
- [x] Event to trigger Deepmaintey anomalies to pull people wandering maint in
- [ ] Maybe add some extra junk for variety's sake
- [x] Possibly optimize the generation procs on the DM side to not have to do a locate() over and over or rely on hand-entered Z-level but use caching instead
- [ ] Long-term, extend the mapgen library to be able to extend existing maps nicely or freeze portions and regenerate the rest (e.g. for regenerating only the tiles not visible to the player at the moment)


code/modules/urist/admin/verbs.dm
scrdest, line 1: Forgot to mention that in the commit message; these changes are meant to optimize those operations by turning them asynchronous. We never cared about the return value here, so should behave exactly the same.

code/modules/urist/modules/wfc/core/calllib.dm
```diff
@@ -0,0 +1,99 @@
...
+						if(turf_type && child_turf && (overwrite_all || child_turf.wfc_overwritable))
+							WFC_CHANGE_TURF(child_turf, turf_type)
+
+	return
+
+
+/proc/generate_wfc_map_full(var/rules_json = "deepmaint.json", var/overwrite_all = FALSE, var/zlevel = 1)
+	var/mapname = "genmap.json"
+	generate_wfc_map(rules_json)
+	generate_from_wfc_file(mapname, overwrite_all = FALSE, zlevel = zlevel)
```
CPU-Blanc, line 97: Is this necessary? The argument name matches the parameter name and we're not skipping over any.. Seems a little odd to me
```suggestion
	generate_from_wfc_file(mapname, overwrite_all = FALSE, zlevel)
```

code/modules/urist/modules/wfc/_module.dm
```diff
@@ -0,0 +1,32 @@
...
+
+#else
+
+#include "fractal_coords.dm"
+#include "modules\deepmaint\hooks.dm"
+
+#endif
+
+#include "core\calllib.dm"
+//#include "core\debug.dm"
```
CPU-Blanc, line 22: Could be an idea to put this behind some kind of debug `#ifdef`? Along with more verbose logging, like we did in the GOAI project?

code/modules/urist/modules/wfc/core/turf.dm
```diff
@@ -0,0 +1,40 @@
...
+
+/proc/WfcChangeTurf(var/turf/trg, var/new_type)
+	if(isnull(new_type))
+		return
+
+	if(!ispath(new_type))
+		return
+
+	if(!ispath(new_type, /turf))
+		return
```
CPU-Blanc, line 26: Seems like some redundant checks here. `ispath(new_type, /turf)` would return false if it is not a path, or if it were null.
```suggestion
	if(!ispath(new_type, /turf))
		return
```

code/modules/urist/modules/wfc/core/turf.dm
```diff
@@ -0,0 +1,40 @@
...
+
+	if(!isnull(trg?.associated_wfc_overwritables))
+		for(var/overwritten in trg.associated_wfc_overwritables)
+			qdel_from_weakref(overwritten)
+
+	trg.icon_state = initial(trg.icon_state)
```
CPU-Blanc, line 32: Seems odd we're accounting for a null `trg` above, but not here?

code/modules/urist/modules/wfc/core/turf.dm
```diff
@@ -0,0 +1,40 @@
...
+
+	#ifdef STUB_OUT_SS13
+	var/turf/new_turf = new new_type(trg)
+	#else
+	trg.ChangeTurf(new_type)
```
CPU-Blanc, line 37: Same as above. No null check but one prior. Is this intentional?

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
```diff
@@ -0,0 +1,36 @@
+// Adminbuse stuff for deepmaint
+
+/datum/admins/proc/generate_deepmaint()
+	set category = "Fun"
+	set desc = "(Re-)Generate DeepmaintRooms; place the map template first!"
+	set name = "Generate DeepmaintRooms"
+
+	if(!GLOB.wfc_deepmaint_zlevels_by_instance)
+		log_and_message_admins("- no Deepmaint Z-levels registered!")
```
CPU-Blanc, line 9: This log is lacking some context and might be out of place for other staff. Would a simple `to_chat` with error span class be more suitable?

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
```diff
@@ -0,0 +1,36 @@
...
+		return
+
+	for(var/area/map_template/deepmaint_wfc/deepmaint_area in GLOB.wfc_deepmaint_zlevels_by_instance)
+		deepmaint_area.generate_deepmaint("deepmaint.json", "genmap.json")
+		log_and_message_admins("has generated a DeepmaintRooms map.")
+		break
```
CPU-Blanc, line 15: Seems a little odd to break on the first iteration without any kind of checks. Could you not just access `wfc_deepmaint_zlevels_by_instance` by the first index instead?
```suggestion
	var/area/map_template/deepmaint_wfc/deepmaint_area = GLOB.wfc_deepmaint_zlevels_by_instance[1]
	deepmaint_area.generate_deepmaint("deepmaint.json", "genmap.json")
	log_and_message_admins("has generated a DeepmaintRooms map.")
```

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
```diff
@@ -0,0 +1,36 @@
...
+/datum/admins/proc/throw_into_deepmaint(var/mob/M)
+	set category = "Fun"
+	set desc = "Throws the target into Deepmaint."
+	set name = "Send to DeepmaintRooms"
+
+	if(!GLOB.wfc_deepmaint_zlevels_by_instance)
+		log_and_message_admins("tried to throw into Deepmaint, but no Deepmaint Z-levels are currently registered!")
+		return
+
+	var/atom/movable/target = M
```
CPU-Blanc, line 29: Any reason not to typecast this in the proc parameters instead of reassigning like this? I don't _think_ the admin datum type checks when calling the proc, so would a `istype(M)` be required in place of the `isnull` on the next line?

CPU-Blanc (commented):
I'm sometimes smoothbrain, but I thought a smoothbrain second pair of eyes are better than nothing?

Just some little bits I spotted, not gone through all of it just yet but it's getting late

code/modules/urist/admin/verbs.dm
```diff
@@ -11,13 +12,14 @@
 		M.loc = get_turf(usr)
 		message_admins("[key_name(usr)] has warped all players to their location.")
 		log_admin("[key_name(src)] has warped all players to their location.")
```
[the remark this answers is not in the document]
scrdest: Oh, I think you're right; probably part of the reason this is a hella heavy proc.

code/modules/urist/modules/wfc/_module.dm
scrdest: I think that's redundant in this case, if you want these you can just uncomment the line or manually include debug. It's mostly debug _verbs_ anyway.

code/modules/urist/modules/wfc/_wfc_defines.dm
```diff
@@ -0,0 +1,43 @@
+# define STEP_TRIGGER_INVISIBILITY 101
+
+#define DEEPMAINT_WALL_ICON 'icons/urist/turf/walls.dmi'
+#define DEEPMAINT_WALL_ICONSTATE_BASE "arust"
+#define DEEPMAINT_WALL_ICONSTATE_DYNAMIC(FromVar) "arust[FromVar]"
+
+#define DEEPMAINT_FLOOR_ICON 'icons/turf/flooring/plating.dmi'
+#define DEEPMAINT_FLOOR_ICONSTATE "plating"
+
+#define WFC_CHANGE_TURF(From, To) WfcChangeTurf(From, To)
```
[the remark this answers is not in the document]
scrdest: It was originally different between the two projects and in the two ifdef branches below. I kept it here as a way of having a zero-cost abstraction over the implementation if things change _again_.

code/modules/urist/modules/wfc/core/calllib.dm
scrdest: I generally prefer named arguments, especially for optional params - not only is it clearer, it's also robust to refactoring changing the order of optional args.

Also, I'm used to Python screaming at you for putting positionals after keyword args.

code/modules/urist/modules/wfc/core/turf.dm
scrdest: I left those in as placeholders for more specific debug logging ("oh, it failed because it's a null/not a turf/an instance")

code/modules/urist/modules/wfc/core/turf.dm
scrdest: Good spot, oversight.

code/modules/urist/modules/wfc/core/turf.dm
scrdest: Nope :^)

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
scrdest: IIRC the method logs what proc was called by whom; it's a weird API

scrdest: Also this follows the convention of other Template-adjacent stuff IIRC

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
scrdest: No, because _you're not guaranteed to have an index 1_. The current code handles a failure gracefully and doesn't report success on a failure.

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
scrdest: IIRC I did this to extend this to all atoms later; it was just annoying to have this verb pop up for all atoms, for time being.

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
CPU-Blanc: Normal logs do not include calliing procs, liines etc. Only `WARNING`, `crash_with` and `ARGS_DEBUG` insert the calling proc and line automatically.
`log_and_message_admins` just inserts the calling user/mob with the `(?)` href link

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
CPU-Blanc: I actually thought about this after I turned the PC off. Fair point!

code/modules/urist/admin/verbs.dm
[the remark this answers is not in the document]
CPU-Blanc: Also this can just be a single call to` log_and_message_admins`, dropping the `key_name` as it's automatically inserted

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
```diff
@@ -0,0 +1,143 @@
...
+
+	if(isnull(trg.loc))
+		// No un-qdel-ing by accident!
+		return
+
+	if(istype(trg, /turf) || istype(trg, /area))
+		// This would be silly
+		return
+
+	if(istype(trg, /mob) && !istype(trg, DEEPMAINT_TELEPORTABLE_MOBTYPE))
```
CPU-Blanc, line 14: Not sure what DM does under the hood, but I'd hope using a built-in proc would be better for this?
```suggestion
	if(ismob(trg) && !istype(trg, DEEPMAINT_TELEPORTABLE_MOBTYPE))
```

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
```diff
@@ -0,0 +1,143 @@
...
+		// Don't teleport AI eyes, ghosts, irate admins...
+		return
```
CPU-Blanc, line 48: Same as above comments

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
```diff
@@ -0,0 +1,143 @@
...
```
CPU-Blanc, line 82: As above

code/modules/urist/modules/wfc/modules/deepmaint/turfs.dm
```diff
@@ -0,0 +1,146 @@
...
+	. = ..(Obj)
+
+	spawn(src.wfc_teleporter_delay)
+		var/teleported = FALSE
+
+		if(!teleported && src.wfc_has_stationexit)
+			teleported = deepmaint_conditional_stationyeet(Obj, tele_proba = src.wfc_teleproba_station)
+
+		if(!teleported && src.wfc_has_variantlink)
+			teleported = deepmaint_conditional_yeet(Obj, tele_proba = src.wfc_teleproba_deepmaint)
```
CPU-Blanc, line 74: Updating `teleported` here doesn't seem to do anything as it's immediately discarded. Is this intentional?

code/modules/urist/modules/wfc/modules/deepmaint/turfs.dm
```diff
@@ -0,0 +1,146 @@
...
+	// % chance of disappearing
+	var/improbability = 10
+
+
+/turf/deepmaint/wall/probably/New(var/atom/loc, var/improbability_override = null)
+	var/wall_improba = (isnull(improbability_override) ? src.improbability : improbability_override)
+
+	if(prob(wall_improba))
+		spawn(0)
+			WFC_CHANGE_TURF(src, /turf/deepmaint/plating)
```
CPU-Blanc, line 143: Will this not cause some init issues? iirc it hates it when stuff are qdel'd before being initialised as a call to parent is never made here. Might be wrong, but I remember having similar issues with landmarks

ss13_wfc.dll
CPU-Blanc, line 1: Any thoughts of open-sourcing this dll? Is there a way we could open a new repo under Urist for this/fork it from your repo?

CPU-Blanc (commented):
The next batch of smoothbrain comments. Ignore what are irrelevant, but just some general comments I had

code/modules/urist/modules/wfc/modules/deepmaint/generation.dm
```diff
@@ -0,0 +1,47 @@
...
+	var/mapname = "genmap.json"
+	var/_sleeptime = max(1, sleeptime)
+
+	spawn(0)
+		while(TRUE)
+			deepmaint_rebuild_mapfile(ruleset)
+			deepmaint_from_mapfile(mapname, overwrite_all = FALSE, zlevel = zlevel)
+			to_chat(world, "Regenerating in 20s")
+			sleep(_sleeptime)
+			to_chat(world, "Regenerating!")
```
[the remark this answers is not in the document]
CPU-Blanc: Also left in the output to `world` :P 
Might want to make this a debug log instead
```suggestion
	spawn(0)
		while(TRUE)
			deepmaint_rebuild_mapfile(ruleset)
			deepmaint_from_mapfile(mapname, overwrite_all = FALSE, zlevel = zlevel)
			log_debug("Deepmaint - Regenerating in 20s")
			sleep(_sleeptime)
			log_debug("Deepmaint - Regenerating!")
```

code/modules/urist/modules/wfc/modules/deepmaint/hooks.dm
```diff
@@ -0,0 +1,31 @@
+/hook/death/proc/deepmaint_death_handler(var/mob/living/carbon/human/H, var/gibbed = FALSE)
+	var/mob/living/carbon/human/deadman = H
```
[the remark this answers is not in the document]
scrdest: IIRC, DM doesn't _actually_ validate input params, IIRC, so you can wind up runtiming by calling procs on a type that does not have them. 

The cast in L2 here is doing the heavy lifting of nulling the input out if it's not the expected type so that we can do null checks on it to catch both mis-typed and explicitly null inputs.

code/modules/urist/modules/wfc/modules/deepmaint/hooks.dm
```diff
@@ -0,0 +1,31 @@
...
+	if(isnull(T))
+		return FALSE
+
+	if(gibbed)
+		gibs(T, deadman.dna)
+		T.visible_message("A pile of horribly misshapen organs, meat, and blood sprays out of thin air.", "You hear a series of disgustingly wet thuds and squelches.")
+
+	else
+		H.forceMove(T)
+		H.visible_message("[H] materializes on [T], seemingly out of thin air.")
```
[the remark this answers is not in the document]
scrdest: Yeah, good shout. I don't trust text macros much so I forget about them 😅

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
```diff
@@ -0,0 +1,143 @@
+
+/proc/deepmaint_yeet_trg(var/atom/movable/trg, var/xoffset = 1, var/yoffset = 1)
+	if(!istype(trg, /atom/movable))
```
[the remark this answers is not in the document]
scrdest: Maybe, but I find it much less readable. You need to mentally disambiguate between the is-implicit-type case and the is-SOME-type case. Because we cannot have sane APIs in BYOND.

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
```diff
@@ -0,0 +1,143 @@
...
+		return FALSE
+
+	if(isnull(trg.loc))
+		// No un-qdel-ing by accident!
+		return
```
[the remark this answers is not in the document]
scrdest: Eh, maybe, but this would be more painful to port out of SS13 for standalone testing and probably catch less cases.

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
scrdest: I didn't realize this was an actual builtin, since we have a couple of macros that pretend to be builtins.

code/modules/urist/modules/wfc/modules/deepmaint/turfs.dm
scrdest: Yeah; maintainability. This way you can swap the priority of the two checks, add prob() and whatever, and it will still work. 

The intended behavior is that the first successful teleport prevents any further teleports without messing with early returns which can make cleanup logic hard.

ss13_wfc.dll
scrdest: Yeah, I'm planning to; the trick is that this a wrapper Rust library I wrote on top of my _second_ Rust library... and then to top it all off, the SS13 bits are on an _un-pushed branch_ of that second library so the first one is pointed at my _local filesystem_ ATM 🤡 

The base library ('second' above) is this: https://crates.io/crates/morkovmap_rc

code/modules/urist/modules/wfc/modules/landmass/prefabs.dm
```diff
@@ -0,0 +1,110 @@
...
+	bl = /turf/snow
+	bm = /turf/snow
+	br = /turf/snow
+
+
+/proc/state2prefab_landmass(var/state, var/my_varkey)
+	var/turf_type = null
+
+	switch(state)
+		if(1)
```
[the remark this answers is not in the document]
scrdest: This is just me being lazy honestly... plus this is only ever used in this one proc to remap from the JSON output to prefabs and will never be needed elsewhere. 

The overhead of checking what the macro maps to vs what's in the JSON might well be higher than if we keep it plaintexted - it's already obvious what _prefab_ the ID maps to.

code/modules/urist/modules/wfc/modules/deepmaint/hooks.dm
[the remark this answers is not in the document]
CPU-Blanc: Yep, DM doesn't validate types passed to parameters. But iirc, assigning like this _doesn't_ null on incorrect types (might require validation, but I'm pretty sure that's how DM works). Pretty sure DM is kinda dumb like that. You'd have to do a `istype(H)` to catch both nulls and incorrect types

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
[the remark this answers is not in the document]
CPU-Blanc: Fair. I think it's personal preference. For _me_ `istype(trg)` tells me I'm checking if what I have is what I expected, whereas `istype(trg, /atom/movable)` is checking if it's a specific child type of what we currently have.

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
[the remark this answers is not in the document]
CPU-Blanc: Are there cases where a datum might be marked for GC but still have a loc, however? That could be missed here if so

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
CPU-Blanc: Yeah. All the _basic_ datums have built-in procs: `ismob()`, `isobj()`, `isturf()`, `isarea()` and also `ismovable()`. As you said, there _are_ a load that are just `istype` checks wearing coats, like `isAI()`

code/modules/urist/modules/wfc/modules/deepmaint/turfs.dm
CPU-Blanc: Makes sense :+1:

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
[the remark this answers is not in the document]
CPU-Blanc: _This could also be just replaced with the built-in `ismovable(trg)`_

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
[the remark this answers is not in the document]
scrdest: Shouldn't be a problem in that case; even if we teleport it, it will be moved into nullspace by the GC, just from a different location. 

The primary purpose of this is to avoid accidentally creating new refs by moving things out of nullspace, whether it's there due to GC or some nonsense (e.g. being inside an object which then gets deleted, but the logic doesn't transitively delete the contents fsr).

code/modules/urist/modules/wfc/core/turf.dm
scrdest: Done; we will have checked it's not null, me using the null-conditional was reflexive - but if we do use it needlessly, that's probably some overhead.

code/modules/urist/admin/verbs.dm
```diff
@@ -1,23 +1,25 @@
...
 		return
 
 	for(var/mob/living/M in GLOB.player_list)
-		M.loc = get_turf(usr)
-		message_admins("[key_name(usr)] has warped all players to their location.")
-		log_admin("[key_name(src)] has warped all players to their location.")
+		M.forceMove(get_turf(usr))
+		sleep(-1)
+
+	log_and_message_admins(" has warped all players to their location.")
```
CPU-Blanc, line 15: Minor nit-pick, but I'd personally prefer this log to be _before_ the `for` loop. If there's any lag it'll give a better indication as to what has caused it. Also iirc you don't need that leading space within the string, that's already added in the proc

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
```diff
@@ -0,0 +1,36 @@
+// Adminbuse stuff for deepmaint
+
+/datum/admins/proc/generate_deepmaint()
+	set category = "Fun"
+	set desc = "(Re-)Generate DeepmaintRooms; place the map template first!"
+	set name = "Generate DeepmaintRooms"
+
+	if(!GLOB.wfc_deepmaint_zlevels_by_instance)
+		log_and_message_admins("- failed to generate Deepmaint - no Deepmaint Z-levels registered!")
```
CPU-Blanc, line 9: Super nit-pick. Leading Hyphen might look a little odd and out of place with the admin log

code/modules/urist/modules/wfc/modules/deepmaint/generation.dm
```diff
@@ -0,0 +1,47 @@
...
+	var/mapname = "genmap.json"
+	var/_sleeptime = max(1, sleeptime)
+
+	spawn(0)
+		while(TRUE)
+			deepmaint_rebuild_mapfile(ruleset)
+			deepmaint_from_mapfile(mapname, overwrite_all = FALSE, zlevel = zlevel)
+			log_debug("Deepmaint - Regenerating in 20s")
+			sleep(_sleeptime)
+			log_debug("Deepmaint - Regenerating!")
```
CPU-Blanc, line 46: I see logging got fixed but anything re: the dangling `while(TRUE)` loop concern? Or is not an issue with how DM handle their wonk async?

CPU-Blanc (commented):
Last few nit-picks/questions from me

CPU-Blanc:
Just need to add `/area/map_template/deepmaint_wfc` to `maps/away_sites_testing/template_testing_unit_testing.dm` in the `area_usage_test_exempted_root_areas` list to fix that test fail. Not sure why it didn't fail before and only now..

code/modules/urist/modules/wfc/modules/deepmaint/generation.dm
scrdest: Sorry - I remembered seeing this comment and I've literally double-checked but couldn't find your original comment, so I thought you found it's fine yourself and deleted it.

I've had a hunch that it's fine, and it is - BYOND being a lovely mess that it is, that fact is hidden in the documentation for `del()`:

```Procs that are executing with src equal to that object are silently killed, causing execution to return to the caller. If that is not what you want, you should detach the proc from the source object by setting src to null.```

Since this is indeed attached to an object (the area), it should be fine.

File: code/modules/urist/modules/wfc/modules/deepmaint/turfs.dm (hunk 15 above)
Comment (scrdest) [reply code/modules/urist/modules/wfc/modules/deepmaint/turfs.dm:143]: I've ran this, looked fine. 

The trick is, I don't think this _actually_ isn't initialized by the time of qdel(). The reason being that `spawn(0)` - anything behind it gets put at the _end_ of the queue of pending futures to schedule, while negative values prepend it at the _front_ of the queue.

So, the scheduler ordering goes `New() -> ...any spawn(-1)'d code... -> ...any prior spawn(0)'d code... ->WFC_CHANGE_TURF() -> ...any spawn(0)'d code after this...`

**Comment by scrdest** [comment]
@CPU-Blanc I've pushed the source for the libcode I've used for the DLLs here: https://github.com/scrdest/MorkovChain-maptiler/tree/feature/ss13

The actual bindings to DM are in a separate crate, but the source is so extremely tiny I can just specify this inline:
```Rust
use morkovmap_rc;
use byond_fn::byond_fn;

#[byond_fn]
fn wfc() {
    morkovmap_rc::generate_from_file(None)
}

#[byond_fn]
fn from_ruleset(ruleset: &str) {
    morkovmap_rc::generate_from_file(Some(ruleset))
}
```
The only deps are this lib and the byond_fn crate's macro for cajoling the calling convention stuff because I CBA to do it manually.

CPU-Blanc:
> @CPU-Blanc I've pushed the source for the libcode I've used for the DLLs here: https://github.com/scrdest/MorkovChain-maptiler/tree/feature/ss13
> 
> The actual bindings to DM are in a separate crate, but the source is so extremely tiny I can just specify this inline:
> 
> ```rust
> use morkovmap_rc;
> use byond_fn::byond_fn;
> 
> #[byond_fn]
> fn wfc() {
>     morkovmap_rc::generate_from_file(None)
> }
> 
> #[byond_fn]
> fn from_ruleset(ruleset: &str) {
>     morkovmap_rc::generate_from_file(Some(ruleset))
> }
> ```
> 
> The only deps are this lib and the byond_fn crate's macro for cajoling the calling convention stuff because I CBA to do it manually.

Haven't had time tonight to look at the source much just yet, but couldn't you make it a workspace, include the wrapper crate in it and make it depend on the wfc crate? That way both are together on the same repo, and you could build either one independently (wrapper would obviously build both)?

Ooor probably better, make this a crate feature and gate it behind that? ie
```toml
#Cargo.toml
[dependencies]
byond_fn = { version = "blaah", optional = true } 

[features]
byond = ["dep:byond_fn"]
```
```rust
//lib.rs
#[cfg(feature = "byond")]
#[byond_fn]
fn wfc() {
    morkovmap_rc::generate_from_file(None)
}

#[cfg(feature = "byond")]
#[byond_fn]
fn from_ruleset(ruleset: &str) {
    morkovmap_rc::generate_from_file(Some(ruleset))
}
```

Not done a lot of Rust packaging and mostly dealt with application work, so I might be totally off-base here :P

scrdest:
> Haven't had time tonight to look at the source much just yet, but couldn't you make it a workspace, include the wrapper crate in it and make it depend on the wfc crate? That way both are together on the same repo, and you could build either one independently (wrapper would obviously build both)?

Well, I probably _could_, but even then I wouldn't _want_ to. 

Separation of concerns. The WFC crate is meant to be a very generic Rust library for use by... whoever. I don't want to pollute a general-purpose library with an extremely niche use-case.

The integration to SS13 hasn't even been an afterthought, it has been a random lightbulb moment as I was already starting to just port the code to pure DM.

CPU-Blanc:
> > Haven't had time tonight to look at the source much just yet, but couldn't you make it a workspace, include the wrapper crate in it and make it depend on the wfc crate? That way both are together on the same repo, and you could build either one independently (wrapper would obviously build both)?
> 
> Well, I probably _could_, but even then I wouldn't _want_ to.
> 
> Separation of concerns. The WFC crate is meant to be a very generic Rust library for use by... whoever. I don't want to pollute a general-purpose library with an extremely niche use-case.
> 
> The integration to SS13 hasn't even been an afterthought, it has been a random lightbulb moment as I was already starting to just port the code to pure DM.

That's why putting behind a feature gate would work perfectly. Those who don't want the niche byond integration won't  get it, but it's ready to go for those that do. The `byond_fn` dependency won't download unless the feature flag is toggled, and the code won't  be compiled and added to the .dll on build. I don't really see a downside other than the 2KB or so it'd take up in the raw source file?
Just seems really silly having such a tiny wrapper be its own crate imo

scrdest:
Eh, I'd rather keep it this way and potentially extend it later on with more complicated WFC use-cases (like conditional generation API to freeze some tiles as open to force path generation or w/e) and non-WFC features for SS13. 

There's a _lot_ of cool shit Rust DLLs could do, especially if we had the 515+ native FFI rather than the shitty string FFI. I have my eyes on a couple of scientific crates, e.g. one that simulates chemical reactions stochastically, for instance. And that's without getting into atmos and other such nonsense.

baystation12.dme
```diff
@@ -3488,6 +3488,21 @@
 #include "code\modules\urist\modules\weather\weather_helpers.dm"
 #include "code\modules\urist\modules\weather\weather_objects.dm"
 #include "code\modules\urist\modules\weather\weather_turf.dm"
+#include "code\modules\urist\modules\wfc\_wfc_defines.dm"
+#include "code\modules\urist\modules\wfc\fractal_coords.dm"
+#include "code\modules\urist\modules\wfc\core\calllib.dm"
+#include "code\modules\urist\modules\wfc\core\maptricks.dm"
+#include "code\modules\urist\modules\wfc\core\prefabs.dm"
```
CPU-Blanc, line 3495: Not sure what's going on here. But I think your includes have got very mixed up?
Many of these are included in `code/modules/urist/modules/wfc/_module.dm` already, however that file isn't ever included anywhere? Some of these includes are behind `ifdef`s too. Are these all supposed to be in the dme, rather than a single module file which then further includes?

Might need a pass?

scrdest: I left a comment in the module file and the commit message 😅 

TL;DR - DM was acting weird with call()s on code included from the submodule, and this had been hanging around for a long time now so I'd rather get it in and fix the includes separately.

The current `_module.dm` is NOT included to avoid double-including and the tick-included featureset corresponds to the correct SS13 set of includes.

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
```diff
@@ -0,0 +1,42 @@
...
+	set desc = "(Re-)Generate DeepmaintRooms; place the map template first!"
+	set name = "Generate DeepmaintRooms"
+
+	if(!GLOB.wfc_deepmaint_zlevels_by_instance)
+		log_and_message_admins("failed to generate Deepmaint - no Deepmaint Z-levels registered!")
+		return
+
+	for(var/area/map_template/deepmaint_wfc/deepmaint_area in GLOB.wfc_deepmaint_zlevels_by_instance)
+		var/success = deepmaint_area.generate_deepmaint("deepmaint.json", "genmap.json")
+		if(success)
```
CPU-Blanc, line 14: `success` var is never read. Could we not save an allocation?
```suggestion
		if(deepmaint_area.generate_deepmaint("deepmaint.json", "genmap.json"))
```

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
```diff
@@ -0,0 +1,42 @@
...
+			log_and_message_admins("has generated a DeepmaintRooms map.")
+			break
+		else
+			log_and_message_admins("has unsuccessfully attempted to generate a DeepmaintRooms map.")
```
CPU-Blanc, line 18: Would these not make more sense outside of the loop, to avoid spam in case of concurrent failures? Would also give a reason to use the `success` var defined in the previous comment

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
```diff
@@ -0,0 +1,42 @@
...
+
+	return
+
```
CPU-Blanc, line 21: Redundant
```suggestion
```

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
```diff
@@ -0,0 +1,42 @@
...
+
+/datum/admins/proc/throw_into_deepmaint(var/mob/M)
+	set category = "Fun"
+	set desc = "Throws the target into Deepmaint."
+	set name = "Send to DeepmaintRooms"
+
+	if(!GLOB.wfc_deepmaint_zlevels_by_instance)
+		log_and_message_admins("tried to throw [M] into Deepmaint, but no Deepmaint Z-levels are currently registered!")
```
CPU-Blanc, line 29: Text macro might make sense here, as it's for general mobs?
```suggestion
		log_and_message_admins("tried to throw \a [M] into Deepmaint, but no Deepmaint Z-levels are currently registered!")
```

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
```diff
@@ -0,0 +1,42 @@
...
+		return
+
+	var/atom/movable/target = M
+	if(isnull(target))
+		return
+
+	var/success = deepmaint_send_trg(target)
+	if(success)
```
CPU-Blanc, line 37: Again, unnecessary allocation
```suggestion
	if(deepmaint_send_trg(target))
```

code/modules/urist/modules/wfc/modules/deepmaint/generation.dm
```diff
@@ -0,0 +1,51 @@
+/proc/deepmaint_rebuild_mapfile(var/ruleset = "deepmaint.json")
+	var/success = generate_wfc_map(ruleset)
+	return success
+
+/area/map_template/deepmaint_wfc/proc/deepmaint_rebuild_mapfile(var/ruleset = "deepmaint.json")
+	var/success = generate_wfc_map(ruleset)
+	return success
```
CPU-Blanc, line 7: Are these procs actually necessary at all? `generate_wfc_map()` already returns a boolean, so these wrapper procs are adding nothing at all other than a name change. Can we not just call `generate_wfc_map(ruleset)` where necessary instead as there's no extra logic here?

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
```diff
@@ -0,0 +1,151 @@
...
+	while(tries --> 0)
+		var/turf/T = pick_area_turf(/area/map_template/deepmaint_wfc/alpha, predicates)
+
+		if(isnull(T))
+			continue
+
+		var/mob/M = trg
+		if(istype(M))
+			M.Stun(1)
+			M.flash_eyes(intensity = FLASH_PROTECTION_MAJOR * 10, override_blindness_check = TRUE, affect_silicon = TRUE, visual = FALSE, type = /obj/screen/fullscreen/blackout)
```
CPU-Blanc, line 58: Nothing major, just a little peeve. As these parameters are quite long, could we newline this for better readability?
```suggestion
			M.flash_eyes(
				intensity = FLASH_PROTECTION_MAJOR * 10, 
				override_blindness_check = TRUE,
				affect_silicon = TRUE,
				visual = FALSE,
				type = /obj/screen/fullscreen/blackout
			)
```

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
```diff
@@ -0,0 +1,151 @@
...
+
+		var/turf/T = pick_area_turf(location, predicates)
+
+		if(isnull(T))
+			continue
+
+		var/mob/M = trg
+		if(istype(M))
+			M.Stun(1)
+			M.flash_eyes(intensity = FLASH_PROTECTION_MAJOR * 10, override_blindness_check = TRUE, affect_silicon = TRUE, visual = FALSE, type = /obj/screen/fullscreen/blackout)
```
CPU-Blanc, line 99: As above

maps/deepmaint_wfc/wfc_deepmaint.dm
```diff
@@ -0,0 +1,25 @@
...
+
+/obj/effect/landmark/map_data/deepmaint_wfc/lvl1
+	name = "Maintrooms"
+	height = 1
+
+
+/datum/map_template/ruin/deepmaint_wfc
+	name = "Maintrooms"
+	id = "deepmaint_wfc"
+	description = "Somewhere inbetween. How did we get here? How do we leave?"
```
CPU-Blanc, line 23: Time to be picky with grammar :D
```suggestion
	description = "Somewhere in-between. How did we get here? How do we leave?"
```

CPU-Blanc (commented):
Latest batch of smooth-brain comments :D

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
scrdest: It's a stack value, it shouldn't matter - whether you assign it to a var or not, it's _there_ in the memory already (otherwise you couldn't if on it).

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
scrdest: Yes, but I hate implicit returns; you have to think more to guess the return value and it makes refactoring uglier if a diff accidentally joins two procs together.

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
scrdest: Sure, why not?

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
scrdest: As above, TL;DR suggestion does not save memory and makes refactoring later harder.

code/modules/urist/modules/wfc/modules/deepmaint/teleporters.dm
scrdest: Sure, more readability = more good

code/modules/urist/modules/wfc/modules/deepmaint/generation.dm
scrdest: Not strictly needed, no. They are largely a legacy of a weird bug I've seen with the module-style includes.

For some reason callproc would flat out ignore 'functional' procs from module-included code but would recognize 'method' procs from it... so I bound it to the area.

Partially, I've also kept it in as an OOP-ish API into the underlying functional API, pytorch-style.

maps/deepmaint_wfc/wfc_deepmaint.dm
scrdest: I'll change it, but it's a direct copypasta from Bay Deepmaint as-is.

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
scrdest: Fair shout, but as-is it gives you an indication of how badly messed up it is (normally, it should work on first try; if you see multiple failures, something has gone off the rails.

code/modules/events/event_container.dm
```diff
@@ -176,6 +176,7 @@ var/global/list/severity_to_string = list(EVENT_LEVEL_MUNDANE = "Mundane", EVENT
 		new /datum/event_meta(EVENT_LEVEL_MODERATE, "Drone Uprising",						/datum/event/rogue_maint_drones,		25,		list(ASSIGNMENT_ENGINEER = 30)),
 		new /datum/event_meta(EVENT_LEVEL_MODERATE, "Supermatter Power Surge",				/datum/event/power_surge,				100,	list(ASSIGNMENT_ENGINEER = 10)),
 		new /datum/event_meta(EVENT_LEVEL_MODERATE, "Bluespace Drive Instability",			/datum/event/bsd_instability,			50),
+		new /datum/event_meta(EVENT_LEVEL_MODERATE, "DeepmaintRooms",						/datum/event/backrooms_deepmaint,		50),
```
CPU-Blanc, line 179: Should this really be a moderate event? I've not poked the event code much so please do correct any misconceptions, but when this event ends, aren't all people stuck in deepmaint just... SOL and trapped forever?

code/modules/events/event_container.dm
```diff
@@ -189,7 +190,8 @@ var/global/list/severity_to_string = list(EVENT_LEVEL_MUNDANE = "Mundane", EVENT
 		new /datum/event_meta/no_overmap(EVENT_LEVEL_MAJOR, "Electrical Storm",		/datum/event/electrical_storm, 		0,	list(ASSIGNMENT_ENGINEER = 10, ASSIGNMENT_JANITOR = 5)),
 		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Drone Revolution",				/datum/event/rogue_maint_drones,	0,	list(ASSIGNMENT_ENGINEER = 10,ASSIGNMENT_MEDICAL = 10,ASSIGNMENT_SECURITY = 10)),
 		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Spider Infestation",				/datum/event/spider_infestation, 	0,	list(ASSIGNMENT_SECURITY = 10), 1),
-		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Bluespace Drive Instability",	/datum/event/bsd_instability,		0,		null, 1),
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Bluespace Drive Instability",		/datum/event/bsd_instability,		0,	null, 1),
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "DeepmaintRooms",					/datum/event/backrooms_deepmaint,	0,	null, 1),
```
CPU-Blanc, line 194: Again, seems weird to have this in both event categories where the only difference here is it'll only happens once. Is this an oversight?

code/modules/urist/modules/wfc/modules/deepmaint/event.dm
```diff
@@ -0,0 +1,109 @@
+/datum/event/backrooms_deepmaint
+	announceWhen = 10
+	endWhen = 1000
+	var/entrypoints = 100
+	var/list/spawned_teleporters
+
+
+/datum/event/gravity/setup()
+	endWhen = rand(600, 6000)
```
CPU-Blanc, line 9: Seems odd to overload the gravity event here. Error when copying an existing event as a template?
```suggestion
/datum/event/backrooms_deepmaint/setup()
	endWhen = rand(600, 6000)
```

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
CPU-Blanc: Eh. I personally stick to whatever code style the rest of the code uses, which in this case is implicit returns if you're not returning a var directly. I'm not fussed too much either way, but it's not really done anywhere else so stands out.

I believe this has been addressed in 515 though. iirc you can, much in the style of python, define a return type for a proc.

code/modules/urist/modules/wfc/modules/deepmaint/generation.dm
CPU-Blanc: iirc DM _sucks_ when it comes to proc calls and their related overhead. If you want to keep them that's fine, but for me it just feels like unnecessary clutter if you're circumventing the whole module system anyway

code/modules/urist/modules/wfc/modules/deepmaint/admins.dm
CPU-Blanc: Eh, do we really need to spam all admins with it though? Each call is also an I/O write to file iirc. Could you not make a counter var if you want to see how badly it screwed up, and count the amount of attempts and log that instead?

code/modules/urist/modules/wfc/modules/deepmaint/event.dm
scrdest: Oh, lol, yeah

code/modules/events/event_container.dm
scrdest: They aren't, generally. Even during the event, you can escape. This is not a property of the event, just of Deepmaint itself. 

All standard Deepmaint turfs have a small chance of having a hidden station teleport trait, which itself has a probability of triggering on each crossing.

Why the double-probability? Well, for one, it keeps the number of rolls down, but mostly - this is the main actual _threat_. Not monsters or other hostiles (even though they may rarely be added eventually) - there is a small-moderate chance you get walled in a section with no station teleporters in it.

Even then, if you die, you will get yeeted to regular maint.

code/modules/events/event_container.dm
scrdest: TBH I was copying BSD Instability above :^)

code/modules/urist/modules/wfc/modules/deepmaint/generation.dm
scrdest: It it, yeah, but even worst-case scenario, this is not on the hot path anywhere (except possibly periodic deepmaint rebuild, which is not even currently used by anything).

I would like to see what API people use more, initially.

code/modules/events/event_container.dm
CPU-Blanc: Oh. DERP. Just saw that it deletes all the _entrances_. My brain misread this as all teleporters for wfc.

CPU-Blanc (approved):
LGTM. All my outstanding comments are things that honestly are mostly tweaks that are fine either way. Haven't looked through the DLL Rust repo just yet but as that's just a straight single file drop-in I'm happy to look over that at a later time and get this in for now :+1:

Glloyd:
thanks for taking the lead on this @CPU-Blanc

PR #1190, opened 2023-09-10
feature: Config-ifies a bunch of Hunger-related values. Balance tweaks.

scrdest:
1. Most Hungers give more Suppression by default now. 
2. Made Vamp fangs actually usable for blood-drinking (aim for the head). 
3. Made regen much more powerful (up to a *full-on revive*) but more expensive and does NOT work on fire wounds (because both vamps and The Thing are weak to fire so it fits).

### DIFF
--- a/baystation12.dme
+++ b/baystation12.dme
@@ -3637,7 +3637,6 @@
 #include "interface\skin.dmf"
 #include "maps\_map_include.dm"
 #include "maps\_maps.dm"
-#include "maps\away\icarus\icarus_areas.dm"
 #include "maps\deepmaint\deepmaint.dm"
 #include "maps\deepmaint\deepmaint_rooms\template.dm"
 #include "maps\deepmaint\deepmaint_rooms\core\_core_rooms.dm"
--- a/baystation12.dme
+++ b/baystation12.dme
@@ -3637,6 +3637,7 @@
 #include "interface\skin.dmf"
 #include "maps\_map_include.dm"
 #include "maps\_maps.dm"
+#include "maps\away\icarus\icarus_areas.dm"
 #include "maps\deepmaint\deepmaint.dm"
 #include "maps\deepmaint\deepmaint_rooms\template.dm"
 #include "maps\deepmaint\deepmaint_rooms\core\_core_rooms.dm"
--- a/code/modules/urist/antagonist/revenant/archetypes/chimera.dm
+++ b/code/modules/urist/antagonist/revenant/archetypes/chimera.dm
@@ -97,8 +97,8 @@
 
 
 
-/* Simple heal. Does NOT regrow limbs. */
-/mob/proc/bsrevenant_minor_heal()
+/* Big heal. */
+/mob/proc/bsrevenant_incremental_regen()
 	set name = "Fleshmend"
 	set category = "Anomalous Powers"
 	set desc = "Unnatural regeneration. Revive on death, heal wounds/bloodloss otherwise."
@@ -177,7 +177,7 @@
 			var/tox_dmg = L.getToxLoss()
 			if(tox_dmg)
 				L.adjustToxLoss(-tox_dmg)
-				to_chat(L, SPAN_NOTICE("You purge absorbed toxins from your cells (though not your bloodstream)!"))
+				to_chat(L, SPAN_NOTICE("You purge necrotic tissue from your system."))
 
 				if(valid_revenant)
 					revenant.total_distortion += BSR_DISTORTION_GROWTH_OVER_SECONDS(tox_dmg, BSR_DEFAULT_DISTORTION_PER_TICK, BSR_DEFAULT_DECISECONDS_PER_TICK)
@@ -202,7 +202,7 @@
 	return !pending
 
 
-/datum/power/revenant/bs_power/minorheal
+/datum/power/revenant/bs_power/fleshmend
 	flavor_tags = list(
 		BSR_FLAVOR_CHIMERA,
 		BSR_FLAVOR_VAMPIRE,
@@ -213,7 +213,7 @@
 	activate_message = "<span class='notice'>You have an uncanny healing factor. You can heal minor damage and even stop bleeding - all at the low low price of just a little bit of damage to the local laws of physics.</span>"
 	name = "Fleshmend"
 	isVerb = TRUE
-	verbpath = /mob/proc/bsrevenant_minor_heal
+	verbpath = /mob/proc/bsrevenant_incremental_regen
 	distortion_threshold = 24000 // 20 mins
 
 
--- a/code/modules/urist/antagonist/revenant/archetypes/chimera.dm
+++ b/code/modules/urist/antagonist/revenant/archetypes/chimera.dm
@@ -124,7 +124,7 @@
 
 				// Short-circuit to a return; because reviving is more distortey
 				if(valid_revenant)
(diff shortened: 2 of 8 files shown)


code/modules/urist/antagonist/revenant/archetypes/vampiric.dm
```diff
@@ -372,7 +372,7 @@
 
 /* FANGS - gives mobs a new unarmed attack that requires head and ignores restraints. */
 /datum/unarmed_attack/bsrevenant_vampbite
-	attack_verb = list("bit", "sank his fangs in")
+	attack_verb = list("bit", "sank their fangs in")
```
Glloyd, line 375: you can use the /his macro here

baystation12.dme
```diff
@@ -3634,10 +3634,10 @@
 #include "interface\skin.dmf"
 #include "maps\_map_include.dm"
 #include "maps\_maps.dm"
-#include "maps\away\icarus\icarus_areas.dm"
 #include "maps\away\geminus_city\road_and_pavements.dm"
 #include "maps\away\geminus_city\signs.dm"
 #include "maps\away\geminus_city\streetobjects.dm"
+#include "maps\away\icarus\icarus_areas.dm"
```
Glloyd, line 3640: this shouldn't be part of this pr

separately, icarus_areas needs to be moved to icarus.dm but @CPU-Blanc may have already done that in an upcoming pr

CPU-Blanc: `icarus_areas` is already included in `icarus.dm`, without any PRs or anything like that.

Glloyd: definitely should be removed from the dme entirely in that case, if you want to just do that here @scrdest

scrdest: Oh, sure. IDK what's up with that, it's the DME acting weird again and it's not even that the IDE was open this time

code/modules/urist/antagonist/revenant/archetypes/vampiric.dm
scrdest: Ah, cool, wasn't sure when it resolves and how

Glloyd:
if you resolve the conflicts and the things mentioned above today, i can have it in for this weekend.

code/modules/urist/antagonist/revenant/archetypes/chimera.dm
```diff
@@ -101,51 +101,103 @@
...
+				if(valid_revenant)
+					revenant.total_distortion += BSR_DISTORTION_GROWTH_OVER_SECONDS(clone_dmg, BSR_DEFAULT_DISTORTION_PER_TICK, BSR_DEFAULT_DECISECONDS_PER_TICK)
+					return FALSE
+
+		// Tox
+		if(pending)
+			var/tox_dmg = L.getToxLoss()
+			if(tox_dmg)
+				L.adjustToxLoss(-tox_dmg)
+				to_chat(L, SPAN_NOTICE("You purge absorbed toxins from your cells (though not your bloodstream)!"))
```
Naxodile, line 180: For humans toxloss is just damage to organs (excluding the brain) nowadays.

code/modules/urist/antagonist/revenant/archetypes/chimera.dm
```diff
@@ -101,51 +101,103 @@
...
+		// First things first - if they are dead, revive
 		if(pending)
-			for(var/obj/item/organ/I in H.internal_organs)
-				if(I.damage >= 5)
-					I.damage = max(I.damage - 5, 0)
-					to_chat(H, SPAN_NOTICE("Your [I.name] itches as you knit its wounds shut."))
-					pending = FALSE
-					break
+			if(H.stat == DEAD)
+				H.adjustOxyLoss(-rand(15,20))
```
Naxodile, line 118: You probably want to bump this up since anyone dead's going to have 200 oxyloss and will get pretty significant brain damage by the time they're awake.

code/modules/urist/antagonist/revenant/archetypes/chimera.dm
```diff
@@ -101,51 +101,103 @@
...
+				continue
 
-			for(var/datum/wound/W in E.wounds)
-				if(W.bleeding())
-					W.bleed_timer = 0
-					E.status &= ~ORGAN_BLEEDING
-					to_chat(H, SPAN_NOTICE("You knit together severed veins, stemming the bleeding from your [E.name]."))
-					pending = FALSE
-					break
+			E.rejuvenate()
```
Naxodile, line 137: This is going to also full heal all organs inside each limb. bad_external_organs also includes any limb with internal damage (excluding brain damage) even if it lacks external damage.

Glloyd:
Nax the goddess of medical saving the day yet again. thank you.

code/modules/urist/antagonist/revenant/archetypes/chimera.dm
scrdest: Yeah, but I'm accounting for **non**humanoids, potentially.

code/modules/urist/antagonist/revenant/archetypes/chimera.dm
scrdest: This value is actually stolen from Bay's psionics revive code. Keep in mind, this power is meant to potentially be used multiple times in a row - obviously, for a higher total cost in realityfuckery - so brain damage can be healed by that too.

code/modules/urist/antagonist/revenant/archetypes/chimera.dm
Naxodile: Besides maybe the phoron pistol I'm not sure there's many instances of simple mobs or any other living mobs taking toxin damage. It's just going to be a kind of strange message for most circumstances.

code/modules/urist/antagonist/revenant/archetypes/chimera.dm
scrdest: That's kind of the point; I would do that by individual organs as prior except, err, that code didn't work properly fsr and would routinely block all other healing (which is why there's a prob onnit; stochastically it will eventually heal everything, just potentially on multiple tries)

code/modules/urist/antagonist/revenant/archetypes/vampiric.dm
scrdest: Done

baystation12.dme
scrdest: Should be sorted

code/modules/urist/antagonist/revenant/archetypes/chimera.dm
scrdest: Adjusted

code/modules/urist/antagonist/revenant/archetypes/chimera.dm
Naxodile: Since neither this or the blood restoration trigger any distortion increases burn damage is healed for free if you just cast this enough.

Naxodile: Never mind I missed the 300 seconds catch all increase. The cost being detached from the amount of damage healed should still be considered though.

scrdest: Nope, they do - that's what the fallthrough case at the very bottom does. Unless a heal-type has a specific custom cost, it gives it a generic cost.

Glloyd:
I'll move BSR to votable for next weekend so we can get more chances to test the new values.

PR #1195, opened 2023-09-12
Re-re-works lung damage

Naxodile:
Bends the lung damage curve and changes how it deals damage. If you aren't a doctor the tl;dr is that inaprovaline will save your life now. Normal pressure lung ruptures are mostly non-lethal and make oxygenation hover at a little below 85% causing minor brain damage. Info for doctors is that inaprovaline decreases the minimum amount of oxygen breathed to achieve internal oxygenation.


Glloyd:
i feel like you need to add one more re- at this point. Thanks for this 👍

PR #1197, opened 2023-09-12
Expands away map testing & enables template unit tests

CPU-Blanc:
Quite a few changes here, mainly to GitHub workflows/tests. End goal of this PR is to bring all away maps into line and pass unit tests, ensure new away maps are always tested, and enable testing for templates which were previously disabled due to OOM issues.

- Adds a new env variable to `run-test.sh`: `TEMPLATES`. If set to true, this enables templates to be loaded during unit testing by passing the `-T` flag to `dm.sh`
- Adds a new variable flag `-T` to `dm.sh`. When used, sets the `MAP_TEST_TEMPLATES` define to be used in the .dme.

This system was the cleanest way I could think to have template testing be toggled; defining `MAP_TEST_TEMPLATES` in the test map file itself does not work as it's defined too late for the compiler to then include the template loading call. This way, templates can be toggled independently of any other test variables/conditions if need be.

Includes various tweaks and additions to the GitHub testing workflow:
- All map checks now depend on DreamChecker passing. It makes no sense to spin up an instance and try to test the game when the compile has failed. In my experience this just results in a lot of red-herrings. Compile errors should always be tackled first before trying to test in a live environment. No point wasting all the time these map checks take to run.
- Awaymaps has been moved out of the map testing matrix. This is to allow it to run alongside (and depend on) a new test `check_away_testing.sh`. As like above, running tests on the away sites testing is pointless if a new away has been added to the game, but isn't actually being tested.
- New test `check_away_testing.sh` has been added. This has been added as a step in the `AwayMaps` job. This is a rather crude and simple script that scans the /away/ directory, and checks that each directory name has an associated #include in `away_sites_testing.dm`. If the include it expects is missing  and it's not added to the exclusions list, the test will fail; prompting the new away to be added so it is included in unit testing. This will also prevent the away_sites_testing test to spin up an instance and start running tests; as said above, this would be pointless. Downside of this is the naming format for aways is rather strict now, but 99% of aways all followed it (those that didn't have been renamed). That being `away/[away name]/[away name].dm`. It's a little strict currently, but my shell script regex-fu is lacking, but if you follow the format used everywhere else, it'll work fine.
- Templates GitHub test has been added. It uses some of the stuff already defined in `away_sites_testing.dm` (such as the blank map, lobby options etc), but does not load any aways, simply loading a single blank map. Doing this minimises memory usage and actually allows all templates to load in without issue. Without this, attempting to load the templates on any other map causes DreamDaemon to run out of memory and crash.

And finally:
- Away maps that were missing from `away_site_testing.dm` have been added, and have had changes applied so that they now pass.

There's quite a few big changes in here, including to the GitHub workflow, so any and all feedback is very welcome. There's probably other ways of tackling this issue, but this is what I could come up with.


CPU-Blanc:
Damn Bucket pos in past warning :<
I count that as a pass. Really need to cherrypick that new timer system in

maps/away/smugglers/smugglers_items.dm
```diff
@@ -0,0 +1,95 @@
+/obj/item/paper/smug_1
+	name = "suspicios note"
```
Glloyd, line 2:
```suggestion
	name = "suspicious note"
```

maps/nerva/nerva.dm
```diff
@@ -86,10 +86,10 @@
 	#include "../../code/datums/music_tracks/space_oddity.dm"*/
 
 	//keep your clothing hacks to yourself
-	#include "../torch/items/clothing/solgov-accessory.dm"
+	// #include "../torch/items/clothing/solgov-accessory.dm"	These don't seem to actually be used, but are causing issues with double-importing
```
Glloyd, line 89: should we just move all the torch clothing the the dme? seems weird to include some of this stuff in noctis and some in nerva.

test/check-away-testing.sh
```diff
@@ -0,0 +1,59 @@
+FAILED=0
+TOTAL=0
+SKIPPED=0
+
+EXCLUDE=("stations" "geminus_city")
```
Glloyd, line 5: i pre-empted you on this a bit, you can remove geminus from this as the folder no longer exists

```suggestion
EXCLUDE=("stations")
```

Glloyd (changes requested):

maps/away/icarus/icarus.dm
```diff
@@ -42,6 +42,7 @@
 	suffixes = list("icarus/icarus-1.dmm", "icarus/icarus-2.dmm")
 	spawn_cost = 2
 	generate_mining_by_z = 1
```
Glloyd, line 44:
```suggestion
	generate_mining_by_z = list(1,2)
```

Glloyd: not part of this PR but might as well do it here

maps/nerva/nerva.dm
CPU-Blanc: Personally, I'd like to move towards a more modular approach, where each away/map/template only includes what it uses, and the .dme itself only includes the current map - dme loads nerva, nerva loads all required dependencies, or dme loads glloydstation, glloydstation loads its dependencies. Seems strange to me that we're loading so many things that aren't actually used.
Only issue with this approach is currently, although they claim otherwise, double-including seems to be an issue. Not sure if it's _actually_ being double-included, or if dreamchecker isn't smart enough to discard the extra include, but that causes errors.

Right now I'm playing around with some preprocessor macros to see if I can come up with some elegant solution to resolve double includes, but, these are a pain to debug, so I wouldn't count on it just yet.

Honestly happy to go with whatever solution you feel is best.

maps/away/smugglers/smugglers_items.dm
CPU-Blanc: lmao didn't even notice that. Just copy and pasted it over

maps/nerva/nerva.dm
Glloyd: My instinct is to keep things in the dme for use in events and such, but you're probably right that there's a lot that can be cut out. Either way, since nerva always loads noctis, it's probably fine for now.

PR #1205, opened 2023-09-21
refactors how component_weapons work

Glloyd:
As in the title. Makes it so component weapons can shoot anything, not just /obj/item/projectiles, potentially allowing for AI shipweapons that shoot mobs. This also obsoletes the awful landmark system, and the new system is a lot more flexible, especially if we actually want to expand Blanc's PVP shipcombat framework. 

This PR also turns missiles into meteors because this began as me trying to get Bay's pdc to work with our combat system and spiraled into this mess.

Also adds some new weapons for hostile ships, they're not on any atm though.

Players shouldn't notice any difference, except that projectiles might be a little more random in where they hit the ship, and the bottom deck will probably get hit more.

removed the landmarks from Nerva and excised that old code
added the pdc stuff to the Nerva


Glloyd:
@CPU-Blanc I'd love if you could take a quick look at this if you get a chance, because i'm refactoring your own refactor of component weapons Fire()

code/game/gamemodes/meteor/meteors.dm
```diff
@@ -140,11 +140,17 @@ var/global/list/meteors_cataclysm = list(\
...
+	var/flicker_range //do we flicker lights? if so, in what range
+	var/shake_range //override Bay's calculation for shake range
 
 /obj/effect/meteor/proc/get_shield_damage()
-	return max(((max(hits, 2)) * (heavy + 1) * rand(30, 60)) / hitpwr , 0)
+	if(shield_damage_override)
+		return shield_damage_override
+
+	else
+		return max(((max(hits, 2)) * (heavy + 1) * rand(30, 60)) / hitpwr , 0)
```
scrdest, line 153: Can be simplified
```suggestion
	return max(((max(hits, 2)) * (heavy + 1) * rand(30, 60)) / hitpwr , 0)
```

code/game/gamemodes/meteor/meteors.dm
```diff
@@ -225,12 +231,23 @@ var/global/list/meteors_cataclysm = list(\
...
+		for(var/obj/machinery/light/L in range(flicker_range, src))
+			L.flicker(rand(5,15))
+
+	if(shake_range)
+		for(var/mob/M in range(shake_range, src))
+			if(!M.stat && M.client && !istype(M, /mob/living/silicon/ai))\
+				shake_camera(M, 3, 1)
+
+	else
+		for(var/mob/M in GLOB.player_list) //this is an insane way of doing this bay
```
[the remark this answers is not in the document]
scrdest: _It's not that bad._ 

This list only contains mobs with attached clients; even if our pop doubled, it would be at most 40 iterations, done _once_, over a homogeneous list. Hell, the game does much more every single tick of the health system.

You couldn't even do it by block(), since _all_ mobs on the z-level get _some_ screenshake, and even if you made dist > 20 not do any, you'd have to check a 40x40x1 block of turfs - 160 iterations minimum.

The only real optimization here would be to have a global cache of players **per z-level**, which would remove everything L246 needs to deal with anyway.

code/modules/urist/modules/shipbattles/overmapships/components/component_weapons.dm
```diff
@@ -55,83 +59,126 @@
...
+		P.launch(target_turf) //projectiles have their own special proc
+
+	else if(istype(projectile, /obj/effect/meteor))
+		var/obj/effect/meteor/M = projectile
+		M.dest = target_turf
+		spawn(0)
+			walk_towards(M, M.dest, 3) //meteors do their own thing too
+
+	else
+		projectile.throw_at(target_turf) //anything else just uses the default throw proc. this potentially allows for ship weapons that do things like throw mobs. clown cannon anyone?
```
scrdest, line 83: Seems to me there's a general-purpose `launch_anything(var/turf/start_turf, var/turf/target_turf, var/projectile_type)` proc underneath the ship bits here that could be used for other stuff and the shipweapons could just feed it their projectile type and endpoints.

code/modules/urist/modules/shipbattles/shipweapons/shipweapons.dm
```diff
@@ -273,13 +273,41 @@
...
-		var/obj/effect/overmap/visitable/ship/combat/T = target
-		var/obj/effect/urist/projectile_landmark/ship/P = pick(T.landmarks)
-		P.Fire(projectile_type)
+	if(projectile_type)
+		if(istype(target, /obj/effect/overmap/visitable/ship/combat))
+			HandlePvpFire()
+
+		else
+			var/obj/effect/urist/projectile_landmark/target/P = pick(GLOB.target_projectile_landmarks) //i'll come back to this to get rid of landmarks for ai ships
+			P.Fire(projectile_type)
```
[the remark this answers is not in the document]
scrdest: Good, embrace the hatred! ⚡ 👐 ⚡

code/modules/urist/modules/shipbattles/shipweapons/shipweapons.dm
```diff
@@ -273,13 +273,41 @@
...
+	else if(istype(projectile, /obj/effect/meteor))
+		var/obj/effect/meteor/M = projectile
+		M.dest = target_turf
+		spawn(0)
+			walk_towards(M, M.dest, 3) //meteors do their own thing too
+
 	else
-		var/obj/effect/urist/projectile_landmark/target/P = pick(GLOB.target_projectile_landmarks)
-		P.Fire(projectile_type)
+		projectile.throw_at(target_turf) //anything else just uses the default throw proc.
```
scrdest, line 310: Ah, case in point - this logic is duplicated. Could extract this to a shared proc.

code/modules/urist/modules/shipbattles/ship_overmaps.dm
```diff
@@ -17,6 +17,8 @@
 	var/can_escape = TRUE
 	var/dam_announced = 0
 	var/pvp_cooldown = 0
+	var/list/target_dirs //what directions can we be shot from. this is to mimic the effect of the old landmark system that mostly prevented hits on the SM.  if this is empty it will use GLOB.cardinal
+	var/list/target_zs //what zs can be hit in ship combat. this is mainly here to prevent the top deck getting hit on nerva.
```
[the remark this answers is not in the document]
Glloyd: You're right that that might be a better way to handle this, but that's probably best left for a future PR that expands on PVP, because that method would take a lot more work to function with how awaymap zs are loaded. This is fine for now because it mimics the existing landmark layout, where there's no landmarks at the rear anyways because of z-level explosions,

code/game/gamemodes/meteor/meteors.dm
```diff
@@ -225,12 +231,23 @@ var/global/list/meteors_cataclysm = list(\
...
+			var/turf/T = get_turf(M)
+			if(!T || T.z != src.z)
```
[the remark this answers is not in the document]
Glloyd: For some reason Bay doesn't use that operator much, even though Lummox has said it's more performant. Since I'm moving this code around I might as well change it though

code/game/gamemodes/meteor/meteors.dm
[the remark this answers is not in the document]
Glloyd: This whole file is full of some baffling baycode decisions

code/modules/urist/modules/shipbattles/projectile_landmarks.dm
```diff
@@ -47,11 +47,13 @@
 	fire_type = 2
 
 /obj/effect/urist/projectile_landmark/target/Fire(projectile_type)
```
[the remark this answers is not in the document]
Glloyd: they're on nerva-6.dmm, but are pretty much unused in practice

code/modules/urist/modules/shipbattles/projectile_landmarks.dm
```diff
@@ -47,11 +47,13 @@
...
-	var/obj/item/projectile/P = new projectile_type
+	if(istype(projectile_type, /obj/item/projectile))
```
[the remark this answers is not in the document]
Glloyd: good question. I'll fix that, although this whole system needs its own rework at some point.

code/modules/urist/modules/shipbattles/shipweapons/shipweapons.dm
```diff
@@ -273,13 +273,41 @@
...
+		if(istype(target, /obj/effect/overmap/visitable/ship/combat))
+			HandlePvpFire()
+
+		else
+			var/obj/effect/urist/projectile_landmark/target/P = pick(GLOB.target_projectile_landmarks) //i'll come back to this to get rid of landmarks for ai ships
+			P.Fire(projectile_type)
+
+/obj/machinery/shipweapons/proc/HandlePvpFire() //come back to this to add handling for burst fire
+	var/obj/effect/overmap/visitable/ship/combat/target_ship = target
+	var/target_x = rand(target_ship.target_x_bounds[1],target_ship.target_x_bounds[2])
```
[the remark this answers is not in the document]
Glloyd: Blanc reviews her old code better than i did 2 or 3 years ago

code/modules/urist/modules/shipbattles/shipweapons/shipweapons.dm
Glloyd: good point, i'll turn it into a helper proc

code/modules/urist/modules/shipbattles/overmapships/components/component_weapons.dm
```diff
@@ -5,48 +5,52 @@
...
+	var/projectile_type
 	var/obj/machinery/shipweapons/weapon_type = /obj/machinery/shipweapons/beam/lightlaser
 
 /datum/shipcomponents/weapons/DoActivate()
 	if(ready && !broken)
 		Fire()
 
-/datum/shipcomponents/weapons/proc/Fire(obj/effect/urist/projectile_landmark/ship/map_landmark) //now infinitely less hacky and broken
-	if(broken) //check one more time, just in case. doing it this way can stop burst fire mid burst
+/datum/shipcomponents/weapons/proc/Fire(var/turf/start_turf, var/turf/target_turf, var/target_edge, var/target_z) //now infinitely less hacky and broken
```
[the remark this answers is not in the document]
Glloyd: this is intentional, yeah. both for potential future content, and also the potential for ship weapons that target specific areas. I'm thinking a mob cannon that yeets mobs at the bridge.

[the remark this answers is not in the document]
CPU-Blanc: Ahh that makes sense. Monkey cannon at the bridge window, yes please.

code/modules/urist/modules/shipbattles/projectile_landmarks.dm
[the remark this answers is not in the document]
CPU-Blanc: I know you're trying to move away from the landmarks, but as these aren't actually present in the ship templates, no projectiles are actually getting yeeted there at all, which could make firing on the ship while an away team is there more interesting

code/modules/urist/modules/shipbattles/shipweapons/shipweapons.dm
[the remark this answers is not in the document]
CPU-Blanc: LMAO OOPS. Okay, my bad. Hey, at least it shows self-growth maybe :sweat_smile: ?

code/modules/urist/modules/shipbattles/shipweapons/shipweapons.dm
Glloyd: done

code/modules/urist/_helpers.dm
```diff
@@ -384,3 +384,21 @@
 		if(light_amt < max_light)
 			return 1
 	return 0
+
+//for throwing things from one turf to another
+
+/proc/launch_atom(var/atom/movable/projectile, var/turf/start_turf, var/turf/target_turf)
+	if(ispath(projectile))
+		new projectile(start_turf)
```
CPU-Blanc, line 392: I'm not 100%, but does this actually shadow `projectile` automatically with the new spawned atom? I would've thought not?

code/modules/urist/modules/shipbattles/projectile_landmarks.dm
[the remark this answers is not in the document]
Glloyd: I was never sure how I felt about it, especially in the context of player-controlled pirates. Might be cool, but I'd only want it to kick in after boarding is opened. I might mess with it before this gets merged.

code/modules/urist/_helpers.dm
Glloyd: already fixed locally, i'm just testing things with the pdc right now. good catch though, i had a minute where i was very confused why the Nerva wasn't getting blasted

code/modules/urist/modules/shipbattles/projectile_landmarks.dm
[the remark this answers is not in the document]
Glloyd: projectile landmark code is now fully yeeted

code/modules/urist/modules/shipbattles/overmapships/components/component_weapons.dm
```diff
@@ -5,133 +5,160 @@
...
+					shot_number = 0
+					ready = TRUE
+			else
+				spawn(salvo_delay)
+					Fire(start_turf, null, target_edge, target_z) //doing this so we fire from the same spot at different targets.
+																//this gets us sustained fire on parts of the ship instead of a straight line of rockets, or the random scatter we had in the distant past.
 
+/datum/shipcomponents/weapons/proc/locate_target(var/target_z)
+	var/target_x = rand(mastership.target_ship.target_x_bounds[1],mastership.target_ship.target_x_bounds[2])
+	var/target_y = rand(mastership.target_ship.target_y_bounds[1],mastership.target_ship.target_y_bounds[2])
```
scrdest, line 59: This is currently uniform random; if you want Gaussian noise to simulate aiming at the center with random error instead, I can get you a sampler implementation.

code/game/gamemodes/meteor/meteors.dm
```diff
@@ -208,7 +213,7 @@ var/global/list/meteors_cataclysm = list(\
 
 /obj/effect/meteor/use_tool(obj/item/tool, mob/user, list/click_params)
 	// Pickaxe - Delete meteor
-	if (istype(tool, /obj/item/pickaxe))
+	if (istype(tool, /obj/item/pickaxe) && !ismissile)
```
scrdest, line 216: Why would a pickaxe be a missile? (aside from being dorfy af)

code/modules/urist/_helpers.dm
```diff
@@ -384,3 +384,21 @@
 		if(light_amt < max_light)
 			return 1
 	return 0
+
+//for throwing things from one turf to another
+
+/proc/launch_atom(var/projectile_type, var/turf/start_turf, var/turf/target_turf)
```
scrdest, line 390: If you do it like this:
```suggestion
/proc/launch_atom(var/turf/start_turf, var/turf/target_turf, var/projectile_type =<something> )
```
the API might be a bit nicer - if there's a sane default, you only have to specify endpoints.

code/modules/urist/modules/shipbattles/ship_overmaps.dm
```diff
@@ -273,4 +268,7 @@
 
 /obj/effect/overmap/visitable/ship/combat/proc/assign_target_zs() //this is a proc so it can be overridden by non-nerva ships. restricting zs from being hit for awaymaps is a little more complicated because they don't have set z-levels, but can be done here if you get creative.
 	if(!target_zs)
-		target_zs = map_z
\ No newline at end of file
+		target_zs = map_z
+
+/obj/effect/overmap/visitable/ship/combat/proc/pve_mapfire(var/projectile_type)
```
CPU-Blanc, line 273: Seems odd to me. I _think_ what you intended to do was make `pve_mapfire` a proc of `/mob/living/simple_animal/hostile/overmapship`? That way each template can handle how projectiles come in etc? Doesn't really make sense for different player ships to need this function

maps/nerva/nerva_overmap.dm
```diff
@@ -65,3 +65,9 @@
 	shuttle = "Hadrian"
 	fore_dir = EAST
 	vessel_mass = 1000
+
+/obj/effect/overmap/visitable/ship/combat/nerva/pve_mapfire(projectile_type)
+	if(ispath(projectile_type))
+		var/turf/start_turf = spaceDebrisStartLoc(pick(GLOB.cardinal), src.z)
+		var/turf/target_turf = locate(100, 100, src.z) //set up values for enemy ships
+		launch_atom(projectile_type, start_turf, target_turf)
```
CPU-Blanc: If things remain how they are, won't the Nerva be shooting at itself here?

CPU-Blanc: Actually it'll be shooting at the actual overmap :L

CPU-Blanc: Welp. No idea why github was showing me old files.. GG. I see you already fixed this

code/modules/urist/_helpers.dm
CPU-Blanc: If this is supposed to a generic helper proc, I'm not even sure what a sane default _would_ be. Because this proc doesn't have to be limited to ship weapons; any sort of projectile could use it

code/game/gamemodes/meteor/meteors.dm
Glloyd: ismissile here refers to the meteors, a var Bay added and did almost nothing with

code/modules/urist/modules/shipbattles/ship_overmaps.dm
Glloyd: This is intentional so each combat ship can define where they want hostile ships spawned in. This way you could add a blank z level to a new combat ship awaymap as the last map loaded, and then use max(map_z) to fire projectiles there. Boarding code would need a slight rework to make ships spawn there properly, but that's for a future PR.

code/modules/urist/modules/shipbattles/overmapships/components/component_weapons.dm
Glloyd: I think it's alright for the purposes of this. having random start locations for projectiles now makes things feel even more random anyways. Might be something worth looking into in the future though.

Glloyd:
thanks for the help from both of you, this system is a lot better than the old one and we finally got rid of the old landmark hell.

code/game/gamemodes/meteor/meteors.dm
scrdest: Okay, sure, but my point is... why would the tool be a pickaxe AND a meteor? 

If the first istype() passes, I don't see why would the added check ever NOT be True, because it's a pickaxe, not a missile/meteor.

If it didn't pass, the whole check gets short-circuited - there's no point checking further ANDed conditions once the first has failed (similarly, ORing stuff stops at the first True check because the whole thing will be True regardless if anything else fails the check).

So, the `!ismissile` check will either **not run at all**, or **will always pass**. As such, it seems entirely superfluous.

CPU-Blanc: Nono, `ismissile` is on the `src` _meteor_ (`!src.ismissile`), NOT the tool. The check makes sense

code/modules/urist/modules/shipbattles/ship_overmaps.dm
CPU-Blanc: I think I'm getting confused between the current state and the WIP rework I was playing around with that allowed fr multiple ship templates spawning in over multiple Zs. Been a long time since I looked at this code XD
Think there's definitely a followup PR somewhere in all this

code/game/gamemodes/meteor/meteors.dm
scrdest: Oh derp. That's why I hate the implicit-src convention =_=

Glloyd: yeah, i still use src in a lot of places for my own readability, this might've been one of those places where it would help

code/modules/urist/modules/shipbattles/ship_overmaps.dm
Glloyd: 👀 that sounds very interesting

PR #1207, opened 2023-09-21
Leathercrafting and cloth weaving

CaraS212:
Adds more recipes to leather and animal furs
adds a loom and cotton plants.
folds up some of the icons for clothing objects
adds new cape colors from a greyscale version Vak made (<3), comes in historic dye colors that fit departments

misc: 
adds cotton seeds to the forest away
fixes some capitalization in heads/uristclothes.dm


code/modules/urist/items/clothstuff.dm
```diff
@@ -94,3 +94,60 @@
 				else
 					to_chat(user, "<span class='warning'>\The [src] is used up, but there are more wounds to treat on \the [affecting.name].</span>")
 			use(used)
+
+/obj/item/loom
+	name = "table loom"
+	desc = "A loom small enought to only take up a table instead of the whole floor."
```
CPU-Blanc, line 100:
```suggestion
	desc = "A loom small enough to only take up a table instead of the whole floor."
```

code/modules/urist/items/clothstuff.dm
```diff
@@ -94,3 +94,60 @@
...
+	display_name = "cotton plant"
+	chems = list(/datum/reagent/cottonfiber = list(6,1))
+
+/datum/seed/cotton/New()
+	..()
+	set_trait(TRAIT_MATURATION,1)
+	set_trait(TRAIT_PRODUCTION,1)
+	set_trait(TRAIT_YIELD,7)
+	set_trait(TRAIT_POTENCY,10)
+//	set_trait(TRAIT_PRODUCT_ICON,"cotton")
```
CaraS212, line 123: commented out for the time being because for unknown reasons, this is not recognized as an icon even though its in the DMI. and the potato icon would be an ok substitute but it isnt showing up as white.

code/modules/urist/items/clothstuff.dm
```diff
@@ -94,3 +94,60 @@
...
+/obj/item/reagent_containers/food/snacks/grown/attackby(obj/item/W, mob/user)
+
+	if(seed)
+		if(seed.chems)
+			if(isLoom(W))
+				if(!isnull(seed.chems[/datum/reagent/cottonfiber]))
+					user.visible_message(SPAN_NOTICE("\The [user] weaves the cotton into cloth \the [src]."))
+					new /obj/item/stack/material/cloth(user.loc)
+					qdel(src)
+					return
```
CPU-Blanc, line 152: Few issues with this bit:
- This is `isLoom` proc & macro are really unnecessary as it's only a single item. The `is_` procs are usually used as multiple different types can be used in place of the exact tools.
- Unnecessary `if` check; you could group these in a single OR, or better use the conditional null operator `?.`
- This chunk also needs to be moved to `code/modules/hydroponics/grown.dm` Ln:169 - What you're doing here is overloading the `attackby` proc for `/obj/item/reagent_containers/food/snacks/grown/`. `..()` is a call to the _parent_ type, meaning everything else in the previous proc define for this obj type will be replaced, not added to (it'll call `attackby()` on `/obj/item/reagent_containers/food/snacks`, not the existing proc on `/obj/item/reagent_containers/food/snacks/grown`)
```suggestion
/obj/item/reagent_containers/food/snacks/grown/attackby(obj/item/W, mob/user)
	if(seed?.chems)
		if(istype(W, /obj/item/loom))
			if(!isnull(seed.chems[/datum/reagent/cottonfiber]))
				user.visible_message(SPAN_NOTICE("\The [user] weaves the cotton into cloth \the [src]."))
				new /obj/item/stack/material/cloth(user.loc)
				qdel(src)
				return
```

CPU-Blanc (changes requested):
Couple of issues that'll need to be addressed

code/modules/urist/modules/materials/recipes_skin.dm
```diff
@@ -1,3 +1,63 @@
+/material/skin/goat/generate_recipes(reinforce_material)
+	. = ..()
+	if(reinforce_material)	//recipies below don't support composite materials
+		return
```
CPU-Blanc, line 4: `reinforce_material` check is already carried out in the parent proc, just pass it through
```suggestion
	. = ..(reinforce_material)
```

code/modules/urist/modules/materials/recipes_skin.dm
```diff
@@ -1,3 +1,63 @@
...
+
+	. += new/datum/stack_recipe/goatpelt(src)
+	. += new/datum/stack_recipe/sheeppelt(src)
+
+/material/skin/fur/generate_recipes(reinforce_material)
+	. = ..()
+	if(reinforce_material)	//recipies below don't support composite materials
+		return
```
CPU-Blanc, line 12: As above

code/modules/urist/modules/materials/recipes_skin.dm
```diff
@@ -1,3 +1,63 @@
...
+	. += new/datum/stack_recipe/brownbearpelt(src)
+	. += new/datum/stack_recipe/bisonpelt(src)
+	. += new/datum/stack_recipe/bisonhat(src)
+	. += new/datum/stack_recipe/lionpelt(src)
+	. += new/datum/stack_recipe/foxpelt(src)
+
+/material/skin/fur/grey/generate_recipes(reinforce_material)
+	. = ..()
+	if(reinforce_material)	//recipies below don't support composite materials
+		return
```
CPU-Blanc, line 23: And again

code/modules/urist/modules/materials/recipes_skin.dm
```diff
@@ -1,3 +1,63 @@
...
+
+	. += new/datum/stack_recipe/graywolfpelt(src)
+
+/material/skin/fur/white/generate_recipes(reinforce_material)
+	. = ..()
+	if(reinforce_material)	//recipies below don't support composite materials
+		return
```
CPU-Blanc, line 30: aaaand here

code/modules/urist/modules/materials/recipes_skin.dm
```diff
@@ -1,3 +1,63 @@
...
+
+	. += new/datum/stack_recipe/whitewolfpelt(src)
+	. += new/datum/stack_recipe/whitebearpelt(src)
+	. += new/datum/stack_recipe/whitefoxpelt(src)
+
+/material/skin/fur/black/generate_recipes(reinforce_material)
+	. = ..()
+	if(reinforce_material)	//recipies below don't support composite materials
+		return
```
CPU-Blanc, line 39: once more

code/modules/urist/modules/materials/recipes_skin.dm
```diff
@@ -1,3 +1,63 @@
...
+
+	. += new/datum/stack_recipe/pantherpelt(src)
+	. += new/datum/stack_recipe/blackbearpelt(src)
+
+/material/skin/lizard/generate_recipes(reinforce_material)
+	. = ..()
+	if(reinforce_material)	//recipies below don't support composite materials
+		return
```
CPU-Blanc, line 47: Almost there

code/modules/urist/modules/materials/recipes_skin.dm
```diff
@@ -1,3 +1,63 @@
...
+
+	. += new/datum/stack_recipe/gatorpelt(src)
+	. += new/datum/stack_recipe/lizardpelt(src)
+
+/material/skin/lizard/generate_recipes(reinforce_material)
+	. = ..()
+	if(reinforce_material)	//recipies below don't support composite materials
+		return
```
CPU-Blanc, line 55: and finally here

icons/obj/hydroponics_products.dmi
CaraS212, line 1: i know we are supposed to keep bayfile edits to a minimum but using a custom file for plants and seeds would be a big hack job (at least for me)

code/modules/urist/items/clothstuff.dm
```diff
@@ -94,3 +94,46 @@
...
+	set_trait(TRAIT_WATER_CONSUMPTION, 6)
+
+/obj/item/seeds/cotton
+	seed_type = "cotton"
+
+/obj/item/reagent_containers/food/snacks/grown/attackby(obj/item/W, mob/user)
+	if(seed?.chems)
+		if(istype(W, /obj/item/loom))
+			if(!isnull(seed.chems[/datum/reagent/cottonfiber]))
+				user.visible_message(SPAN_NOTICE("\The [user] weaves the cotton into cloth \the [src]."))
```
CPU-Blanc, line 135: This grammatically doesn't make sense. If from a cotton plant, you'd have "[user] weaves the cotton into cloth cotton", but "[user] weaves the cotton into cloth potato" if xenoflora do magic.
Would 
```suggestion
				user.visible_message(SPAN_NOTICE("\The [user] weaves \the [src] into cotton cloth"))
```
be a better fit?

CPU-Blanc (approved):
lgtm

CaraS212:
stupid fucking checks failing and passing at a coin toss

code/modules/urist/items/clothstuff.dm
```diff
@@ -94,3 +94,46 @@
...
+	desc = "A loom small enough to only take up a table instead of the whole floor."
+	icon = 'icons/urist/items/wood.dmi'
+	icon_state = "loom"
+
+/datum/reagent/cottonfiber
+	name = "cotton fiber"
+	description = "A mass of cotton fibers."
+	taste_description = "cotton"
+	reagent_state = SOLID
+	color = "#ffffff"
```
scrdest, line 109: It's usually not a good idea to use 100% saturation colors, doesn't leave much colorspace to work with for blending.
```suggestion
	color = "#f0f0f0"
```

code/modules/urist/items/clothstuff.dm
```diff
@@ -94,3 +94,46 @@
...
+	chems = list(/datum/reagent/cottonfiber = list(6,1))
+
+/datum/seed/cotton/New()
+	..()
+	set_trait(TRAIT_MATURATION,10)
+	set_trait(TRAIT_PRODUCTION,1)
+	set_trait(TRAIT_YIELD,8)
+	set_trait(TRAIT_POTENCY,10)
+	set_trait(TRAIT_PRODUCT_ICON,"cotton")
+	set_trait(TRAIT_PRODUCT_COLOUR,"#ffffff")
```
scrdest, line 124: And here
```suggestion
	set_trait(TRAIT_PRODUCT_COLOUR,"#f0f0f0")
```

code/modules/urist/items/clothstuff.dm
```diff
@@ -94,3 +94,46 @@
...
+	set_trait(TRAIT_POTENCY,10)
+	set_trait(TRAIT_PRODUCT_ICON,"cotton")
+	set_trait(TRAIT_PRODUCT_COLOUR,"#f0f0f0")
+	set_trait(TRAIT_PLANT_ICON,"bush2")
+	set_trait(TRAIT_WATER_CONSUMPTION, 6)
+
+/obj/item/seeds/cotton
+	seed_type = "cotton"
+
+/obj/item/reagent_containers/food/snacks/grown/attackby(obj/item/W, mob/user)
```
Glloyd, line 131: i have to agree with Blanc on this, this seems like a very hacky and difficult to maintain way of doing this. It also just seems a bit unintuitive. Why are we calling the attackby() on the cotton itself and not the loom. Seems more intuitive to hit the loom with the cotton and not the other way around.

code/modules/urist/items/clothstuff.dm
```diff
@@ -94,3 +94,46 @@
 				else
 					to_chat(user, "<span class='warning'>\The [src] is used up, but there are more wounds to treat on \the [affecting.name].</span>")
 			use(used)
+
+/obj/item/loom
```
Glloyd, line 98: for that matter, is there a particular reason why this is an item and not a structure? It seems like it makes a lot more sense as a structure, particularly with the sprite as it is.

code/modules/urist/modules/jungle/jungle_animals.dm
```diff
@@ -460,20 +464,23 @@
...
 	icon_gib = "cougar_dead"
+	skin_material = MATERIAL_SKIN_FUR
 
 /mob/living/simple_animal/hostile/huntable/bear/grey
 	desc = "A big scary bear, probably best to stay away"
 	icon_state = "greybear"
 	icon_living = "greybear"
 	icon_dead = "greybear_dead"
 	icon_gib = "greybear_dead"
+	skin_material = MATERIAL_SKIN_FUR_BLACK
```
Glloyd, line 475: does this make more sense as `MATERIAL_SKIN_FUR_GREY`?

Glloyd (changes requested):
A couple little things, the big one that definitely needs to be changed before merging is that attackby()

code/modules/urist/items/clothstuff.dm
CaraS212: because i dont know how to make cotton used on the loom turn into cloth

CaraS212: also bc the loom is just not a good sprite so at least as a table loom item instead of a floor loom, it can be moved around and be out of the way

code/modules/urist/modules/jungle/jungle_animals.dm
CaraS212: player convenience, we dont have any grey pelts so at least with black fur they can make a bear pelt from a grey bear.

code/modules/urist/items/clothstuff.dm
CaraS212: fixed with your code
ty <3

Glloyd:
👍 squashing and merging

PR #1208, opened 2023-09-22
Noctis burn chamber OSHA compliance

CaraS212:
Upgrades the interior walls of the Noctis burn chamber to plasteel and moves some of the thrusters so that the point of failure vents into space and not the halls, giving the crew a slightly larger margin of error.
The three thrusters were moved to the sides with 1 added for symmetry.
burn chart


Glloyd:
This is probably the best way of balancing this. failures aren't roundending, but you still need to get the right mix.

PR #1214, opened 2023-09-24
Senior Scientist gets a Nanotrasen account card

ggermanicus:
Currently has anywhere from 11,000T to 44,000T each round. The lockbox is in the senior's locker, making it pretty hard for anyone but an antag or them to grab it, but it's also an antag objective too, so someone will eventually. Also adds an ATM to their office for ease of use.

Also fixes a bug where the account card lockboxes weren't updating their icons when you locked them again, and fixes them saying you locked the box when you unlocked the box.


Naxodile (changes requested):
It's still strange to me we have this setup on the lockbox instead of a LateInitialize on the cards themselves. Was there some timing issue with it?

PR #1215, opened 2023-09-24
The Little Things - September Edition

Naxodile:
Collection of verified issues in #bug-reports
Fixes long range relays being unanchored
Bluespace RPEDs function remotely again
Chem master can be anchored/unanchored normally
HoP dress is allowed for HoP on Glloydstation and SO on Nerva
Cryo engineering voidsuit's suit onmob is fixed
Holdout pistol's in hand and onmob are fixed
Makeshift mask now has an onmob, pulled down state, and slightly revised colors
Blue/red forensic suits now have omobs
The non-prototype smg from research now has the correct magazine and name
Screenshake and scoped accuracy now properly reset on unzooming


scrdest (approved):
LGTM

PR #1216, opened 2023-09-24
Fixes Noctis Thrusters

Vakothu:
For some reason Bay decided Thrusters on ships should silently break if they're not facing the same way this one specific variable says.
See `code\modules\overmap\ships\engines\gas_thruster.dm` line 96


Vakothu:
Note that prior to our recent Baymerge the ONLY thing this variable did was change which direction stars fly when moving.

PR #1217, opened 2023-09-24
Allows Teshari to play submap roles

Naxodile:
The only place this define is used strangely enough


Glloyd:
Huh, didn't even know this was a define. 👍

PR #1220, opened 2023-09-25
restores relays to get long distance relays working again

Glloyd:
Brings relays back, minus being constructable. The reason Bay removed them has never been an issue here, but this prevents it from ever being an issue.

### DIFF
--- a/code/game/machinery/telecomms/machine_interactions.dm
+++ b/code/game/machinery/telecomms/machine_interactions.dm
@@ -381,7 +381,6 @@
 	else
 		src.listening_levels = GLOB.using_map.contact_levels
 		return 1
-	return 0
 
 // RELAY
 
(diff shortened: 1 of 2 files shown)


Glloyd:
very funny that the old relay code doesn't pass dreamchecker

code/game/machinery/telecomms/machine_interactions.dm
```diff
@@ -363,3 +363,49 @@
...
+	var/turf/position = get_turf(src)
+
+	// Toggle on/off getting signals from the station or the current Z level
+	if(src.listening_levels == GLOB.using_map.contact_levels) // equals the station
+		src.listening_levels = GetConnectedZlevels(position.z)
+		return 1
+	else
+		src.listening_levels = GLOB.using_map.contact_levels
+		return 1
+	return 0
```
Glloyd, line 384: either this proc is very stupid or i am very tired, but it's hurting my brain.

PR #1221, opened 2023-09-25
fixes shuttles going random places, fixes cargo tug not taking damage, gives forensic borg a camera

ggermanicus:
also forensic borgs can pick up folders and camera film to refill said camera. they could already pick up photos idk why they didn't have a camera

### DIFF
--- a/code/modules/overmap/overmap_shuttle.dm
+++ b/code/modules/overmap/overmap_shuttle.dm
@@ -46,13 +46,6 @@
 	var/distance_mod = get_dist(waypoint_sector(current_location),waypoint_sector(next_location))
 	return move_time * (1 + distance_mod)
 
-/datum/shuttle/autodock/overmap/process_launch()
-	if(prob(10))
-		var/places = get_possible_destinations()
-		var/place = pick(places)
-		set_destination(places[place])
-	..()
-
 /datum/shuttle/autodock/overmap/proc/set_destination(obj/effect/shuttle_landmark/A)
 	if(A != current_location)
 		next_location = A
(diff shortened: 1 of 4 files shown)


code/modules/overmap/overmap_shuttle.dm
```diff
@@ -46,13 +46,6 @@
 	var/distance_mod = get_dist(waypoint_sector(current_location),waypoint_sector(next_location))
 	return move_time * (1 + distance_mod)
 
-/datum/shuttle/autodock/overmap/process_launch()
-	if(prob(10))
```
Glloyd, line 50: why is this even a thing...

code/modules/vehicles/vehicle.dm
```diff
@@ -80,45 +80,56 @@
...
-	else if(hasvar(W,"force") && hasvar(W,"damtype"))
-		user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
-		switch(W.damtype)
-			if("fire")
-				adjust_health(-W.force * fire_dam_coeff)
-			if("brute")
-				adjust_health(-W.force * brute_dam_coeff)
-		..()
-		healthcheck()
+			to_chat(user, SPAN_NOTICE("Unable to repair while [src] is off.</span>"))
```
Glloyd, line 108: code test is failing here, you don't need the </span>

```suggestion
			to_chat(user, SPAN_NOTICE("Unable to repair while [src] is off."))
```

Glloyd (changes requested):

ggermanicus:
should be good now

PR #1224, opened 2023-09-26
removes bay teargas

CaraS212:
Apparently teargas was extra strong due to both bays gas and ours in tgitems.dm being applied at the same time in 1 case. at least thats what i understand nyx to have said. This comments out bay's spicy gas and uses our mild recipe.

### DIFF
--- a/code/game/objects/items/weapons/grenades/chem_grenade.dm
+++ b/code/game/objects/items/weapons/grenades/chem_grenade.dm
@@ -269,26 +269,25 @@
 	icon_state = initial(icon_state) +"_locked"
 
 
-///obj/item/grenade/chem_grenade/teargas
-//	name = "tear gas grenade"
-//	desc = "Concentrated Capsaicin. Contents under pressure. Use with caution."
-//	stage = 2
-//	path = 1
+/obj/item/grenade/chem_grenade/teargas
+	name = "tear gas grenade"
+	desc = "Concentrated Capsaicin. Contents under pressure. Use with caution."
+	stage = 2
+	path = 1
 
 
-///obj/item/grenade/chem_grenade/teargas/Initialize()
-//	. = ..()
-//	var/obj/item/reagent_containers/glass/beaker/large/B1 = new(src)
-//	var/obj/item/reagent_containers/glass/beaker/large/B2 = new(src)
-//	B1.reagents.add_reagent(/datum/reagent/phosphorus, 40)
-//	B1.reagents.add_reagent(/datum/reagent/potassium, 40)
-//	B1.reagents.add_reagent(/datum/reagent/capsaicin/condensed, 40)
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/grenades/chem_grenade.dm
scrdest, line 1: Probably would have been better to change the recipe here and comment out the pure-Urist version.

For one, this means we don't have to worry as much about maintaining the object itself. For another, we should be using Initialize(), not New(), where possible.

scrdest (changes requested):

Glloyd:
how intense is the bay version compared to our old one when it's not being double applied? like scr says, it makes more sense to just get rid of our old one, but i don't know if it's even necessary to change the bay values

CaraS212:
glloyd: bays is 2x as strong as ours

code/game/objects/items/weapons/grenades/chem_grenade.dm
CaraS212: copy, ill change it such that the bay file uses an amount that is halved so it is equal to our old stuff. should i comment out ours in the tgitems file?

CaraS212:
see description of most recent commit

code/game/objects/items/weapons/grenades/chem_grenade.dm
```diff
@@ -278,13 +278,12 @@
...
-	var/obj/item/reagent_containers/glass/beaker/large/B1 = new(src)
-	var/obj/item/reagent_containers/glass/beaker/large/B2 = new(src)
-	B1.reagents.add_reagent(/datum/reagent/phosphorus, 40)
-	B1.reagents.add_reagent(/datum/reagent/potassium, 40)
-	B1.reagents.add_reagent(/datum/reagent/capsaicin/condensed, 40)
-	B2.reagents.add_reagent(/datum/reagent/sugar, 40)
-	B2.reagents.add_reagent(/datum/reagent/capsaicin/condensed, 80)
+	var/obj/item/reagent_containers/glass/beaker/B1 = new(src)
+	var/obj/item/reagent_containers/glass/beaker/B2 = new(src)
+	B1.reagents.add_reagent(/datum/reagent/capsaicin, 30)
```
Glloyd, line 283: what is the difference mechanically between capsaicin and capsaicin/condensed?

CaraS212: condensed is what is supposed to be in a grenade and teargas, normal is what is in food and just gives taste and pain if you eat too much of it.

CaraS212:
i double checked and pepperspray does use condensed capsaicin, i believe prior to the merge the grenades did as well, knowing from taking them apart. 
also its been a week on something minor, i have no idea what i was doing code wise beyond making teargas not be able to gas out an entire deck.

PR #1225, opened 2023-09-26
FBP buffs: Eject MMI verb, eating food for charge

ggermanicus:
- All Full Body Prosthetics should be able to eject their MMI from their bodies. Think pez dispensers. You'll need someone else to put you back into yourself.
- FBPs (that can eat) can eat nutriment-containing food for charge. They can't drink for charge (including drinking nutriment), and they won't benefit from animal protein (eating meat). Congrats on your Skrellian diet. Rechargers are still your main source of charge but you won't collapse in the bar.
- For reference, you get around 200 charge for most (vegetarian) meals. Protein bars give 190.
- CPR messages without a mouth have been clarified so it is more obvious that you are still being useful even if you have no mouth.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -74,14 +74,14 @@
 		if(I_HELP)
 			if(MUTATION_FERAL in M.mutations)
 				return 0
-			
+
 			# ifdef INCLUDE_URIST_CODE
 			if(urist_status_flags & STATUS_UNDEAD)
 				return 0
 			# endif
-			
+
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -108,10 +108,10 @@
 						resuscitate()
 
 				if(!H.check_has_mouth())
-					to_chat(H, SPAN_WARNING("You don't have a mouth, you cannot do mouth-to-mouth resuscitation!"))
+					to_chat(H, SPAN_WARNING("You've done chest compressions, but you don't have a mouth to do mouth-to-mouth resuscitation!"))
```
Glloyd, line 111: thank you, this was confusing me as a dionae

Glloyd:
👍 for "pez dispensers"

PR #1226, opened 2023-09-26
feature: Empty Z-level LODs + 3d AI

scrdest:
Bay mob AI, as of today, does not handle Z-levels at all. 

Even if the AI could pathfind across Z-levels (which it cannot), on our pop there's bound to be a lot of Z-levels without players.

The SSpresence subsystem currently throttles ALL processing on empty Z-levels for AI (strategic & tactical) and Mob processing. 

This had been an understandable attempt at preserving computing power for stuff that is not used, but it leads to Considerable Silliness like mobs who are supposed to be corpses dying only when discovered, or floating in mid-air and falling when players enter the level.

This had been addressed in two parts in this PR:

- The throttling had received a new config option to 'soften' it stochastically; instead of hard-skipping a subsystem tick, you can specify a percentage chance of skipping, applied only if the subject's Z-level has no player mobs. 
- The AI pathfinding and movement got ported to use 3d-aware versions of adjacency, distance and movement procs (lots of credit to @CPU-Blanc for these from a while back).

The soft-throttle rate defaults to 100 (so, same as Baycode, not processing anything) but config'd to 75% because empirically this makes the AI feel responsive but still saves us doing 3 calls out of 4 on average.

The AI changes are kind of a first pass; the mobs can and will move across Z-levels if ordered, but I didn't mess with the targetting logic so they won't _chase_ players up Zs yet because they cannot infer the target changed Z-levels.

Furthermore, I used the plain distance procs; this means the AI doesn't mind switching Z-levels very much, so it will happily jump off ledges instead of taking the stairs.

Simplemobs don't take fall damage, so it's not a major issue, and most AIs that use Astar for pathing (required for multi-z) represent either dumb monsters or commando types who could conceivably do dynamic entry like that for coolness factor.


scrdest:
🖕 🪣 🪣

Glloyd:
I don't know how I feel about the soft-throttle rate. I like it in theory, but in practice it might make more sense to apply only to station_zs or select multi-z aways after a shuttle has landed there. Which is a bit more involved, but otherwise you have a whole lot of processing being done on mobs that players probably won't ever see in a round. Especially with how few awaymissions have been happening in the last couple weeks.

scrdest:
> I don't know how I feel about the soft-throttle rate. I like it in theory, but in practice it might make more sense to apply only to station_zs or select multi-z aways after a shuttle has landed there. Which is a bit more involved, but otherwise you have a whole lot of processing being done on mobs that players probably won't ever see in a round. Especially with how few awaymissions have been happening in the last couple weeks.

Not _that_ hard to do. Before checking SSpresence, check a cache for that Z-level. If value for the z is missing/false, check SSpresence. 

If the cached value is positive AND the SSpresence value is negative, roll the dice as in this PR. If both are negative, skip unconditionally. If SSpresence is positive, cache the positive and tick unconditionally.

This way, any 'visited' Z becomes 'activated' forever (1) and will use the leaky logic if nobody's on it, unvisited Zs are always skipped, and Zs with mobs work as usual.

I can probably knock this system out in like half an hour if you like the concept.

Glloyd:
concept is good, i just need to get the time to look through this

scrdest:
@Glloyd smol pester-ping

Glloyd:
i might need @CPU-Blanc to take a quick look through this for me if possible, i haven't had the time that i'd hoped.

CPU-Blanc:
Un-stalling. I'll be around next week to review

code/controllers/subsystems/presence.dm
```diff
@@ -42,7 +45,24 @@ SUBSYSTEM_DEF(presence)
...
 	return levels["[level]"] || 0
+	#else
+	// Urist edit - caching levels that HAD pop at some point
+	var/raw_pop = levels["[level]"]
+	var/pop = raw_pop
+	if(raw_pop)
+		activated_levels["[level]"] = raw_pop
+	else
+		pop = 0
+	return pop
```
CPU-Blanc, line 58: This whole `pop`/`raw_pop` assignment seems really odd to me an a little clunky. Can't we just return `pop` or 0 if null?
```suggestion
	var/pop = levels["[level]"]
	if(pop)
		activated_levels["[level]"] = pop
	return pop || 0
```

code/controllers/subsystems/ai.dm
```diff
@@ -23,14 +24,28 @@ SUBSYSTEM_DEF(ai)
...
 		if (!config.run_empty_levels && !SSpresence.population(get_z(ai.holder)))
 			continue
+		#else
+		// Allow 'leaky' processing of empty, activated z-levels
+		if (!config.run_empty_levels)
+			var/holder_z = get_z(ai.holder)
+			var/zlevel_has_pop = SSpresence.population(holder_z)
+			if(!zlevel_has_pop)
+				var/zlevel_had_pop = SSpresence.population_from_cache(holder_z)
+				if(!zlevel_had_pop || (zlevel_had_pop && throttle_on_empty))
```
CPU-Blanc, line 46: Extra `zlevel_has_pop` is redundant as we've already checked the state in the previous OR branch which is lazily evaluated.

Would it also be possible to factor in populated connected z-levels in here too? Something kinda like...

```suggestion
				var/zlevel_had_pop = SSpresence.population_from_cache(holder_z)
				var/connected_has_pop = FALSE
				for(var/connected_z in GetConnectedZlevels(holder_z))
					if(SSpresence.population(connected_z))
						connected_has_pop = TRUE
						break
				if((!zlevel_had_pop && !connected_has_pop) || throttle_on_empty)
```

code/controllers/subsystems/aifast.dm
```diff
@@ -23,14 +24,28 @@ SUBSYSTEM_DEF(aifast)
...
 		if (!config.run_empty_levels && !SSpresence.population(get_z(ai.holder)))
 			continue
+		#else
+		// Allow 'leaky' processing of empty, activated z-levels
+		if (!config.run_empty_levels)
+			var/holder_z = get_z(ai.holder)
+			var/zlevel_has_pop = SSpresence.population(holder_z)
+			if(!zlevel_has_pop)
+				var/zlevel_had_pop = SSpresence.population_from_cache(holder_z)
+				if(!zlevel_had_pop || (zlevel_had_pop && throttle_on_empty))
```
CPU-Blanc, line 46: As above

code/controllers/subsystems/mobs.dm
```diff
@@ -25,12 +26,26 @@ SUBSYSTEM_DEF(mobs)
...
 		if (!config.run_empty_levels && !SSpresence.population(get_z(mob)))
 			continue
+		#else
+		// Allow 'leaky' processing of empty, activated z-levels
+		if (!config.run_empty_levels)
+			var/holder_z = get_z(mob)
+			var/zlevel_has_pop = SSpresence.population(holder_z)
+			if(!zlevel_has_pop)
+				var/zlevel_had_pop = SSpresence.population_from_cache(holder_z)
+				if(!zlevel_had_pop || (zlevel_had_pop && throttle_on_empty))
```
CPU-Blanc, line 46: And here

CPU-Blanc (commented):
Couple of minor nitpicks but in general looks fine to me.
Also a minor possible improvement if we want to utilise connected z-levels to allow an unpopulated level to process at a throttled rate

scrdest:
I will get back to this; between work and GOAI breakthroughs, I've simply not had the time.

code/controllers/subsystems/presence.dm
scrdest: Yeah, you're right, idk why I hadn't done it this way in the first place. 🧠 💨 I guess.

code/controllers/subsystems/ai.dm
scrdest: Best I can recall, I've done it this way just to be explicit about the logic. 

For connected Zs, I'd like to test it out with the leaky processing on one Z first. My concern is that this might heavily de-optimize the current LODs in the worst case where players are heavily scattered across levels.

CPU-Blanc: Could we not edit `SSpresence::population_from_cache()` to also cache over connected Z levels then? Ie for all connected z-levels, store in the cache that they have previously been populated?

Not a big deal, but it'd be nice for instances like the Nerva if no-one has been on deck 1, but to still allow AI on that level to process at a reduced rate

scrdest:
AI ROUGE

CPU-Blanc (approved):
Forgor to approve this.

I'd still like to see a possible follow-up with caching/LODing connected Z-levels at a later date

PR #1240, opened 2023-09-30
Adds 'Generate Worths' workflow dispatch & Fixes caching

CPU-Blanc:
As requested by @scrdest; a manually triggered workflow that exports all material and object worths to JSON files for debugging. During testing, I noticed caching actually was (and hasn't) been working at all, a fix for that has been included while I was working on it.

Changes:

- Added "Generate Worths" GitHub workflow. This is a manually triggered action, which you can point to a branch to run on. Dropdown menu selects what map includes to export worths from (ie, items/materials included within those map files).  "None" will include no extra maps and export for only base items, whereas "all" exports for Nerva, Away Maps, Example + Base. Once complete, the resulting JSON files are uploaded as an artefact which can be downloaded as a zip by the end-user.
- Fixed caching over all workflows. During testing I noticed caching was generating an error, and after looking into it, I don't think our cache has been properly running for a long time. This is due to changes(?) in shell formatting with paths. Namely `$HOME` is depreciated in favour of `~`, and env var notation must be used. I'd suggest if/when this is merged, the current caches are deleted and allowed to rebuild; their sizes are a little sus.
- Moves template testing to its own file. I kinda cheated before, but really it should be its own file. Helps with this new test as it allows a spawn of a blank map without loading any aways or templates if the flag isn't set.
- Added a new flag to `dm.sh` - `-W`. This is used for enabling the worths export subsystem, much like how unit testing is handled. Arguments for this flag are a `|` separated list of map names to include.
- Added `generate-worths.sh` script. It takes one env var `MAP_PATH`, which is either `all`, `none` or much like away_sites_testing, a map name.
- Added the `Worths test` subsystem. This is the meat of all this and works on the same principle as the unit test system. By default it is not even included unless `DEBUG_GENERATE_WORTHS` is defined (which is set by using `-W` in `dm.sh`). This disables some unnecessary subsystems, starts the world, calculates every possible combination of materials (base and reinforced), and generates a list of resulting details (listed in the schema below). It'll then spawn in all defined traders, and then work through all objects in the game (excluding those in the exclusions), getting their base worth using the fallback  method, as well as getting the min & max worth trying to sell to each spawned trader. Deletes the world when done.

Currently, there are several runtimes generated by this, however these  (with the exception of the duplicate culture runtime - sue me) are all caused by faulty `New`, `Init` or `Destroy` procs of the items spawned. Amusingly, this has actually  proved to be a good way of identifying runtimes in objects we don't frequently use.

As said, this generates 2 JSON files, one for objects, one for materials. The schema for each are..

Materials:
```
{
    <material_type>: {
        "id": <String>,
        "units_per_sheet": <Int>,
        "base_recipes": [
            {
                "product": <type_path>,
                "mat_cost": <Int>,
                "prod_amount": <Int>
            }
        ],
        "reinforced_recipes": {
            <material id (String)>: [
                {
                    "product": <type_path>,
                    "mat_cost": <Int>,
                    "prod_amount": <Int>
                }
            ]
        }
    }
}
```

Objects:
```
{
    <object_path>: {
        "trade_value_low": <Int> (Nullable),
        "trade_value_high": <Int> (Nullable),
        "worth": <Int>,
        "composition": {
            <material id (String)>: <Int>
        }
    }
}
```

As always, let me know if there's any issues!


CPU-Blanc:
I ran the workflow over on my repo, so you can see logs/end JSON files

https://github.com/CPU-Blanc/UristMcStation/actions/runs/6359852863

code/worths_test/~helpers.dm
```diff
@@ -0,0 +1,25 @@
...
+					"prod_amount" = stack_recipie.res_amount
+				)))
+		else if(istype(listed_recipie, /datum/stack_recipe))
+			var/datum/stack_recipe/stack_recipie = listed_recipie
+
+			recipes.Add(list(list(
+				"product" = stack_recipie.result_type,
+				"mat_cost" = stack_recipie.req_amount,
+				"prod_amount" = stack_recipie.res_amount
+			)))
```
scrdest, line 23:
```suggestion
	for(var/listed_recipe in src.get_recipes(reinf_mat))
		if(istype(listed_recipe, /datum/stack_recipe_list))
			var/datum/stack_recipe_list/srl = listed_recipie
			for(var/datum/stack_recipe/stack_recipe in srl.recipes)
				recipes.Add(list(list(
					"product" = stack_recipe.result_type,
					"mat_cost" = stack_recipe.req_amount,
					"prod_amount" = stack_recipe.res_amount
				)))
		else if(istype(listed_recipe, /datum/stack_recipe))
			var/datum/stack_recipe/stack_recipe = listed_recipe

			recipes.Add(list(list(
				"product" = stack_recipe.result_type,
				"mat_cost" = stack_recipe.req_amount,
				"prod_amount" = stack_recipe.res_amount
			)))

```
Someone was coding on an empty stomach 😄

code/worths_test/~worths_test_subsystems.dm
```diff
@@ -0,0 +1,282 @@
...
+		var/material/mat = src.material_queue[length(src.material_queue)]
+		LIST_DEC(src.material_queue)
+
+		log_worths_test("**** [uppertext(mat.name)] ****")
+
+		var/list/base_recipes = mat.dump_recipes()
+		var/list/reinforced_recipes = list()
+
+		var/mat_name = lowertext(mat.name)
+		log_worths_test("-- [mat_name]: [length(base_recipes)] recipie\s")
```
scrdest, line 136:
```suggestion
		log_worths_test("-- [mat_name]: [length(base_recipes)] recipe\s")
```

code/worths_test/~worths_test_subsystems.dm
```diff
@@ -0,0 +1,282 @@
...
+		for(var/material/reinf_mat in SSmaterials.materials)
+			if(istype(reinf_mat, /material/placeholder))
+				continue
+
+			var/list/recipes = mat.dump_recipes(reinf_mat)
+
+			if(!length(recipes))
+				continue
+
+			log_worths_test("-- [mat_name] (Reinforced - [lowertext(reinf_mat.name)]): [length(recipes)] recipie\s")
```
scrdest, line 147:
```suggestion
			log_worths_test("-- [mat_name] (Reinforced - [lowertext(reinf_mat.name)]): [length(recipes)] recipe\s")
```

code/worths_test/~worths_test_subsystems.dm
```diff
@@ -0,0 +1,282 @@
...
+
+			reinforced_recipes[lowertext(reinf_mat.name)] = recipes
+
+		src.processed_materials[mat.type] = list(
+			"id" = mat_name,
+			"base recipes" = base_recipes,
+			"reinforced recipes" = reinforced_recipes
```
scrdest, line 154:
```suggestion
			"base_recipes" = base_recipes,
			"reinforced_recipes" = reinforced_recipes
```
Easier to parse this way

scrdest (changes requested):
You've done this miles better than I would! 🗿 
Just a few typos plus space-to-underscore conversion.

code/worths_test/~helpers.dm
CPU-Blanc: Can't blame hunger for this, that was straight up dyslexia :laughing:

code/worths_test/~worths_test_subsystems.dm
CPU-Blanc: No idea why I didn't do that here, I did for the materials.  I blame coding at 3am

scrdest:
Also, I noticed that it seems like all the low/high trader values are Null. Is this expected?

CPU-Blanc:
> Also, I noticed that it seems like all the low/high trader values are Null. Is this expected?

It's not _all_. Just a _lot_. Seems like there's a lot of items that are  just not sold to traders at all, but I've definitely seen several with entries there

scrdest:
> > Also, I noticed that it seems like all the low/high trader values are Null. Is this expected?
> 
> It's not _all_. Just a _lot_. Seems like there's a lot of items that are just not sold to traders at all, but I've definitely seen several with entries there

Yeah, you're right - I can see the internal organs have value, for instance. Still, this workflow should make it easier to find ones without values!

scrdest (approved):
LGTM - many, many thanks for doing this!

PR #1244, opened 2023-10-01
Significantly shifts spillover damage chances

Naxodile:
Shifts the chance of an attack causing internal organ damage from being the sum of organ relative_size + damage threshold to strictly damage threshold. The choice to halve the threshold is arbitrary and may need to be reduced further. 

In terms of gameplay effects this will be a massive nerf to fast firing and (relatively) low damage weapons being able to kill quickly.
Lasers in general will also be more likely to kill via blood loss or infection than organ disintegration. Higher caliber bullets retain a decent chance of penetration per hit.

### DIFF
--- a/code/modules/organs/external/_external_damage.dm
+++ b/code/modules/organs/external/_external_damage.dm
@@ -139,7 +139,7 @@
 	if(!damage_amt)
 		return FALSE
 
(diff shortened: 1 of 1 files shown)


Glloyd:
i'll merge this for testing tonight

PR #1248, opened 2023-10-01
balance: BSR balance pass.

scrdest:
Generally, meant to throttle the sheer insanity of BSR as-is. 

- Added thresholds for (some) Distortion FX.
- Hostile NPC spawns less likely.
- Tuned down Distortion effect rate significantly; now defaults to 5% chance of effect per tile in the worst case and the slope to that point is much longer. 
- Fixed goldeater Hunger (was using SStrade unnecessarily)
- Fixed vampbites (was too low priority)
- Fixed some strings.
- Configified some stuff.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -456,6 +456,15 @@
 	//       from the POV of the total Distortion spread. It's not exactly equivalent - because the smaller the tickrate, the
 	//       more responsive the Distortion system will be, but it will also drain more resources to run.
 
+	// Probability of triggering a BSR Distortion for a tile depends on the current level of 'irradiation'.
+	// This represents the turf-level value that corresponds to 50% of max_chance_per_turf of triggering.
+	var/static/bluespace_revenant_distortion_maxchance_halfway_point = 500
+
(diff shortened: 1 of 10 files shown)


scrdest:
🖕 🪣 🪣

code/controllers/configuration.dm
```diff
@@ -961,6 +976,12 @@
 			if ("bluespace_revenant_tickrate")
 				bluespace_revenant_tickrate = max(1, text2num(value))
 
+			if ("bluespace_revenant_distortion_max_chance_per_turf")
+				bluespace_revenant_tickrate = max(1, text2num(value))
+
+			if("bluespace_revenant_distortion_maxchance_halfway_point")
+				bluespace_revenant_distortion_maxchance_halfway_point = max(0, text2num(value))
```
Glloyd, line 983: should probably add all these values to config.txt even if we're keeping the default values

scrdest: Done - I didn't add them initially to keep the config clean, but I can see how it helps. 

There's a deliberate difference between the default code value and our config value for one option (Bloodburner) BTW. The power is disabled from normal play anyhow.

PR #1250, opened 2023-10-02
Mining generates ores better, EVA shutter button, bar windoor, antags restricted by species

ggermanicus:
- Big drill effectiveness up 5000%! (number made up)
- Antags now have a list of valid species you can be. You cannot be a not-valid species and an antag. You cannot appearance-edit to a not-valid species as an antag. RIP Dionantag.
- EVA shutters are no longer robotics shutters. Bar window has an access restriction.

### DIFF
--- a/code/modules/mob/living/carbon/human/appearance.dm
+++ b/code/modules/mob/living/carbon/human/appearance.dm
@@ -13,10 +13,9 @@
 		return
 
 	var/datum/antagonist/antag = get_antag_data(mind?.special_role)
-	if(!(new_species in antag.valid_species))
+	if(!(new_species in antag?.valid_species))
 		return
-	else
-		set_species(new_species)
+	set_species(new_species)
 	if (antag && antag.required_language)
 		add_language(antag.required_language)
 		set_default_language(all_languages[antag.required_language])
--- a/code/modules/mob/living/carbon/human/appearance.dm
+++ b/code/modules/mob/living/carbon/human/appearance.dm
@@ -12,8 +12,11 @@
 	if(!(new_species in all_species))
 		return
 
-	set_species(new_species)
-	var/datum/antagonist/antag = mind && player_is_antag(mind)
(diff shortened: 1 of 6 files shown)


scrdest (approved):

code/modules/mob/living/carbon/human/appearance.dm
```diff
@@ -12,8 +12,11 @@
 	if(!(new_species in all_species))
 		return
 
-	set_species(new_species)
-	var/datum/antagonist/antag = mind && player_is_antag(mind)
+	var/datum/antagonist/antag = get_antag_data(mind?.special_role)
+	if(!(new_species in antag.valid_species))
+		return
+	else
+		set_species(new_species)
```
CPU-Blanc, line 19: We're null checking mind prior, but if mind is null here, `antag` will also be null. We're now not checking for a null in `antag` where it's possible it will be. Also redundant else as prior branch returned
```suggestion
	var/datum/antagonist/antag = get_antag_data(mind?.special_role)
	if(!(new_species in antag?.valid_species))
		return
	set_species(new_species)
```

code/modules/mob/living/carbon/human/appearance.dm
```diff
@@ -172,6 +177,8 @@
 			continue
 		if(length(blacklist) && (current_species_name in blacklist))
 			continue
+		if(antag && !(current_species_name in antag.valid_species))
```
CPU-Blanc, line 180: We don't need the double access here. Conditional null operator works a treat
```suggestion
		if(!(current_species_name in antag?.valid_species))
```

CPU-Blanc (changes requested):
Missing null check and a minor nitpick.
Personally I would've preferred the antag restrictions to be its own PR as the rest are bundled bugfixes, but that's just me.

CPU-Blanc (approved):

PR #1251, opened 2023-10-02
streamlines mining asteroids

ggermanicus:
- NT mining station and its asteroid cluster are now one big map (two dmms)
- Away missions that were just other mining locations with nothing interesting should no longer spawn.
- Fixes trading station ATMs.
- Removes the mining stuff from randomzlevels which are never ever used anyways and yet still somehow caused conflicts


maps/away/stations/nerva_stations.dm
```diff
@@ -37,10 +38,12 @@
 	name = "NanoTrasen trading station"
 	id = "awaysite_nanotrasentrading"
 	description = "A NanoTrasen trading station, dock here to trade with its many merchants."
```
Glloyd, line 40: should probably update this to indicate that you can mine here

Glloyd (changes requested):
this all seems fine, and is probably a better way of doing things than having a shuttle between two separate awaymaps. if you update the language in the descriptions we can get this merged.

Glloyd (approved):

PR #1252, opened 2023-10-02
Adds traitor item that removes Clumsy mutation

ggermanicus:
It's in the Devices & Tools section for 16 telecrystals.

### DIFF
--- a/code/modules/urist/items/declumsifier.dm
+++ b/code/modules/urist/items/declumsifier.dm
@@ -5,7 +5,7 @@
 	usr.mutations.Remove(MUTATION_CLUMSY)
 	. = ..()
 
-/obj/item/device/uplink_service/attack_self(mob/user)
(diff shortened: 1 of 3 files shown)


code/modules/urist/items/declumsifier.dm
```diff
@@ -0,0 +1,19 @@
+/obj/item/device/uplink_service/declumsifier
+	service_label = "Declumsifier"
+
+/obj/item/device/uplink_service/declumsifier/enable(mob/user = usr)
+	usr.mutations.Remove(MUTATION_CLUMSY)
+	. = ..()
+
+/obj/item/device/uplink_service/attack_self(mob/user)
```
scrdest, line 8:
```suggestion
/obj/item/device/uplink_service/declumsifier/attack_self(mob/user)
```

scrdest (changes requested):
Think you missed a path component

ggermanicus:
i did indeed, thanks

PR #1255, opened 2023-10-03
Fixes out of breath whispering

Naxodile:
whisper_say would call say and cause an infinite loop as of baymerge (hopefully?)


scrdest (approved):
Good spot. Some lovely mutual recursion going on here.

PR #1257, opened 2023-10-03
Re-adds Frag Grenades back to Uplink, SM Grenades and Increases Emag Uses.

Yantarii:
- Readds Fragmentation Grenades and Bundle back to the Emag, which we definitely had before the merge.
- Enables Supermatter Grenades (Singular) to be bought, I don't believe they do a crazy amount of damage, but if they do we can adjust.
- Changes Emag to have uses between 18-24 before frying. -  https://github.com/UristMcStation/UristMcStation/pull/1128

### DIFF
--- a/code/game/objects/items/weapons/cards_ids.dm
+++ b/code/game/objects/items/weapons/cards_ids.dm
@@ -155,7 +155,11 @@
 	icon_state = "emag"
 	item_state = "card-id"
 	origin_tech = list(TECH_MAGNET = 2, TECH_ESOTERIC = 2)
-	var/uses = 10
+	var/uses = 18
+
+/obj/item/card/emag/New()
+	uses = rand(18, 24)
+	..()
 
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/cards_ids.dm
```diff
@@ -155,7 +155,11 @@
 	icon_state = "emag"
 	item_state = "card-id"
 	origin_tech = list(TECH_MAGNET = 2, TECH_ESOTERIC = 2)
-	var/uses = 10
+	var/uses = 18
+
+/obj/item/card/emag/New()
```
scrdest, line 160: Might need to use Initialize() instead

Glloyd (changes requested):
should definitely use Initialize() on that

Yantarii:
Fixed it up now, should be good.

PR #1261, opened 2023-10-04
cane/coin fixes

CaraS212:
The traitor caneshotgun is no longer safe, can be fired
removed material name from loot coins


CaraS212:
I HATE BUCKET POS I HATE BUCKET POS

PR #1268, opened 2023-10-04
fix: Nerfs the eyerape of the Noir mutation

scrdest:
The current Noir color matrix is not only absurdly high contrast, it doesn't even filter out the UI from being affected - which means getting it breaks the UI entirely as it becomes a uniform mass of color.

This slaps this down to a warm pseudo-greyscale, which should make this less of a ggnore.

This code is... not great, but since mutations are normally nearly unobtainable (aside from adminbus or some xenoarch artifacts), I didn't want to spend too much time on fixing this.


Glloyd:
genetics is such a mess

PR #1271, opened 2023-10-05
fix: NT Station NPC fixes

scrdest:
- Stops NPC traders from trading beyond the grave. Dead or unconscious NPCs don't trade anymore.
- Fixed cops going psycho at *everyone*. Instead, they get mad at whoever is going psycho, by name. This has two side-effects:
  1. A clever traitor can disguise themselves as an innocent crewmember before going on a spree. This turns the station hostile _to the framed crewmember_
  2. Unfortunately, cops stop shouting a warning. This is due to Bay AI code check skipping a taunt if the source has an attacker.

This is currently a rough, quick fix. I'm not very happy with the police logic, but a more global solution is a lot more work and I'd rather fix the immediate issues first.

### DIFF
--- a/code/modules/urist/modules/newtrading/NPC/npc.dm
+++ b/code/modules/urist/modules/newtrading/NPC/npc.dm
@@ -113,7 +113,7 @@
 		p.ai_holder.react_to_attack(attacker)
 
 /mob/living/simple_animal/passive/npc/proc/can_use(mob/M)
-	if(M.stat || M.restrained() || M.lying || !istype(M, /mob/living) || get_dist(M, src) > 1)
+	if(!istype(M, /mob/living) || M.stat || M.restrained() || M.lying || get_dist(M, src) > 1)
 		return 0
 	return 1
 
--- a/code/modules/urist/modules/newtrading/NPC/npc.dm
+++ b/code/modules/urist/modules/newtrading/NPC/npc.dm
@@ -104,8 +104,13 @@
 
 /datum/ai_holder/simple_animal/passive/nowandering/react_to_attack(atom/movable/attacker)
 	. = ..()
+
+	if(!attacker)
+		return
+
 	for(var/mob/living/simple_animal/hostile/scom/civ/combat/police/p in orange(7, holder))
-		p.ai_holder.hostile = TRUE
+		p.ai_holder.add_attacker(attacker)
+		p.ai_holder.react_to_attack(attacker)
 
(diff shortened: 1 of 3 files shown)


code/modules/urist/modules/newtrading/NPC/npc_interact.dm
```diff
@@ -7,23 +7,29 @@
...
 /mob/living/simple_animal/passive/npc/attack_hand(mob/living/user)
-	if(user && istype(user) && can_use(user))
+	if(src.stat != CONSCIOUS)
+		to_chat(user, "[src] doesn't look like he's in any shape to trade right now...")
+		interacting_mob = null
+		return
 
-		if(interacting_mob && !can_use(interacting_mob))
-			interacting_mob = null
+	if(!(user && istype(user) && can_use(user)))
```
CPU-Blanc, line 15: A lot of this is unnecessary. `can_use` already performs an `istype` check for user, which would also return null if `user` is null. If we switch the check order in `can_use` ever so slightly, it'll become null safe and allow us to skip a lot of the doubled up checks
```suggestion
	if(!can_use(user))
```

code/modules/urist/modules/newtrading/NPC/npc.dm
```diff
@@ -104,8 +104,13 @@
...
+	if(!attacker)
+		return
+
 	for(var/mob/living/simple_animal/hostile/scom/civ/combat/police/p in orange(7, holder))
-		p.ai_holder.hostile = TRUE
+		p.ai_holder.add_attacker(attacker)
+		p.ai_holder.react_to_attack(attacker)
 
 /mob/living/simple_animal/passive/npc/proc/can_use(mob/M)
 	if(M.stat || M.restrained() || M.lying || !istype(M, /mob/living) || get_dist(M, src) > 1)
```
CPU-Blanc, line 116: Not part of this PR, but if we adjust this slightly, we can make it null safe
```suggestion
	if(!istype(M, /mob/living) || M.stat || M.restrained() || M.lying || get_dist(M, src) > 1)
```

CPU-Blanc (commented):
Just a little redundancy that's bugging me

scrdest:
@CPU-Blanc the bits you commented on got modified fairly incidentally (just went on a de-nesting spree), but if I'm improving quality, I guess I might as well do these...

CPU-Blanc (approved):
Yeah, but might as well while the files are being changed.
LGTM!

PR #1272, opened 2023-10-05
Mapping prep

Naxodile:
Collection of fixes/features needed for upcoming maps.

- Map previews in general should be accurate.
- Safes will now store items on their starting turf
- Tables and racks are now able to be mapped with color. Fixes rack/dark
- d_green light changes
- Cash bundles can be mapped without starting rotation
- Blood stains can be mapped


Glloyd:
this is great. Walls are another pet peeve of mine when it comes to this, but there's not much that can be done about that one.

PR #1277, opened 2023-10-05
fix: Zombie fixes cherrypick

### LINKED ISSUE #1274: Bay-Zombies - Prying Door & Damaging Synthetics + Mechs.
### Describe the problem you're trying to solve

- Bay Zombies cannot pry doors, and cannot get past any non-general access doors, meaning they have to stay in hallways and other areas.

- Bay Zombies cannot damage Mechs at all, meaning they have no ability to counter Mechs & Borgs in a round.

- Bay Zombies cannot do anything towards Synthetic Crew, i.e the consume verb and such. Meaning they have to keep consistently attacking them until they are fully crit and dead.

### Describe the solution you'd like to see

Zombies should be able to pry open doors, allowing them access into areas with the disarm or grab intent. Additionally being able to destroy doors or smash them open would be a nice feature.

Zombies should be able to damage Mechs slowly in some way, so that 2-3 zombies working together can do some decent damage against them. Additonally being able to attack borgs would be very nice.

Having the consume verb work somewhat on downed synthetic crew, or for example just FBP would be nice, not fully damaging/killing them but damaging internal synth organs or something similar.

scrdest:
Fixes zombies not being able to pry doors or fight mechs.
Fixes #1274 

Mostly an early-bird port of https://github.com/Baystation12/Baystation12/pull/34161

During the discovery for this, I've noticed that for some bizarre reason Baycode _qdel()s_ inventory from zombified mobs.

Considering this is _bloody stupid_, I've moved some bits that need to be dropped for zombie attacks to drop inventory instead, and disabled dropping the rest (e.g. IDs) altogether (because why, even???)

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -32,6 +32,8 @@
 				SPAN_DANGER("\The [user] smashes through \the [src]!"),
 				SPAN_DANGER("You smash through \the [src]!")
 			)
+			if (MUTATION_FERAL in user.mutations)
+				qdel(src)
 		else
 			user.visible_message(
 				SPAN_DANGER("\The [user] [attack_verb] \the [src]!"),
--- a/code/game/machinery/doors/door.dm
+++ b/code/game/machinery/doors/door.dm
@@ -198,7 +198,26 @@
 			do_animate("deny")
 	return
 
+/obj/machinery/door/airlock/attack_generic(mob/user)
+	if (MUTATION_FERAL in user.mutations)
+		user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN*2)
+		playsound(loc, damage_hitsound, 50, 1)
+		attack_animation(user)
+
+		if((MACHINE_IS_BROKEN(src)||!arePowerSystemsOn(src)) && density)
+			visible_message(SPAN_DANGER("\The [user] manages to pry \the [src] open!"))
+			open(1)
+		else
+			visible_message(SPAN_DANGER("\The [user] smashes into \the [src]!"))
+			damage_health(10)
+		return
+	..()
+
 /obj/machinery/door/attack_hand(mob/user)
+	if (MUTATION_FERAL in user.mutations)
+		attack_generic(user, 15)
+		return
+
 	..()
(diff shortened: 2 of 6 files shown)


code/game/machinery/doors/door.dm
```diff
@@ -198,7 +198,26 @@
...
+		else
+			visible_message(SPAN_DANGER("\The [user] smashes into \the [src]!"))
+			damage_health(10)
+		return
+	..()
+
 /obj/machinery/door/attack_hand(mob/user)
+	if (MUTATION_FERAL in user.mutations)
+		attack_generic(user, 15)
+		return
```
CPU-Blanc, line 219: This isn't necessary? In the parent call done in the next line, this is essentially already done
https://github.com/UristMcStation/UristMcStation/blob/12e0ffe8681baa752a6e68a4b324eca269b0cf3a/code/game/machinery/_machines_base/machinery.dm#L237-L243

code/_onclick/other_mobs.dm
```diff
@@ -32,6 +32,8 @@
 				SPAN_DANGER("\The [user] smashes through \the [src]!"),
 				SPAN_DANGER("You smash through \the [src]!")
 			)
+			if (MUTATION_FERAL in user.mutations)
+				qdel(src)
```
CPU-Blanc, line 36: I'm not sure I like this qdel. Is it necessary? Shouldn't the atoms decide how they handle destruction if their health reaches 0? Unless I'm reading this proc wrong..

code/game/machinery/doors/door.dm
scrdest: Keep in mind, I just cherrypicked Bay directly aside from qdel -> drop stuff.

I assume this is a bypass of the early return in the parent.

code/_onclick/other_mobs.dm
scrdest: It does seem quite goofy, blame upstream.

Glloyd:
This is just a straight Bay cherrypick, so if we want to improve the code, we can do that in the future. It seems to work fine as is.

PR #1283, opened 2023-10-06
new dolls

CaraS212:
Adds 2 new dolls to the loot list.

makes one of the lactera infested apartments into a neet den from the former inhabitant.


CPU-Blanc (approved):
Bocchers

PR #1315, opened 2023-10-10
fix: Fixes oregen never actually looking up the map by index and instead using the index itself FOR EIGHT YEARS, lmao

scrdest:


scrdest:
Superseded by #1316 (all according to keikaku)

PR #1317, opened 2023-10-10
NPC bugfixes

### LINKED ISSUE #1279: NPC Runechat looks garbled/non functional
### Description of issue

NPCs when talking into Runechat don't make completley visible text, which is an issue for any newer NPCs that talk or shout barks, I've ran into a similar issue with https://github.com/UristMcStation/UristMcStation/pull/1263  Where I've added lines for Maneaters, and I can barely see the text when spoken.

image

Example from Discord

All text when spoken should be visible in Runechat above the character.

### Steps to reproduce

- Spawn an NPC that speaks dialogue, or shouts barks.
- Look at the Runetext above them

It seems all Simple Mobs have the same issue, making me think it's related to Runechat itself.

Current Baymerge.

### Issue bingo

### LINKED ISSUE #1302: Simple mobs attack exosuits without pilots
### Description of issue

Simple mobs will destroy exosuits regardless of who or if anything is inside them.

Less hatred towards unactive pieces of metal

### Steps to reproduce

Spawn exosuit, spawn hostile simplemob, observe

This is due to exosuits being /mob/living subtypes with a default faction of neutral

Client Version: 514
Server Revision: [e16a6956a9e107b9070d46043020346113a5325e](https://github.com/UristMcStation/UristMcStation/commit/e16a6956a9e107b9070d46043020346113a5325e) - 070d46043020346113a5325e - 2023-10-06
Game ID: cqa-atjF
Current map: ICS Nerva

### Issue bingo

CPU-Blanc:
Couple of small NPC bufixes

- Hostile AIs now respect `destructive` and don't always attack exosuits, regardless of if they're empty (Fixes #1302 ). Caused by the fact exosuits are indeed "living", so the second type check was essentially unreachable.
- Tweaked the aforementioned targeting logic. Previously the AI would pick the first pilot of the exosuit and decide to attack or not based off of them. This has been reworked slightly to check if the AI can attack _any_ of the pilots, weighted towards attacking.
- NPCs no longer speak in rune speak in floating text (Fixes #1279 ). Caused the NPCs actually having no language. As this is auto-generated from the NPC's species' default language during `New()`, humankind's default has been set to GalCom. I don't _think_ this will cause any unintended side-effects. But might need another pair of eyes to verify.

### DIFF
--- a/code/modules/culture_descriptor/culture/cultures_human.dm
+++ b/code/modules/culture_descriptor/culture/cultures_human.dm
@@ -5,6 +5,7 @@
 /singleton/cultural_info/culture/human
 	name = CULTURE_HUMAN
 	description = "You are from one of various planetary cultures of humankind."
+	default_language = LANGUAGE_GALCOM
 	secondary_langs = list(
 		LANGUAGE_GALCOM,
 		LANGUAGE_SOL_COMMON,
--- a/code/modules/ai/ai_holder_targeting.dm
+++ b/code/modules/ai/ai_holder_targeting.dm
@@ -122,23 +122,24 @@
 
 			if (L.status_flags & NOTARGET)
 				return FALSE
+
+		else if (istype(the_target, /mob/living/exosuit))	//Exosuits are living??? This check needs to go in here now
+			var/mob/living/exosuit/M = the_target
+			var/will_attack = FALSE
+			for (var/mob/pilot in M.pilots)
+				will_attack =  will_attack || can_attack(pilot)
+			return length(M.pilots) ? will_attack : destructive // Empty mechs are 'neutral'.
+
 		if (L.stat)
 			if (L.stat == DEAD && !handle_corpse) // Leave dead things alone
 				return FALSE
 			if (L.stat == UNCONSCIOUS)	// Do we have mauling? Yes? Then maul people who are sleeping but not SSD
-				if (mauling)
-					return TRUE
-				else
-					return FALSE
+				return mauling
+
 		if (holder.IIsAlly(L))
 			return FALSE
 		return TRUE
 
-	if (istype(the_target, /mob/living/exosuit))
-		var/mob/living/exosuit/M = the_target
-		for (var/mob/pilot in M.pilots)
-			return can_attack(pilot)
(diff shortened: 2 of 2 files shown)


scrdest (approved):
LGTM

code/modules/culture_descriptor/culture/cultures_human.dm
```diff
@@ -5,6 +5,7 @@
 /singleton/cultural_info/culture/human
 	name = CULTURE_HUMAN
 	description = "You are from one of various planetary cultures of humankind."
+	default_language = LANGUAGE_GALCOM
```
Glloyd, line 8: this shouldn't be a problem. it was like this before Bay fully removed GalCom from the game anyways.

Glloyd: this will also fix adminspawned humans not having any language actually

ggermanicus:
@CPU-Blanc Since player humans now get GalCom by default, this forces it in the charsetup and effectively eliminates one of their language slots and deprives awaygoers of the opportunity not to have GalCom. I am not a huge fan of either; the first is a human nerf, and awaygoers not having GalCom could be pretty fun. Is there any workaround that fixes the bug and doesn't give humans universal Galactic Common?

Vakothu:
Yeah, this unfortunately makes GalCom, which is European English, mandatory for every single human culture full stop, even the ones that already had a default language. For example, if you decide to be a Lunarian you only get a single language slot, because GalCom and Selenian are both mandatory.

CPU-Blanc:
A fix for this will rather annoying and jank to handle, as this is baked into the `say` proc for simple_animals; using the mob's `species_language`, which is generated during the mob's `New` call from their culture's default language)

I've got a couple of ideas in mind - I'll poke about and see what I can manage. However as Glloyd said above, this was the old way it was done and also fixes admin spawned humans having no language

Vakothu:
Or do the extremely hacky fix and have all the other human cultures have `default_language = _null_`

CPU-Blanc:
Got a slightly non-hacky fix. I'll tag you both in the PR :+1:

PR #1318, opened 2023-10-11
actually fixes xenophage larva faction

Glloyd:
what #1311 failed to consider is that this code is bad, and xenophage larva don't get their faction from the species datum.


Naxodile:
I swear I tested this and it got overridden by larva becoming antags

Glloyd:
it does, but aliens will target and kill hatching larva before their faction is assigned by the antagonist setup. Also, this makes adminspawned larva not be targeted as well.

PR #1320, opened 2023-10-11
feature: FoV Cones port from Nebula

scrdest:
With some refactoring and compile-time-, config- and mob- level switches.

TL;DR if either the config option ENABLE_FOV_CONES is enabled or `force_fov=TRUE` is set on the mob, the mob gets a roughly 90-degree wide transparent dark overlay behind them.

All mobs and items within that zone (with some exceptions like mobs grabbing and pulling you) are invisible. Obscured mobs walking around generate a small footstep visual effect.

I've refactored the original Nebula code to add a bunch of logic to add config flags and make this set up and tear down nicely if the options are toggled; most of the override logic does not process if the mob in question doesn't have FoV enabled.

Additionally, if we're REALLY concerned about performance, the whole thing can be removed at compile-time by changing a single Urist define.

[FoVDemo1.webm](https://github.com/UristMcStation/UristMcStation/assets/4363639/f3e35b15-91b1-4396-b193-c7110b76e2f9)


scrdest:
> you'll need to add the new option to the config. separately, I would like an admin button to enable vision cones midround. both for testing and !!fun!! purposes.

I'll add it defaulting to commented out for now.

Right now, you can adminbus it quite easily through MC -> Config -> varedit (that, in fact, is how I was testing this and verifying it can be toggled on/off cleanly).

I can add a verb easily enough, but keep in mind it will add to the button-pollution of the admin interface.

Yantarii:
Would be cool to test this on for a few rounds this week and see how it fairs

Glloyd:
@scrdest if you can update your branch, we can try this out on the server tonight.

scrdest:
> @scrdest if you can update your branch, we can try this out on the server tonight.

On it

code/modules/urist/modules/fov/overrides/cone.dm
```diff
@@ -0,0 +1,130 @@
...
+			I = image("split", M)
+			I.mouse_opacity = 0
+			I.override = 1
+			src.client.images += I
+			src.client.hidden_atoms += I
+			src.client.hidden_mobs += M
+			if(LAZYISIN(M.grabbed_by, src))//If we're pulling them we don't want them to be invisible.
+				I.override = 0
+
+			M.in_vision_cones[src.client] = 1
```
scrdest, line 120:
```suggestion
			if(LAZYISIN(M.grabbed_by, src))//If we're pulling them we don't want them to be invisible.
				I.override = 0
			
			if(src.client)
				src.client.images += I
				src.client.hidden_atoms += I
				src.client.hidden_mobs += M

				M.in_vision_cones[src.client] = 1
```

code/modules/urist/modules/fov/overrides/cone.dm
```diff
@@ -0,0 +1,132 @@
...
+		if(!O.InCone(center, dir))
+			true_conelist -= O
+	return true_conelist
+
+/mob/living/carbon/human/update_vision_cone()
+	set waitfor = FALSE
+
+	if((src.force_fov == FALSE) || !(config.enable_fov_cones || src.force_fov))
+		var/delay_cleanup = 10
+		for(var/image/HI in src.client.hidden_atoms)
```
scrdest, line 66:
```suggestion
		for(var/image/HI in (src.client?.hidden_atoms || list()))
```

scrdest:
AI ROUGE

Glloyd:
Were the issues with FOV not working properly ever fixed?

scrdest:
Some of them were, but I'd need to pull some testers together or do some shenanigans to connect from another machine to test it properly.

Most of what caused trouble are edge-cases with things like clients disconnecting/reconnecting or players seeing each other IIRC.

CPU-Blanc:
If a playtest is all that's required currently, could we not just checkout + build this branch and host it for a night and see how it goes?

Would probably need a rebase/merge from master first though

scrdest:
@CPU-Blanc say less fam

scrdest:
Caught some runtimes in solo testing, but fixed now. Fixed the config not working too =_=

No runtimes solo now, tested flipping the cones on an off and that worked fine too, tested with a hostile to check mobs - works (although it's now _terrifying_ to try to flee a bulletspammer like a Skrellorist - you have to hear the bullets zip behind you _somewhere_ and pray for Baymiss).

CPU-Blanc:
As we've had good pop lately, could this branch be tested on the server this weekend some point @Glloyd ?

Glloyd:
I do want to get back to this, I've just been occupied with baymerging. This seems like it should be relatively straightforward to rebase onto the new baymerge branch, so once that is in, we can test out this, and GOAI again on the updated branch.

PR #1321, opened 2023-10-12
Hunger and thirst give slowdown

ggermanicus:
No effects if you are 'not' or 'a bit' hungry/thirsty, after that being quite hungry/starving and being quite thirsty/dying of thirst each incur their own slowdown penalties.


Glloyd:
adding mechanical effects to hunger/thirst is long overdue.

PR #1329, opened 2023-10-12
Research Wing Layout Changes + Medbay Breakroom APC Fix

Yulisapilled:
Completely reworks the layout of the Research wing, including the small second deck office. Added a small reception/lobby area and centralized atmospherics for all three sub-departments. Expedition prep, Xenoarch, and Anomaly research have all been moved to be closer to the Hadrian (which also moved slightly). Xenoflora has been shrunk slightly. Xenobiology is largely expanded, but with 3 cells instead of 4.  Minimal changes were made to maintenance around the starboard side of the research wing and disposals.

Also included a fix for the First Deck Medbay Breakroom APC, which previously was just an unwired frame laying on the floor.

image


scrdest:
I'm not sure about Xenobio cells... having more cells probably helps manage different blobs better than fewer, bigger cells.

CPU-Blanc:
Thanks for the PR!
I'll take a good look through this later today, however, for now you've got multiple failing tests for the Nerva map. If you're confused/need help understanding those, let me know!

CPU-Blanc:
Did you use the mapmerge tool on nerva-1.dmm? Doesn't look like you did.

CPU-Blanc:
Did you use the mapmerge tool on `nerva-1.dmm`? Doesn't look like you did to me.

maps/nerva/nerva-1.dmm
CPU-Blanc, line 1: ### Nanotrasen Office (Lower)

- General lack of floor decals here. All just white tiles. Could we add some boarders or something to break it up?
- Lots of double single-tile airlocks. Would double-tile airlocks look better?
- Paper bin in the lobby area has no pen. Should probably have one
- Newscaster in the lobby area overlaps awkwardly with the light on the left. The light can simply be deleted (There's plenty of light there)
- Holopad in the sitting area is in a but of an odd location. Could we make it more central? Inline with the bottom table and water cooler could work nicely
- Lobby console uses the AI slot type. This should be replaced with the `preset/nervasci` type instead
- Scrubber under a vending machine in both the sitting area and lobby. Please try to avoid placing scrubbers/vents like this if possible!
- The sitting areas is quite a large room for only 1 vent+scrubber. Could we add another pair?

### Senior Scientist Office

- Console also has the AI slot preset. As science is _not_ actually a part  of the crew directly under the captain, I'd argue against this. Should only be possible on the bridge or AI core
- Currently has no windows into science itself. The airlock could be changed to a glass one, and a window installed under it (the lockdown button will have to be moved)
- The painting in the top right looks awkward where it is, as does the requests console overlapping with the light. Could we perhaps move the painting onto the top wall somewhere (perhaps above the console), and move the request console next to it?
- Lack of window  tint control. Would be nice to allow the Senior Scientist some privacy without the people in the lobby gawking in
- Chair spawns at an odd place and facing an odd direction. Could we face it east, and maybe  nudge it up one tile to sit at the console?

### Gas Storage

- I'd like to see vision restricted here to allow antags to operate. Could we replace the glass airlock with a non-glass one and remove the windows?
- There's a lack of access here too. If you were working xenobio and wanted to turn off the freezer, you have to walk all the way to the lobby, through that interior door, through to xenoflora, then back up into the storage room. Could we add an airlock either where the top window currently is, or perhaps to the right above the vent?

### Xenobiology

- Some of the lockdown shutters aren't facing the right direction. Namely the far right pen, and the two leading to the lobby area
- Sink is facing the wrong direction
- Newscaster is in a really odd position here. Could we move it perhaps into the seating area in the office part?
- "Xeno" poster is placed on the wall, making it visible from maintenance. It should be placed on the tile above, then nudged down
- I agree with scr. With these pens, quantity tends to be more useful than size.
- The pens should really have two tiles space between the two interior doors. This allows scientists to trap themselves in with the slimes so they cannot move, and efficiently kill them with water.
- Pen interior doors should probably be named much like they currently are on the live server
- The wall splitting the top two pens, those back-to-back lights can and probably should be removed. Looks odd with the poster and there's generally too much light here.
- The disposal outlet that leads to space, it's pipes currently run through a LOT of walls. As a general rule, we try to keep piping and wires running under tiles and airlocks rather than walls. Can we reroute it somehow here?
- The disposal outlet to space should have hatched markings
- The pen on the right has a scrubber under the disposals outlet. Can we move this?
- The intercom is a little tucked out of the way in the corner, can we make it a little more central?
- Only the top left pen has warning marks outside, the others do not. This looks strange.
- Quite a large room for 1 vent+scrubber pair. Probably needs 2

### Xenoflora

- The biohazard locker on the left and the scientist locker at the bottom should probably swap positions. Bit odd like this currently
- That extra shower over where the biohazard locker is looks out of place. We already have one below with a curtain. Perhaps this can be removed?
- Emitters need to be hardwired. You'll need to route a cable to it, much like an APC terminal
- Under the emitter, you currently have 2 overlapping hazard stripes. One can be deleted.
- The manifold under the disposal bin is visible. This should be replaced by the hidden subtype
- You've got chromatic lights, but as far as I can see no way to control them? The button should be added
- The "sci" poster has the same problem as the xeno poster, it's on the wall. It either needs to be placed a tile up or down and then shifted, depending on what side you want it visible from. The wall under is also missing its fabulous pink paint
- The requests console to the right has the same problem. It should be moved up one tile, then shifted down
- The "Xeno" poster on the left wall also has the same issue
- The intercom on the right side is a little cramped with that camera. Could we perhaps move it up a few tiles? The room is also rather large, so another one on the otherside might also be a good idea.
- Some emergency shutters are missing in the isolation area. At the very least, the glass tile left of the access airlock needs one.
- The walls shared with cargo shouldn't be painted pink

### Expedition Prep

- Random shut-off valve decal under the bottom locker. This should be removed
- The airlocks internal to the department (leading into xenoflora) should probably be glass
- Once again, scrubber and vent under the suit storage/cycler. Please avoid this

### General

- The science wing shut off valve has been removed. Now cargo and science share one. Could this be reworked so that they are independent again?
- Honestly? I'm _personally_ not a fan of the arrow decals next to the valve. Might just be because they're not used anywhere else, but they're a little out of place
- Some general TLC; there's just a little lack of decals and such going on and a lot of unbroken white tiles. Look at the current xenoflora and xenobiology for some inspiration c:
- Try to avoid placing scrubbers + vents under tables and such unless it's really necessary.

CPU-Blanc (commented):
Had a little look, on the whole, I like it. Just a few things I've noticed

maps/nerva/obj/nerva_machinery.dm
```diff
@@ -25,3 +25,6 @@
 	mask = /obj/item/clothing/mask/breath
 	req_access = list()
 	islocked = 1
+// Research Specific Air Alarms
+/obj/machinery/alarm/nervasci
+	req_access = list(list(access_xenobiology, access_atmospherics, access_engine_equip))
```
CPU-Blanc, line 30: This for me plus the area extending outside of isolation to trick the air alarm into covering those rooms in a no-go. If we're only doing this in 2 places (which I'll get around to in a moment), this doesn't need to be a new subtype define, just make a new instance in the map editor and set the vars there. This is unnecessary 

If we're doing this do give sci the ability to control the vents in the isolation room and such, then I'm personally against this approach. Rather than using the air alarms as a hacky way to do it, they should have an injector and be controlled remotely by an air control console.

CPU-Blanc (changes requested):

ErrorsWindows:
does it fix the thing where i can't open the desk windoors on the hallway side even though I need them to be open to put things on the desk

maps/nerva/nerva-1.dmm
Yulisapilled: Excellent feedback, thanks, I'll look at implementing most of the changes/fixes later this weekend.

Glloyd:
unstaled this PR because i would like to see this picked up and finished at some point even if Yulisa isn't able to finish it.

CPU-Blanc:
I'll probably adopt this PR if Yulisa hasn't returned to it in a few weeks when I'm off from work.

CPU-Blanc:
Superseded by #1408 
Further contributions by the OP can be PR'd to my branch if they return

PR #1330, opened 2023-10-12
adds a contract + a couple more things to derelict

ggermanicus:
- Adds a derelict contract to re-setup the singularity.
- Adds the necessary parts to actually do the above to the derelict.
- Fixes only some of the cursed wires, makes a couple guns empty, and adjusts some doors on the derelict.
- Removes an unnecessary and incorrect ammo label on U2442 mags.

### DIFF
--- a/maps/away/russianderelict/russianderelict.dm
+++ b/maps/away/russianderelict/russianderelict.dm
@@ -82,6 +82,7 @@
 	name = "\proper космическая-станция-13"
 	desc = "Sensors detect an orbital station with an unusual profile and no life signs."
 	icon_state = "object"
+	assigned_contracts = list(/datum/contract/russianderelict)
 	known = FALSE
 
 /datum/map_template/ruin/away_site/russianderelict
@@ -221,6 +222,12 @@
 	icon = 'icons/obj/drinks.dmi'
 	icon_state = "art_bru"
 
+/obj/item/gun/projectile/automatic/spaceak/empty
+	magazine_type = null
+
+/obj/item/gun/projectile/automatic/c20r/empty
+	magazine_type = null
+
 /obj/random/single/russiancola/spawn_choices()
 	return list(
 		/obj/item/reagent_containers/food/drinks/cans/syndicolax,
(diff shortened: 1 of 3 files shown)


maps/away/russianderelict/russianderelict.dm
```diff
@@ -252,3 +259,34 @@
...
+/obj/machinery/the_singularitygen/russianderelict
+	var/contract_enabled = FALSE
+
+/obj/machinery/the_singularitygen/russianderelict/Initialize()
+	.=..()
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/machinery/the_singularitygen/russianderelict/LateInitialize()
+	if(!contract_enabled && GLOB.using_map.use_overmap)
+		contract_enabled = TRUE
```
Glloyd, line 281: this whole bit really isn't necessary. the reason i did it this way for the destroyed_colony one is because the radio beacon spawns an object on the overmap when the distress signal is activated, which causes runtimes when the map is loaded without an overmap, such as in awaymap testing. The contract_enabled var was because deactivate() could be called multiple times on the beacon, so it prevents it from checking for contract completion every time it's deactivated. However, the singulogen qdels itself after it is activated, so that isn't an issue here either.

maps/away/russianderelict/russianderelict.dm
```diff
@@ -252,3 +259,34 @@
 
 /obj/effect/floor_decal/urist/russian
 	icon_state = "derelict1"
+
+//contract
+
+/datum/contract/russianderelict
+	name = "Derelict Repower Contract"
+	desc = "Restore power to this station by setting up the singularity."
```
Glloyd, line 267: This is the description that shows up in the contract database, so this should probably be expanded on a little. I had humoured the thought of adding in a USSSR faction datum to pay the money out from, but it's up to you whether you want it to be an NT contract or not. If you think that sounds neat, i'll add it in right now and you can set the contract faction as "russian". Otherwise you don't need to set the faction, because it autosets it as GLOB.using_map.trading_faction if one isn't set.

Glloyd (changes requested):

maps/away/russianderelict/russianderelict.dm
ggermanicus: It's a neat concept but I don't think this contract alone justifies a new faction that otherwise has very little lore presence compared to our other ones.

maps/away/russianderelict/russianderelict.dm
ggermanicus: Done along with slight desc expansion

Glloyd:
Looks good 👍

PR #1342, opened 2023-10-15
Reverts GalCom default human language + NPC floating text alternative fix 

CPU-Blanc:
As raised in the comments of #1317, GalCom being the default language for humans had some other side effects, namely forcing away map players to adopt GalCom, and essentially forcing human players to lose a language slot.
I personally have no skin in the game, so I don't mind if reverting this is rejected or not.

Tagging @Vakothu & @ggermanicus for their input

Changes:

- Galactic Common removed as human's default language
- `simple_animal`'s `say` proc has been expanded to accept an optional language string. If none is provided, the mob's `species_language` is instead used (this defaults to their culture's default language)
- NPCs speak Galactic Common by default in their `say` proc. This can be overloaded for fun.

Only issue with this is it does unfix something mentioned by @Glloyd in the previous PR , which is that admin-spawned NPCs having no languages will be a thing again

### DIFF
--- a/code/modules/culture_descriptor/culture/cultures_human.dm
+++ b/code/modules/culture_descriptor/culture/cultures_human.dm
@@ -5,7 +5,7 @@
 /singleton/cultural_info/culture/human
 	name = CULTURE_HUMAN
 	description = "You are from one of various planetary cultures of humankind."
-	default_language = LANGUAGE_GALCOM
+	// default_language = LANGUAGE_GALCOM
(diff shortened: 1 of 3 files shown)


ggermanicus (approved):

Glloyd:
considering admins can easily add languages, i'm not too concerned about that issue. i'll just have to remember that it's a problem so i don't babble wordlessly into a holopad while trying to threaten the Nerva again.

PR #1343, opened 2023-10-16
Fix A22 rifle not accepting it's magazine/Wrong Caliber + Flamethrower Typo

The32bitguy:
Changes:

- Flamethrower no longer gives a double `the the` when it splutters
- A22 ANFOR rifle now uses the caliber listed in it's description, is able to have it's own magazines reinserted, and it's magazines no longer double on mentioning their caliber

### DIFF
--- a/code/modules/urist/gamemodes/infestation/weapons&armour.dm
+++ b/code/modules/urist/gamemodes/infestation/weapons&armour.dm
@@ -196,7 +196,7 @@
 	icon = 'icons/urist/items/guns.dmi'
 	icon_state = "ANFOR-riflemag"
 	mag_type = MAGAZINE
-	caliber = CALIBER_RIFLE
+	caliber = CALIBER_RIFLE_MILITARY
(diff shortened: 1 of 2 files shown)


code/modules/urist/gamemodes/infestation/weapons&armour.dm
```diff
@@ -167,11 +167,12 @@
 	item_state = "ANFOR-rifle"
 	w_class = 4
 	force = 10
-	caliber = CALIBER_RIFLE
```
CPU-Blanc, line 170: This change requires the same be made to its magazine here:
https://github.com/UristMcStation/UristMcStation/blob/be69851be262cbf2f7d7e77d85c134e80255f3ec/code/modules/urist/gamemodes/infestation/weapons%26armour.dm#L199

CPU-Blanc (changes requested):

CPU-Blanc (approved):

PR #1346, opened 2023-10-16
Fixes severely impaired breathing speech

Naxodile:
Not the most elegant fix but this should prevent severe lung damage from causing the first character to become a emote prefix.


Glloyd:
definitely not the most elegant, but if it works it works.

PR #1355, opened 2023-10-18
Fixes maintenance light deconstruction

### LINKED ISSUE #1319: Maint. Light Fixture Issues
### Description of issue

Deconstructing the Maintenance Light fixtures (/obj/machinery/light/small/red) does not deconstruct them, it instead spawns a normal small light fixture underneath it.

_No response_

### Steps to reproduce

Go into maintenance or spawn on the Noctis
Try to deconstruct a maintenance light fixture
Spawns in a normal light fixture without deleting the maintenance once

_No response_

Client Version: 514
Server Revision: [ec61bad64d715d5dac9c9c5532113b704506a5a8](https://github.com/UristMcStation/UristMcStation/commit/ec61bad64d715d5dac9c9c5532113b704506a5a8) - ac9c9c5532113b704506a5a8 - 2023-10-09
Game ID: cqe-apEf
Current map: ICS Nerva

### Issue bingo

CPU-Blanc:
Fixes #1319 
This was caused by a duplicate definition of `/obj/machinery/light/small/red` causing some wonk behaviour.
As the bay version of `/obj/machinery/light/small/red` emitted a darker red colour (kind of like the emergency light colour), I've kept both. The red light we were using is now `/obj/machinery/light/small/red/maintenance`.

Currently I've only swapped the Nerva over to use this old colour, everywhere else what used the red light fixture will be using the bay version.

A comparison between the two

### Bay's red light

Again, with this PR the Nerva will be using the Urist red, whereas anywhere else will use the new Bay red. If you want to revert this, or have more maps use the Urist red, let me know!

### DIFF
--- a/code/modules/urist/structures/uristlighting.dm
+++ b/code/modules/urist/structures/uristlighting.dm
@@ -6,24 +6,16 @@
 ///////////////////////////////////////////////////////////////////////////
 
 //spoopy red maint lights, Alien-style
-/obj/machinery/light/small/red
+/obj/machinery/light/small/red/maintenance
 	icon_state = "firelight1"
-	construct_type = /obj/machinery/light/small
 	active_power_usage = 2
 	name = "Maintenance light fixture"
 	desc = "A small, low-power lighting fixture used for maintenance lighting."
-	light_type = /obj/item/light/bulb/red
+	light_type = /obj/item/light/bulb/red/maintenance
 
-/obj/machinery/light/small/red/New()
-	..()
-	icon_state = "firelight1"
-
-/obj/item/light/bulb/red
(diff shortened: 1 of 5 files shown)


Glloyd:
i'm fine with it either way really, what do you think?

CPU-Blanc:
Honestly either works. The Bay red is a bit darker and reduces visibility a bit, but it depends on what aesthetics you want to go for.

Removing our version would streamline the code a bit however, as I can just remove it and not touch the Nerva dmms

Glloyd:
I think this is fine as is. We can always revisit it though, our lights are pretty old and might even predate smooth lighting, so the values could probably stand to be tweaked.

PR #1270, opened 2023-10-05
Adds new NPC traders

CaraS212:
Adds some new traders:

- Clothing (casual, formal, hats, shoes, coats, and accessories)
- Artifact Dealer for submap roles with a ship to sell gems and other trinkets
- Components trader: buys and sells stock parts, has a limited inventory of 2nd and 3rd tier parts for those who lack RnD access.
- -Components trader sell prices for ~1.5-2x the value on worths_list.dm (if we can find a way to make this only for the sell to player price, rad, but its not super worth while to print off a bunch of components and load them in a duffle bag to sell to the guy.)

adds an artifact dealer to the bar on the NT station, components trader stands next to the crop trader

PR #1356, opened 2023-10-20
Reopens traders

CaraS212:
i fucked up

https://github.com/UristMcStation/UristMcStation/pull/1270


CPU-Blanc (approved):
As before, LGTM

code/modules/urist/modules/newtrading/trade/trade_items/stock_parts.dm
```diff
@@ -0,0 +1,119 @@
+/datum/trade_item/stock_parts
+	quantity = 8
+	trader_weight = 1
+	value = 160
```
Glloyd, line 4: i've mentioned before, but i don't like just assigning the same value to all basic stock parts. these prices will also affect the prices for trading being done through cargo as well, so they need a little more consideration than a flat number. I'd lean towards letting get_worth handle it and leaving the value null, but i'm not sure what get_worth returns for these objects though.

CaraS212: what would that look like? just: get_worth = what ever?

PR #1359, opened 2023-10-22
Very quick washing machine fix

Vakothu:
Untested, but this is how Bay fixed components unable to be added to a washing machine in https://github.com/Baystation12/Baystation12/pull/33669


Glloyd:
do we need the other fixes from that PR?

Vakothu:
I have no idea if we do or not, I just grabbed as little of the code as possible to avoid any incompatibility that could result. We might wanna test it.

Glloyd:
if they are issues we have, it's probably best to grab it all

Vakothu:
Gonna close this and cherry pick that commit then.

PR #1363, opened 2023-10-24
Various Payroll Bugfixes + Improvements

CPU-Blanc:
This started as a simple bugfix, and once again, turned into much more...

- Fixes runtimes caused when looking at the transaction history for an account. Caused by the transaction datums requiring both entries to be real accounts, not pseudo accounts. These have been changed to use `/datum/transaction/singular` instead.
- Fixes another runtime when looking into the transaction history for a newly created account. Caused by.. some really wonky `new()` arguments passed in account creation. This must've changed at somepoint... I hope
- Fixes the finesmanager not being able to detect the card reader. Not sure if this somehow changed in the merge or at some point? Because this worked in testing and on the live server before? Caused by a data field using the wrong key
- Scientists now get an email from NanoTrasen. I thought it was a little odd that the Nerva would email scientists about their pay, so they now get one from NanoTrasen. On top of this, I can't imagine NanoTrasen subsiding their emails/payroll by taking sponsorships, so instead our corporate overlords will send us a lovely randomly picked message, reminding us of all of the glory of space capitalism.
- The radio sponsor product now matches the email product.
- Added `is_credit`, `get_transaction_amount` and `get_transaction_ledger` helper methods to `/datum/money_account`. This is a bit of a rabbit hole... But essentially `/datum/transaction` does not give a damn whether the amount is positive or negative. If it is negative, it simply removes from `target` and places it in `source`. This is fine by itself, however, `/datum/money_account/transfer()` uses a single transaction for both the source and target accounts, using a boolean in `/datum/money_account/add_transaction()` to denote if funds should be removed or added to the specified account. The problem is that the amount is **always** positive in the single transaction datum both accounts point to, and there's no record in the transaction itself of which way around this occurred. Coupling this with the previous negative amount business means for any singular transaction, you have _no_ clue if money was added to _your_ account or not. Just taking the transaction and seeing a positive value is not enough. These helper methods are all aware of the calling account and return results in relation to that account alone. 
- A lot of general clean up while I was in here. Murdering a lot of unnecessary indents.

### DIFF
--- a/code/controllers/subsystems/payment.dm
+++ b/code/controllers/subsystems/payment.dm
@@ -41,9 +41,9 @@ SUBSYSTEM_DEF(payment_controller)
 	//Don't forget our coporate overlords!
 	var/list/nanotrasen_nags = list(
 		"NanoTrasen staff are reminded that salaries cannot be re-negotiated once employment has commenced",
-		"NanoTrasen would like to remind staff that the Intellectual Property rights of any inventions and/or concepts created during your employment with NanoTrasen are waved in their entirety to NanoTrasen",
+		"NanoTrasen would like to remind staff that the Intellectual Property rights of any inventions and/or concepts created during your employment with NanoTrasen are waived in their entirety to NanoTrasen",
 		"Staff are reminded that unionizing is in direct violation of your employment contract. Dissatisfied staff are to book an appointment with a NanoTrasen Productivity and Morale Advisor, who will arrange an appointment with you within the next 895 working day(s)",
-		"NanoTrasen staff are reminded that minimising overhead and wastage to NanoTrasen is of paramount importance, as well as everyone's joint responsibility",
+		"NanoTrasen staff are reminded that minimising overhead and wastage is of paramount importance to NanoTrasen, and is everyone's joint responsibility",
 		"NanoTrasen would like to remind contracted staff that failure to meet expectations during your annual performance review, can and will result in contract termination (termination fees applicable)",
 		"Staff are reminded that termination of the employment contract will incur a fee of 38.9% of your agreed annual salary for each year you have been employed by NanoTrasen, retrospectively adjusted for current inflation rates",
 		"NanoTrasen reserves the right to reimburse lost productivity caused by staff absence from the workplace for any reason (such as sickness), automatically from staff salary payment(s) until the lost sum is recovered in its entirety",
--- a/code/controllers/subsystems/payment.dm
+++ b/code/controllers/subsystems/payment.dm
@@ -2,63 +2,116 @@ SUBSYSTEM_DEF(payment_controller)
 	name = "Payment"
 	wait = 15 SECONDS
 	var/timerbuffer = 60 MINUTES //buffer for time check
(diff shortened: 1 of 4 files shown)


code/controllers/subsystems/payment.dm
```diff
@@ -2,63 +2,116 @@ SUBSYSTEM_DEF(payment_controller)
...
+		"The proud result of prison labor!",
+		"No refunds",
+		"For the sophisticated shut-in",
+		"Tell your parents it's educational!"
+	)
+
+	//Don't forget our coporate overlords!
+	var/list/nanotrasen_nags = list(
+		"NanoTrasen staff are reminded that salaries cannot be re-negotiated once employment has commenced",
+		"NanoTrasen would like to remind staff that the Intellectual Property rights of any inventions and/or concepts created during your employment with NanoTrasen are waved in their entirety to NanoTrasen",
```
Glloyd, line 44:
```suggestion
		"NanoTrasen would like to remind staff that the Intellectual Property rights of any inventions and/or concepts created during your employment with NanoTrasen are waived in their entirety to NanoTrasen",
```

code/controllers/subsystems/payment.dm
```diff
@@ -2,63 +2,116 @@ SUBSYSTEM_DEF(payment_controller)
...
+		"Staff are reminded that unionizing is in direct violation of your employment contract. Dissatisfied staff are to book an appointment with a NanoTrasen Productivity and Morale Advisor, who will arrange an appointment with you within the next 895 working day(s)",
+		"NanoTrasen staff are reminded that minimising overhead and wastage to NanoTrasen is of paramount importance, as well as everyone's joint responsibility",
```
Glloyd, line 46:
```suggestion
		"NanoTrasen staff are reminded that minimising overhead and wastage is of paramount importance to NanoTrasen, and is everyone's joint responsibility",
```

code/controllers/subsystems/payment.dm
```diff
@@ -2,63 +2,116 @@ SUBSYSTEM_DEF(payment_controller)
...
+		"NanoTrasen would like to remind contracted staff that failure to meet expectations during your annual performance review, can and will result in contract termination (termination fees applicable)",
+		"Staff are reminded that termination of the employment contract will incur a fee of 38.9% of your agreed annual salary for each year you have been employed by NanoTrasen, retrospectively adjusted for current inflation rates",
+		"NanoTrasen reserves the right to reimburse lost productivity caused by staff absence from the workplace for any reason (such as sickness), automatically from staff salary payment(s) until the lost sum is recovered in its entirety",
+		"Staff are once again reminded that comfort breaks are indeed on personal time, and as such will be docked from your pay entitlement"
+	)
```
Glloyd, line 51: I had an idea for another. "NanoTrasen staff are reminded that payment processing fees are automatically deducted from your hourly pay entitlement, along with the mandatory convenience fees for the automatic deduction of the aforementioned processing fees"

Glloyd (changes requested):
just a few little formatting things and a suggestion

code/controllers/subsystems/payment.dm
CPU-Blanc: Love it

CPU-Blanc:
Thanks for those. As a dyslexic, writing is not my forte :sweat_smile:

Glloyd:
They were all very minor things, it looks good overall. Once tests clear I can merge this.

Glloyd (approved):

PR #1365, opened 2023-10-31
gives nerva assistants access to the maintenance tunnels

ErrorsWindows:
read title

yep they didn't have this

🆑 

* tweak: Nerva assistants have access to the maintenance tunnels

### DIFF
--- a/maps/nerva/datums/nerva_jobs.dm
+++ b/maps/nerva/datums/nerva_jobs.dm
@@ -298,6 +298,7 @@
 
 /datum/job/assistant
 	supervisors = "the second officer."
+	access = list(access_maint_tunnels)
 	alt_titles = list(
 	"Technical Assistant","Medical Intern","Cargo Assistant",
 	"Botanist" = /singleton/hierarchy/outfit/job/service/gardener,


CPU-Blanc:
I believe this was intentional. @Glloyd ?

ErrorsWindows:
I mean I really don't think so but let's double check

Screenshot (2573)

CPU-Blanc:
> I mean I really don't think so but let's double check
> 
> Screenshot (2573)

They 100% did not have it prior to the merge. These datums are in the Nerva map folder, so wouldn't be touched by uptream. Also checking the mater branch confirms this

Glloyd:
> I believe this was intentional. @Glloyd ?

it was intentional back in the glloydstation days, and we had a vote deciding that assistants shouldn't have maint access. It hung around for ages because we wanted people to play actual roles instead of assistant. But doing a quick discord search, the last time the issue was revisited was 2016, so i'd be fine with giving it a try, seeing as we're probably 9 years out from the original vote. I'll set up a quick poll though to see what people think.

CPU-Blanc:
Why close this? A vote was underway with the majority (at time of writing) supporting this change

ErrorsWindows:
i had an epiphany

Glloyd:
i'll take this over from here, and approach the issue in a slightly different way.

PR #1367, opened 2023-11-03
Adds a bunch of fax machines

ErrorsWindows:
Via fastdmm2

Now traitors shouldn't have to hack into the bodyguard's room to fake a fax message

Fixes the fax machine tag in the quartermaster's office tagged as "Nanotrasen Scientists" and the tagless fax machine in the senior scientist office, adds a paper bin and pen to the Chief Engineer office (so the stamp has something to be used on) and new fax machines to the 1st deck Docking checkpoint, 2nd deck Chief Engineer office, Journalist office, 3rd deck CMO office, CoS office, and Abandoned office, and the 4th deck Hangar checkpoint

🆑 

* adds: fax machines to 1st deck Docking checkpoint, 2nd deck Chief Engineer office, Journalist office, 3rd deck CMO office, CoS office, and Abandoned office, and 4th deck Hangar checkpoint
* bugfix: the fax machines in the QM and senior scientist offices have tags now

### DIFF
--- a/maps/nerva/nerva-1.dmm
+++ b/maps/nerva/nerva-1.dmm
@@ -226,8 +226,8 @@
 	dir = 6
 	},
 /obj/machinery/computer/cloning{
-	icon_state = "computer";
-	dir = 8
+	dir = 8;
+	icon_state = "computer"
 	},
 /obj/machinery/camera/network/medbay{
 	c_tag = "Cloning Bay";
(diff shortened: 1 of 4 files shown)


CPU-Blanc:
I'm personally against this PR. Fax machines **shouldn't** be in publicly accessible spaces. Heads have a fax in the meeting room they can use, they don't _all_ need one their office - This just makes it harder for them to pay attention to when/if a reply comes through.

If an antag wants to fake a fax, well, they're an antag, and breaking into a location would be easy enough.

Glloyd:
Yeah, remove the other fax machines from this, and we can merge it. Traitors should have to hack in somewhere to fake a fax message.

Issue #1361, opened 2023-10-23
airlock buttons under glass


CaraS212:
### Description of issue

![image](https://github.com/UristMcStation/UristMcStation/assets/107084118/91c08a4c-2b14-4d9b-8eae-129cee710849)
airlock buttons and such appear under glass

### Steps to reproduce

look at airlock
controls inaccessible

### Client version, server revision, & game ID

Client Version: 514
Server Revision: 459b275348172665819d5166c821f071c25dbb74 - 819d5166c821f071c25dbb74 - 2023-10-21
Game ID: cqp-duDY
Current map: ICS Nerva

### Issue bingo

- [X] Issue could be reproduced at least once
- [X] Issue could be reproduced by different players
- [X] Issue could be reproduced in multiple rounds
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] [Couldn't find an existing issue about this](https://github.com/UristMcStation/UristMcStation/issues)

Glloyd:
there's a few airlocks around the nerva broken like this, mostly the solars, but a couple others.

PR #1371, opened 2023-11-11
fixes deck 1 window airlock layers

ggermanicus:
the layer was not above the windows. now it is above the windows and you can see them and use them. yay

closes https://github.com/UristMcStation/UristMcStation/issues/1361


ggermanicus:
noooooooooooooooooooo

Glloyd:
yeah :( the only other fix would be changing the layer of all the buttons, and idk if that would cause other issues.

Issue #1348, opened 2023-10-17
Ion Storm event does not fire, and when firing, inconsistently gives ion laws


ggermanicus:
### Description of issue

RNGing or weighing RNG to trigger the event results in it either not firing or not giving ion laws. Forcing the event results in it either not firing due to 0 weight, or firing with increased weight with ion laws.

Seems to me like added job weight isn't being applied to this event, as adding AIs to the servers with players did not increase the weight. This does work with other events, so I don't know what's up.

### Difference between expected and actual behaviour

Ion storm should happen when there are roles like AI to experience them, and they should give ion laws at least some of the time.

### Steps to reproduce

0. Go in code and mod the weight as you see fit
2. Event Manager Panel
3. (try to) Fire the event either by spamming the timer down until you get it or selecting it as the next event.

### Specific information for locating

event_container.dm line 165, maybe event_dynamic.dm line 42?

### Client version, server revision, & game ID

Client Version: 514
Server Revision: 6276d967caef01480a526406638472aaccb68723 - 0a526406638472aaccb68723 - 2023-10-14
Game ID: cqj-dL0h
Current map: ICS Nerva

### Issue bingo

- [X] Issue could be reproduced at least once
- [X] Issue could be reproduced by different players
- [X] Issue could be reproduced in multiple rounds
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] [Couldn't find an existing issue about this](https://github.com/UristMcStation/UristMcStation/issues)

CPU-Blanc:
Not really a bug, but just how it's currently handled.
The Ion Storm event is of type `no_overmap`. This means that this event has an equivalent event that's triggered via the overmap, and as such is removed from the random rotation (the weight will _always_ be 0)

The event itself seems to be working as intended, just a case of if we want to change this behaviour

ggermanicus:
That's what I get for failing to read. I'd personally be in favor of getting it off overmap exclusivity - letting AIs have a chance of being ion law'd sounds fun.

If it's just a matter of removing the no_overmap bit I can do it myself later today/tomorrow.

CPU-Blanc:
Yeah, just removing the subtype would fix that. Might be something to put up to a vote?

PR #1385, opened 2023-12-18
Lets ion storm be a non-overmap event

ggermanicus:
Previously it only proc'd if you flew through the ion clouds on the overmap. Now it will proc like any random event. It's (about) as rare as bluespace drive instability currently and most ion laws are pretty trivial to reset, so it should be fine.

Closes https://github.com/UristMcStation/UristMcStation/issues/1348


CPU-Blanc (approved):
LGTM
Just a case of whether Glloyd wants this change or not

Glloyd:
Seems fine to me!

PR #1388, opened 2023-12-20
port ipc 'hair' and coloring via markings from bay

ggermanicus:
Does not apply to FBPs.
See https://github.com/Baystation12/Baystation12/pull/34304

PR #1389, opened 2023-12-20
Cooling organ

ErrorsWindows:
Adds the cooling organ for IPC in the lower body/groin
It has 90 units of liquid to work with different reagents having variable effectiveness but requires minimum 1/3rd water, by default (and best) uses 60 Coolant and 30 Water
If the organ is damaged, the liquid start to drop and then the IPC starts overheating :(
The pre-filled organ can be printed in the standard blue prosthetic organ printer, or if you don't want to constantly pull out your own guts you can add the liquid manually using a beaker

cooling_unit

___

Inspiration from me https://github.com/UristMcStation/UristMcStation/pull/1388#issuecomment-1863719903
Copied from https://github.com/Baystation12/Baystation12/pull/34304 and https://github.com/Baystation12/Baystation12/commit/932c39e8214a0527a8a3f1bebd3aa234de8ebc3b


ErrorsWindows:
This was originally going to be in the description but I decided to put it in a comment below - because to be honest after some thought, the point [Mucker brings up in the original pull request](https://github.com/Baystation12/Baystation12/pull/34304#pullrequestreview-1736190587) about this just being a nerf is a valid one: even though we more often than not a medganger manning the chems shack and we don't have the skill system (or that weird "role-bleed" rule that preceded it), this is still essentially the robot thirst meter and I don't know what people other than me will think about it, I probably won't lose any sleep if this were to get reverted after two weeks

CPU-Blanc:
I haven't looked at anything yet, but something I'll say now is _please_ use proper commit messages. "Update mob.dm" is not good enough

Glloyd:
Seconding what Blanc said. You really should be using some sort of git program rather than editing all of the files individually on github, and then committing your changes as one.

ErrorsWindows:
the more I think about it the more I think it was a stupid idea and the more I feel stupid for thinking it wasn't, not even like a thirst meter where you can enjoy a zesty soda it's more like there's no reason to not just take the meta 60 coolant

PR #1390, opened 2023-12-21
Adds Passenger role

ggermanicus:
Passengers spawn in the dorms and look like diet assistants. Make use of the loadout to stand out.
50 thaler passenger fee per hour with an alert in Command if passengers can't pay.
Be useless, yet even more free of responsibilities.


code/controllers/subsystems/payment.dm
```diff
@@ -1,3 +1,5 @@
+#define PASSENGER_FEE 50
```
Glloyd, line 1: do we want this as a hard define or a var in the subsystem? making it a var opens the possibility for it to be changed midround in the future. @CPU-Blanc has done a lot with this code though so i'd want to get their take on it too.

Glloyd:
I'm fine with the role itself, that's part of the reason I access-locked prim tool. I don't think there's anywhere else that they need to be locked from besides the hangar probably.

code/controllers/subsystems/payment.dm
CPU-Blanc: I want to have a good look through this as I've touched quite a bit of the payroll code.
Making the fee a var would be nice, as we could later hook the accounts console into it and allow the captain to up the price mid-round (for extra extortion)

maps/nerva/datums/nerva_jobs.dm
```diff
@@ -446,3 +446,20 @@
...
+	department_flag = CIV
+	total_positions = -1
+	spawn_positions = -1
+	economic_power = 2
+	outfit_type = /singleton/hierarchy/outfit/job/nerva/passenger
+
+/datum/job/passenger/equip(mob/living/carbon/human/H)
+	if (H?.mind?.assigned_role == "Passenger")
+		H.mind.manual_pay_rate = -PASSENGER_FEE
+	. = ..()
```
CPU-Blanc, line 465: This assigned role check is redundant. This also feels a _little_ jank to me, an alternative solution could be to check for the assigned role in `/datum/mind::get_pay()` and return early there.
```suggestion
/datum/job/passenger/equip(mob/living/carbon/human/H)
	. = ..()
	H.mind?.manual_pay_rate = -PASSENGER_FEE
```

CPU-Blanc (commented):
Just a couple of points, but I like the general idea.

One thing that irks me a little is it is currently its bound to the payroll payout. It'd be nice if it was in its own proc in the payment subsystem, so would show as a seperate entry in the nerva account + its own message over the command radio - eg: "x passengers paid their fairs, totalling x Th" or something. This would avoid a rare edgecase where there's just the captain online and some passengers, or a lot of passengers, as the amounts over the radio would go into the negative. The passengers accounts will also currently be paying to "ICS Nerva Automated Payroll System", which doesn't seem right.

code/controllers/subsystems/payment.dm
```diff
@@ -108,9 +121,12 @@ SUBSYSTEM_DEF(payment_controller)
...
-			else
-				message.stored_data = "[H.real_name],\n\nAs your wages have been suspended, your payout has not been processed. Please direct all complaints to your payroll administrator."
+				if(H.mind.assigned_role == "Passenger")
+					message.stored_data = "[H.real_name],\n\n<b>[passenger_fee] Th</b> has been deducted from your account as part of the hourly passenger fee.\nThank you for travelling with us.\n\n<i>This message is brought to you by [sponsor] - [this_slogan]</i>"
+				else if(!H.mind.pay_suspended)
+					message.stored_data = "[H.real_name],\n\n<b>[economic_modifier] Th</b> has been deposited into your account following hourly payroll payouts.\nThank you for your continued hard work.\n\n<i>This message is brought to you by [sponsor] - [this_slogan]</i>"
+				else
+					message.stored_data = "[H.real_name],\n\nAs your wages have been suspended, your payout has not been processed. Please direct all complaints to your payroll administrator."
```
CPU-Blanc, line 129: The final `else` statement is unreachable. You'll need to remove the `pay_suspended` check on LN123 and nudge everything an indent in.

CPU-Blanc (changes requested):

code/controllers/subsystems/payment.dm
```diff
@@ -64,15 +67,21 @@ SUBSYSTEM_DEF(payment_controller)
...
 	var/newbrand = pick(brand) //this is a dumb meme
+	penniless_passengers = list()
 	PayPeople(newbrand)
 	GLOB.global_announcer.autosay("<b>Hourly salary payments have been processed and deposited into your accounts. Thank you for your service to the [GLOB.using_map.station_name]. This message is brought to you by [newbrand]: make sure to buy [newbrand] at your nearest vending machine.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Common")
 	total_paid += moneybuffer
 	var/datum/transaction/singular/T = new(station_account, "Automated Payroll Deposits", -moneybuffer, "[GLOB.using_map.station_name] Automated Payroll System")
 	T.perform()
-	GLOB.global_announcer.autosay("<b>[moneybuffer]Th has been removed from the [GLOB.using_map.station_name]'s main account due to automated payroll services.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Command")
+	GLOB.global_announcer.autosay("<b>[max(0,moneybuffer)]Th has been removed from the [GLOB.using_map.station_name]'s main account due to automated payroll services.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Command")
+	if(!(passenger_count == length(penniless_passengers)))
```
CPU-Blanc, line 78: This is really oddly constructed. Consider this for better readibility
```suggestion
	if(passenger_count != length(penniless_passengers))
```

code/controllers/subsystems/payment.dm
```diff
@@ -85,9 +94,13 @@ SUBSYSTEM_DEF(payment_controller)
 	for(var/mob/living/carbon/H in GLOB.living_players) //we don't pay dead people
 		if(!H.mind?.assigned_role || !H.mind?.initial_account) //this excludes stowaways and other non-crew roles including derelict ships
 			continue
-
 		if(H.mind.assigned_role == "Captain" && GLOB.using_map.name == "Nerva") //todo, make a map var for captain ownership if we have other ship maps in a similar situation (never)
 			continue
+		if(H.mind.assigned_role == "Passenger")
+			passenger_count += 1
+			var/datum/transaction/U = new(H.mind.initial_account, station_account, passenger_fee, "[GLOB.using_map.station_name] Automated Fee System")
```
CPU-Blanc, line 101: I _generally_ don't like using an actual transaction here, as opposed to a singular.
This is so that on the station account, you can group all fees under a single transaction, otherwise it'll be spammed with each individual one.

If you wanted to go the singular route, you could do the passenger side here as normal, then the station account side at LN80 or so, and just do `passenger_fee * paying_passengers`

nano/templates/account_management.tmpl
```diff
@@ -167,6 +167,7 @@
 	{{:helper.link('Department Accounts', 'suitcase', {'state' : 'dep_accounts'}, data.auth != 1 ? 'disabled' : data.state == "dep_accounts" ? 'selected' : null)}}
 	{{:helper.link('System Logs', 'clipboard', {'state' : 'logs'}, data.auth != 1 ? 'disabled' : data.state == "logs" ? 'selected' : null)}}
 	{{if data.state == "crew_accounts"}}
+		<span class='floatRight'>{{:helper.link('Set Passenger Rate', 'gear', {'action' : 'passenger_rate'}, data.auth != 1 ? 'disabled' : null)}}</span>
```
CPU-Blanc, line 170: To me personally, this would make more sense on the "Payroll" screen, rather than crew accounts (as passenger fees are kinda like a reverse payroll?)

code/game/machinery/computer/account_manager.dm
```diff
@@ -429,6 +429,8 @@
 					return TOPIC_NOACTION
 				display_state = "account_overview"
 				return TOPIC_REFRESH
+			if("passenger_rate")
+				SSpayment_controller.passenger_fee = input(user, "Input desired passenger rate:", "Passenger Rate", SSpayment_controller.passenger_fee) as num
```
CPU-Blanc, line 433: Missing a refresh action on return
```suggestion
			if("passenger_rate")
				SSpayment_controller.passenger_fee = input(user, "Input desired passenger rate:", "Passenger Rate", SSpayment_controller.passenger_fee) as num
				return TOPIC_REFRESH
```

CPU-Blanc (commented):
I think this is generally all of my concerns/nit-picks.
I'd personally like a little more whitespace in the payroll `fire()` proc for better readability too

code/controllers/subsystems/payment.dm
ggermanicus: I don't think there will ever be enough passengers or transactions to result in true spam... also every attempt I've made of this today resulted either in runtimes, bad transactions, or bad transaction origin names (/datum/station_account instead of "Automated..."

CPU-Blanc: That's because it's expecting a string, not an account datum. You can look at what the payroll does as an example.

ggermanicus: Getting the origin to display properly throws the runtimes as it starts trying to check .suspended on every payday and failing

CPU-Blanc: That's because the `singluar` subtype is very weird. The first arguement is _always_ the real account, regardless of the direction the funds are moving in. The direction (debit or credit) is effected by the amount being positive or not.

nano/templates/account_management.tmpl
```diff
@@ -172,6 +172,7 @@
 	<br>
 	<hr>
 	{{if data.state == "payroll"}}
+		<span class='floatRight'>{{:helper.link('Set Passenger Rate', 'gear', {'action' : 'passenger_rate'}, data.auth != 1 ? 'disabled' : null)}}</span>
```
CPU-Blanc, line 175: This looks very odd here. Put this up after LN171 with an if gate for the "payroll" state (so it is before the line break and header

code/game/machinery/computer/account_manager.dm
CPU-Blanc, line 1: You should consider adding a check for the passenger roll within the payroll data packet, to exclude them. This will stop them appearing under payroll. ie, look at LN138 & LN143 for reference, with excluding the captain and scientists.

code/controllers/subsystems/payment.dm
```diff
@@ -64,15 +67,25 @@ SUBSYSTEM_DEF(payment_controller)
...
+	var/datum/transaction/singular/U = new(station_account, "Automated Fee Deposits", (passenger_fee * paying_passengers), "[GLOB.using_map.station_name] Automated Fee System")
 	var/datum/transaction/singular/T = new(station_account, "Automated Payroll Deposits", -moneybuffer, "[GLOB.using_map.station_name] Automated Payroll System")
+	U.perform()
 	T.perform()
-	GLOB.global_announcer.autosay("<b>[moneybuffer]Th has been removed from the [GLOB.using_map.station_name]'s main account due to automated payroll services.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Command")
+	GLOB.global_announcer.autosay("<b>[max(0,moneybuffer)]Th has been removed from the [GLOB.using_map.station_name]'s main account due to automated payroll services.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Command")
+
+	if(passenger_count != length(penniless_passengers))
+		GLOB.global_announcer.autosay("<b>[paying_passengers] passenger[paying_passengers > 1 ? "s have" : " has"] paid [(passenger_fee*paying_passengers)]Th to the [GLOB.using_map.station_name]'s main account.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Command")
```
CPU-Blanc, line 83: It'd be nice if we could gate these behind checks if they actually happen.
ie
```
if(paying_passengers)
	var/datum/transaction/singular/U = new(station_account, "Automated Fee Deposits", (passenger_fee * paying_passengers), "[GLOB.using_map.station_name] Automated Fee System")
	U.perform()
	GLOB.global_announcer.autosay("<b>[paying_passengers] passenger[paying_passengers > 1 ? "s have" : " has"] paid [(passenger_fee*paying_passengers)]Th to the [GLOB.using_map.station_name]'s main account.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Command")

```

code/controllers/subsystems/payment.dm
```diff
@@ -64,15 +67,27 @@ SUBSYSTEM_DEF(payment_controller)
...
 	PayPeople(newbrand)
 	GLOB.global_announcer.autosay("<b>Hourly salary payments have been processed and deposited into your accounts. Thank you for your service to the [GLOB.using_map.station_name]. This message is brought to you by [newbrand]: make sure to buy [newbrand] at your nearest vending machine.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Common")
 	total_paid += moneybuffer
+
+	var/paying_passengers = passenger_count - length(penniless_passengers)
+	if(paying_passengers)
+		var/datum/transaction/singular/U = new(station_account, "Automated Fee Deposits", (passenger_fee * paying_passengers), "[GLOB.using_map.station_name] Automated Fee System")
+		U.perform()
+		if(passenger_count != length(penniless_passengers))
+			GLOB.global_announcer.autosay("<b>[paying_passengers] passenger[paying_passengers > 1 ? "s have" : " has"] paid [(passenger_fee*paying_passengers)]Th to the [GLOB.using_map.station_name]'s main account.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Command")
```
CPU-Blanc, line 81: This is essentially the exact same check as on LN77, so is redundant
```suggestion
		GLOB.global_announcer.autosay("<b>[paying_passengers] passenger[paying_passengers > 1 ? "s have" : " has"] paid [(passenger_fee*paying_passengers)]Th to the [GLOB.using_map.station_name]'s main account.</b>", "[GLOB.using_map.station_name] Automated Payroll System", "Command")
```

code/game/machinery/computer/account_manager.dm
```diff
@@ -138,6 +138,9 @@
 				if(H.mind.assigned_role == "Captain" && GLOB.using_map.name == "Nerva")
 					continue	//We don't pay the captain
 
+				if(H.mind.assigned_role == "Passenger" && GLOB.using_map.name == "Nerva")
```
CPU-Blanc, line 141: Passenger job is Nerva only, so this map name check is redundant. I also can't envision passengers needing to be on the payroll for _any_ map
```suggestion
				if(H.mind.assigned_role == "Passenger")
```

CPU-Blanc (changes requested):
This should be the last of it. Everything else looks good :+1:

CPU-Blanc (approved):

CPU-Blanc:
LGTM :+1: 
Will likely need a squash before merge

CPU-Blanc:
Unstalling until Glloyd returns

Glloyd:
Looks good to me!

PR #1391, opened 2023-12-21
Add Small Ram Horns as New Biomod

The32bitguy:
- Adds a new biomod to the character creator for Small Ram Horns from [f13babylon](https://github.com/f13babylon/f13babylon/blob/main/icons/mob/mutant_bodyparts.dmi)

image (7)


CPU-Blanc:
Just to clear up, are these sprites yours, or cherry picked from somewhere else?

The32bitguy:
> Just to clear up, are these sprites yours, or cherry picked from somewhere else?

The icon is from f13babylon/f13babylon

CPU-Blanc:
Then you needed to credit the source in this PR, as part of their license

code/modules/sprite_accessories/accessory_human.dm
CPU-Blanc, line 1: This _could_ probably be moved into an Urist-specific file to avoid merge conflicts with upstream.
Probably something like `code/modules/urist/modules/sprite_accessories/accessory_human.dm`

Up to @Glloyd how he wants to handle imports from different codebases.

Glloyd: would prefer it being moved into a urist file, yeah

The32bitguy: A Urist specific .dm file?

.gitignore
CPU-Blanc, line 1: This shouldn't be in the gitignore, and should be reverted.

icons/mob/human_races/species/human/bodymods.dmi
CPU-Blanc, line 1: Looks like the same icons are in both here, and the urist-specific dmi. You'll need to remove it from here

code/modules/sprite_accessories/accessory_human.dm
```diff
@@ -82,17 +82,3 @@
 	icon_state = "horns_frills_long"
 	name = "Horn Biomods (Long Frills)"
 
-
-/datum/sprite_accessory/marking/human/horns/frills_short
-	icon_state = "horns_frills_short"
-	name = "Horn Biomods (Short Frills)"
```
CPU-Blanc, line 88: Frills has been deleted. This should be restored.

CPU-Blanc (changes requested):

code/modules/sprite_accessories/accessory_human.dm
The32bitguy: This is getting embarrassing for me. Will do

CPU-Blanc: Nothing to fret over. Happens to all of us :+1:

PR #1396, opened 2023-12-25
Ports snow and snow recipes from Polaris

ggermanicus:
Merr crismas.

- You can gather snow with your hands from snow tiles.
- You can make snowballs, snowmen, snowborgs, and snowspiders.
- Pack snowballs with your hands to be denser and do a little damage.

image


Glloyd:
You'll also need to add this to `/turf/simulated/floor/planet/snow` so people can use the snow on Cara's awaymap

Glloyd:
lgtm

PR #1404, opened 2024-02-12
Noisemaker fixes

Yantarii:
**Honk**

- Removes two missing sounds from noisemaker.
- Fixes missing capitals for some sound files.
- Changes description a little


CPU-Blanc (approved):
LGTM.
I'd love to at a later date go in and rename all the source files to be more consistent with their style... Why are we mixing snake case, pascal case and god knows what else lmao 
Though that'd just cause an absolute headache upstream with future merges..

PR #1410, opened 2024-03-10
Exoplanet ruin, The Container Yard

ErrorsWindows:
My First Map™️ 

An exoplanet ruin depicting a warehouse and some shipping containers: tiny and compact, just like [the idea it came from](https://callofduty.fandom.com/wiki/Shipment), connected to an airstrip with a shuttle that can be repaired to become overmap-capable

As I still don't actually know how to make a map, mostly from the [Transshipment Point](https://github.com/Baystation12/Baystation12/pull/33700) (but only a 90% copy - the mapping tool wasn't completely compatible so I reinterpreted a couple of the blank spots to make it more Urist)

🆑 WindowsErrors
map: The Container Yard, an exoplanet ruin
/ 🆑


maps/random_ruins/exoplanet_ruins/container_yard/container_yard.dm
CPU-Blanc, line 1: This file needs to be included in `maps/_maps.dm`, otherwise it's never actually getting compiled.

CPU-Blanc (changes requested):

CPU-Blanc:
I'm also unsure if we actually want to merge this. As appears to be mostly a port from upstream with a few minor changes, once the next baymerge comes we'll essentially have 2 of the same ruins. @Glloyd thoughts?

ErrorsWindows:
yea you're probably right this was mostly just me trying to get a mapping client to work in practice

PR #1417, opened 2024-03-19
Streamlining: mostly engineering-facing

ggermanicus:
- Pull speed has been significantly increased. You will still notice it pulling canisters/lockers/etc, but not as badly.
- <s>Several deconstruction steps with rwalls have had their timers removed or reduced.</s>
- Several construction recipes have had their timers reduced.
- Captains's locker -> captain's locker

### DIFF
--- a/maps/nerva/obj/nerva_closets.dm
+++ b/maps/nerva/obj/nerva_closets.dm
@@ -15,7 +15,7 @@
 	)
 
 /obj/structure/closet/secure_closet/nervacap
-	name = "captains's locker"
+	name = "captain's locker"
 	req_access = list(access_captain)
 	closet_appearance = /singleton/closet_appearance/secure_closet/command/nervacap
 
--- a/code/game/objects/structures/girders.dm
+++ b/code/game/objects/structures/girders.dm
@@ -88,10 +88,8 @@
 	if(isScrewdriver(W))
 		if(state == 2)
 			playsound(src.loc, 'sound/items/Screwdriver.ogg', 100, 1)
-			to_chat(user, SPAN_NOTICE("Now unsecuring support struts..."))
-			if(do_after(user, 4 SECONDS, src, DO_REPAIR_CONSTRUCT))
-				to_chat(user, SPAN_NOTICE("You unsecured the support struts!"))
-				state = 1
+			to_chat(user, SPAN_NOTICE("You unsecured the support struts!"))
+			state = 1
 		else if(anchored && !reinf_material)
 			playsound(src.loc, 'sound/items/Screwdriver.ogg', 100, 1)
 			reinforcing = !reinforcing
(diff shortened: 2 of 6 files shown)


scrdest (changes requested):
tl;dr keep the timers, just make them super-short to make merges less painful (and potentially, avoid weird bugs because something relies on a do_after being there at all)

maps/nerva/obj/nerva_closets.dm
```diff
@@ -15,7 +15,7 @@
 	)
 
 /obj/structure/closet/secure_closet/nervacap
-	name = "captains's locker"
```
scrdest, line 18: the captainses, my preciousssss

scrdest (approved):
LGTM

CPU-Blanc (commented):
I get wanting to remove a lot of tedium, however to me this feels a bit too far, and seriously nerfs rwalls and their ability to well.. secure areas and keep people out.
More time to deconstruct = more time to get caught, and as most of the time this'll be from maintenance, it's already rather hard to spot them as is.

ggermanicus:
Anecdotally speaking noise is the primary detector of crime, not walking into it. Noise might happen faster but it will still happen.

Area security is already a joke; thermite is faster than any form of rwall decon before and after this PR, you can smash lockers open with a fire extinguisher, you can smash open most doors, every safe in the world is openable with a screwdriver and a multitool, and I think you can probably even smash an rwall with an axe if you really want to. Combined with this being pretty much how it was before baymerge, I don't really see it as a nerf more than bringing things back to how it ought to be.

CPU-Blanc:
I disagree; just because other methods exist doesn't mean we should nerf this entrypoint. And the longer someone spends breaking down a wall, the higher the odds that someone will be lingering around nearby to hear it.

I really don't see the need for this. I don't think people are _that_ frequently deconstructing rwalls to the point where the timers are just a hindrance on game flow, it just feels like a nerf for the sake of a nerf. The other areas I agree with; stuff like surgery, dragging, some other construction steps etc _is_ just artificial bloat, but rwalls taking time _does_ serve a purpose.

ggermanicus:
And it still does take time. Not all timer steps are removed and you have to fumble through your inventory to complete all the steps anyways.

I don't see why something that is done so infrequently (probably in part because of the comparative difficulty vs other options) should remain considerably harder, rather than nerfed to being a more viable option in line with all the other options. Even if this change went through, when is breaking an rwall down ever more efficient in time or detection than grabbing a multitool and going through the doors? Anyone who will see you opening an airlock is going to hear you ripping a wall apart for a minute.

scrdest:
Big brain time: as far as I can tell, most if not all of these delays will be calculated at runtime...

We could add a (bunch of) config value(s) for multipliers. Default them to 1, instead of changing Bay values just slap a 'times urist_construction_multiplier' in there.

This means we would be able to a) tune this much more easily (no serb update required), b) tune this at runtime, by VVing the config.

CPU-Blanc:
> And it still does take time. Not all timer steps are removed and you have to fumble through your inventory to complete all the steps anyways.
> 
> I don't see why something that is done so infrequently (probably in part because of the comparative difficulty vs other options) should remain considerably harder, rather than nerfed to being a more viable option in line with all the other options. Even if this change went through, when is breaking an rwall down ever more efficient in time or detection than grabbing a multitool and going through the doors? Anyone who will see you opening an airlock is going to hear you ripping a wall apart for a minute.

Yes it still takes time, but quite a substantial less time than before. Again, I do not see the need for this nerf. How frequently are people deconstructing rwalls where it warrants these timers being shortened? Again, it just feels like a nerf for the sake of it; the only real-world effect being making areas easier to break into. That's the part I really do not see. I'll also argue that hacking the door requires gloves (and can take longer if the airlock is bolted+without power etc), thermite requires chem access, bashing the airlock down makes a _lot_ of noise and takes quite a while. Tools are super easy to come by, the rest at least take a few extra steps

CPU-Blanc:
> Big brain time: as far as I can tell, most if not all of these delays will be calculated at runtime...
> 
> We could add a (bunch of) config value(s) for multipliers. Default them to 1, instead of changing Bay values just slap a 'times urist_construction_multiplier' in there.
> 
> This means we would be able to a) tune this much more easily (no serb update required), b) tune this at runtime, by VVing the config.

That'd be a nice idea for the future. Would also be nice for any downstream servers to adjust as they see fit. It'd still argue rwalls should be a separate entry, and not grouped up with general construction :P

CPU-Blanc (approved):

PR #1418, opened 2024-03-19
Streamlining: medical-facing changes

ggermanicus:
- Surgery steps are now about as fast as a master surgeon would be.
- Vitals monitor now (finally) replaces the operating computers in medbay and robotics.
- Vitals monitor adjusted so we aren't missing anything from the old operating computers with the change.
- Basically you have an operating computer but it looks better and beeps now. (You can turn the beeps off.)
- Health scanner informs you how/if your patient is breathing.


scrdest (approved):
LGTM

CPU-Blanc (approved):

Issue #1415, opened 2024-03-14
assistants don't have access to the general storage doors - despite literally spawning there


ErrorsWindows:
### Description of issue

title

very funny

### Steps to reproduce

1. spawn as assistant
2. in the tool storage
3. be stuck in the tool storage because you don't even have access to the doors
4. be forced to smash windows and live up to your name
5. greyshirts are unironically the most oppressed social group

### Client version, server revision, & game ID

Client Version: 515
Server Revision: [cc40971bd063df88431434211de6e64cfd493b33](https://github.com/UristMcStation/UristMcStation/commit/cc40971bd063df88431434211de6e64cfd493b33) - 431434211de6e64cfd493b33 - 2024-02-17
Game ID: csG-awYT
Current map: ICS Nerva

### Issue bingo

- [X] Issue could be reproduced at least once
- [ ] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] [Couldn't find an existing issue about this](https://github.com/UristMcStation/UristMcStation/issues)

CPU-Blanc:
Are assistants meant to lack access now @Glloyd ? Feels like an oversight.

PR #1419, opened 2024-03-19
Moves Assistant Spawns around Deck 3.

Yantarii:
Assistant no longer spawn in Tool Storage with no way out other than a bit of casual B&E.

Assistants will spawn around the top level bar, the lower observatory, the dormitory, arcade cabinet.

Fixes - 
https://github.com/UristMcStation/UristMcStation/issues/1415


CPU-Blanc:
The lack of access to tool storage is an oversight with the assistants' IDs and not intended, so this mapping fix is not required.

PR #1420, opened 2024-03-19
ports fix for escaping lockers

ggermanicus:
previously people welded in lockers could never free themselves, which is bad


ErrorsWindows:
you can free yourself by hitting the locker while you're inside of it but you need to have weapons with you to begin with

CPU-Blanc (approved):

PR #1421, opened 2024-03-19
Duffles & workboots & QM sunglasses & punctuation

ggermanicus:
- Our leather workboots are in fact overpowered. They have been demoted to supply pack items.
- QM gains sunglasses as is tradition. Supply tech shoes replaced with bay workboots.
- Urist dufflebags with bay equivalents have been replaced with bay dufflebags. Dufflebags are now accessible on the back, carry more, but slow you down. Slowdown subject to change.
- Unnecessary punctuation in the supervisors list removed.


CPU-Blanc:
Token comment so the bot doesn't stale this. I'll take a look this week

Glloyd:
Seems fine to me, I fixed the leather craft so they can also be gotten through that method again.

Glloyd:
We can gauge thoughts on the slowdown, I don't really have a strong opinion in one way or the other.

PR #1428, opened 2024-04-06
Adds Napalk

Yantarii:
More pilk crimes against humanity, combining napalm and pilk to make napalk, very tasty and good for you.
Just something fun while I work on some other PRs

### DIFF
--- a/code/modules/urist/datums/uristchem.dm
+++ b/code/modules/urist/datums/uristchem.dm
@@ -21,3 +21,26 @@
 /datum/reagent/drink/pilk/affect_ingest(mob/living/carbon/M, var/alien, var/removed)
 	..()
 	M.adjustBrainLoss(0.1)
+
+/datum/chemical_reaction/napalk
+	name = "Napalk"
+	result = /datum/reagent/drink/napalk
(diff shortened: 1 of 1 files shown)


CPU-Blanc (approved):
Actually cursed

code/modules/urist/datums/uristchem.dm
```diff
@@ -21,3 +21,26 @@
...
+	color = "#76685f"
+
+	glass_name = "Napalk"
+	glass_desc = "A glass of napalk, a combination of napalm and pilk. What compells you to drink this?"
+	glass_special = list(DRINK_VAPOR)
+
+/datum/reagent/drink/napalk/affect_ingest(mob/living/carbon/M, var/alien, var/removed)
+	..()
+	M.adjustToxLoss(0.5)
+	M.adjustBrainLoss(0.2)
```
scrdest, line 46: Brainloss is kind of a weird effect here. Why not make this work like a diet version of Unholy Water that doesn't have a pass-through on cultists?

Glloyd:
It's inherited from pilk which also gives brainless for being dumb enough
to drink it.

PR #1433, opened 2024-04-06
Adds Jackboots back to Sec & CoS lockers.

Yantarii:
CoS & Security Lockers now contain jackboots.


CPU-Blanc (changes requested):
The lockers are too full. They're failing the test where they are opened and then closed again to make sure everything fits inside. You'll need to remove some items

PR #1434, opened 2024-04-06
Adds additional Holomaps & a wheelchair to Nerva

Yantarii:
This adds additional Holomaps all around the Nerva, usually near key and memorable areas, or specific hallways.
The intent is to make it easier for new players to look around the map aside from viewing that one pinned PNG

**Locations:**

**Deck 1:**
Bottom of Trajan Hangar

**Deck 2:**
Cryo Storage
Bar
Near Security Lobby
Holodeck

**Deck 3:**

Near Deck 3 Stairway Elevator.
Dormitory
Observatory
Near Teleporter Outside.

**Deck 4:**
Top Docking Port.

This also adds a singular wheelchair in the Medical Storage Bay.

**TODO:**

- [ ] Check all dir are working correctly.

### DIFF
--- a/maps/nerva/nerva-1.dmm
+++ b/maps/nerva/nerva-1.dmm
@@ -9919,6 +9919,9 @@
 	icon_state = "intact-supply"
 	},
 /obj/effect/catwalk_plated/dark,
+/obj/machinery/ship_map{
+	dir = 6
+	},
 /turf/simulated/floor/plating,
 /area/logistics/hangar)
 "rr" = (
@@ -17164,6 +17167,9 @@
 /obj/effect/floor_decal/corner/yellow{
 	dir = 9
 	},
+/obj/machinery/ship_map{
+	dir = 4
+	},
 /turf/simulated/floor/tiled/dark,
 /area/logistics/lowercargo)
 "DY" = (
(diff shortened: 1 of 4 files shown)


CPU-Blanc (changes requested):
This branch is out of date and regresses previous map changes, notably ones made in #1418 . You'll need to rebase onto baymerge-testing

maps/nerva/nerva-1.dmm
CPU-Blanc, line 1: The map in the hanger is the wrong direction and in the middle of the floor

CPU-Blanc (changes requested):

maps/nerva/nerva-3.dmm
CPU-Blanc, line 1: - Holomap outside the elevator is literally like 5 steps from another one, do we need them this close?
- Holomap in the dorms overlaps the camera and looks odd
- Holomap in the starboard observatory overlaps a light and looks odd

CPU-Blanc (commented):
Like the idea, however, I have a couple of issues/points.

- Some of the maps' overlays appear to spawn on the wrong tile, and kinda just float in mid-air. This looks odd and would ideally be fixed
- The placement sometimes feels a little random? ie the one outside the teleporter and inside cargo. Is there a way to make them a little more uniform? ie outside stairs, evenly spaced in public hallways in the same tough position on each floor?

maps/nerva/nerva-3.dmm
Yantarii: I'll do a pass tonight and make some changes, need to fix the weird dir issues as well

Glloyd: yeah, they definitely don't need to be that close, especially because they do have a slight performance impact while in use iirc

PR #1436, opened 2024-04-07
take away AI drone access to all radio

ggermanicus:
Why: Being drone should have more downsides. AI too good. Also because it was suggested and is very easy.


CPU-Blanc:
Honestly, I don't mind either way if this is added or not. I _personally_ would like to see changes like this put to a vote to see how the general playerbase feels

CPU-Blanc (approved):

PR #1438, opened 2024-04-08
[DRAFT] Adds a Flashlight Program to the PDA

Laniakeaaa:
**Work in Progress**

- Adds a program that will be default with all PDAs to toggle a flashlight (with a small light radius)
Usable for PDA and tablets.
Low download cost, as it's just a flashlight, nothing fancy.


CPU-Blanc:
If it's a WIP, could you convert it to a draft, then mark it ready for review when it's done?

PR #1439, opened 2024-04-10
tweak: Grants Coder roles FUN/POSESS/REJUV

scrdest:
By and large, the Coder perms already allow the same things as these three perms, except through more cumbersome workarounds like SDQL2 or proc-calls.

As of today, I am a major source of adminbus anyway - being able to use buttons rather than dark SDQL2 incantations would make my life a hell of a lot more pleasant.

I didn't think it's worth a new rank since, again, Coder role in general can effectively do these anyway, and it allows for easier testing of Fun verbs.


CPU-Blanc (approved):
I don't see why not. Any coder with half a mind could do the same things with SDQL + VV and AdvancedProcCall anyway. Unless Glloyd has any reason against it

Glloyd:
Yeah, I honestly thought coders had +FUN already anyways.

PR #1442, opened 2024-04-11
Adds Various Legacy Items to the Loadout Menu

The32bitguy:
Draft PR for now


scrdest:
Did you check the onmobs for these? ISTR some of them might be borked.

The32bitguy:
> Did you check the onmobs for these? ISTR some of them might be borked.

I am going to go through the codebase and make commits adding old items, before finally checking all of them to see if they function, if descriptions and names need to be made more generic and then have a maintainer judge some of the "vibe" issues like role restrictions and if the spritework is up to standard because some of the icons used are very old.

PR #1443, opened 2024-04-12
Makes all loadout items without a defined cost 0 points to buy

ErrorsWindows:
The implementation of [this deliberation](https://discord.com/channels/145674110075273216/168412517851201536/1228026477245628518)

Overall consensus - Items without a significant mechanical impact cost 1 point or are otherwise job restricted to jobs that get given equivalents for free that look different (read: those 1 points stacks up quickly)

Also cleaned up a misleading description for the gear harness - it says it fits body prosthetics when in reality it is just race-restricted to IPCs and dooners, misleading when I tried to make a human FBP character only to be blocked from the item

🆑 
* tweak: makes all loadout items without a defined cost into the public domain - 0 points to buy
* grammar: cleans up a misleading description for the gear harness
/ 🆑

### DIFF
--- a/code/modules/client/preference_setup/loadout/lists/xenowear.dm
+++ b/code/modules/client/preference_setup/loadout/lists/xenowear.dm
@@ -101,7 +101,7 @@
 
 // Misc clothing
 /datum/gear/uniform/harness
-	display_name = "gear harness (Full Body Prosthetic, Diona)"
+	display_name = "gear harness (IPC, Diona)"
 	path = /obj/item/clothing/under/harness
 	slot = slot_w_uniform
 	sort_category = "Xenowear"
--- a/code/modules/client/preference_setup/loadout/loadout.dm
+++ b/code/modules/client/preference_setup/loadout/loadout.dm
@@ -288,7 +288,7 @@ var/global/list/gear_datums = list()
 	var/display_name       //Name/index. Must be unique.
 	var/description        //Description of this gear. If left blank will default to the description of the pathed item.
 	var/path               //Path to item.
-	var/cost = 0           //Number of points used. Items in general cost 1 point, storage/armor/gloves/special use costs 2 points.
+	var/cost = 0           //Number of points used. Items in general cost 1 point, storage/armor/gloves/special use costs 2 points. Apr. 2024 change to 0 points for general items
 	var/slot               //Slot to equip to.
 	var/list/allowed_roles //Roles that can spawn with this item.
 	var/list/allowed_branches //Service branches that can spawn with it.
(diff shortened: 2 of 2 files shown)


CPU-Blanc:
I must say I'm personally against this change. It doesn't really make much sense.
There's already 12 points in the loadout which seems to be _plenty_ to make what you want, without going completely overboard and rendering buying in-game utterly useless. I use the loadout quite extensively and have never gone over 12.

ErrorsWindows:
I don't think the counter argument of rendering in-game purchases useless is very strong when what is being affected is a significant amount of flavor and fluff that has negligible significance outside of personalization, such as accessories or jewelry, and none of the specific item costs are affected such as the medal selection which costs a whole 8 points, also considered in that a good amount of loadout items are variant items that are not even available to be bought on the Nerva (sorry guys I couldn't bring my union identification card with me, I had to leave it home to make room for my iPad)
(on a tangent I should probably allow customization so that it's not just three medals, there are some pretty high and mighty medallions in the torch-bay loadout files)
(on a side tangent I wonder if a purple or the big red would ever do the Torch map event, that would probably be unique and put some butts in the seats)

The32bitguy:
> Lemme just load out like 5 tablets/laptops and sell/boost research with them.

This particular example is possible regardless of this PR.

> This is also the wrong way to go about this, I feel. If points are an issue, which again I strongly contest, the solution shouldn't be allowing to essentially unlimited spawn items, but a sensible raising of the loadout points available to spend.

I suggest a compromise of 0.5 points instead of 0.

CPU-Blanc:
Again, the current points is more than enough. There's 12 points, and most fluff costs 1 point. That's 12 fluff items; _more_ than reasonable.

I have a character which every clothing item is a loadout item, has 3 accessories, and an advanced laptop computer, all within budget.

I'll refrain from commenting further as I've already laid out my case and am just going in circles, but I really don't see the necessity, and at best think a raising of the cap would be a more suitable adjustment _if that is even necessary_

scrdest:
Having a cost of zero is a bit dodgy from a security perspective; with no limit whatsoever, someone malicious could pick an absurd amount of 0-cost loadout items and crash the server.

Glloyd:
I have to agree with Scr on this, it potentially opens up the system for
abuse in a lot of ways. What happens to items that you can't carry, do they
just drop on the ground? Because beyond the potential for this to create
lag/client crashing scenarios/etc. it could also be used to abuse the cargo
system. Spawn in cargo with 1000 of whatever the highest value item is,
throw it in crates and profit.

Yantarii:
Can't say I really agree with this change, 12 points are plenty enough to buy almost everything you'd want uniform and fluff wise to bring with you, bar items with mechanical influence. 

Something that could be a good comprimise is updating our urist clothing vendors with some of the items found in the loadout, so you could buy them in-game at the round start. 

If you've already gotten a fair few items, say six, or seven then you'd probably have enough thalers round start to buy the last piece, i.e a jacket or something else, at the same time I'm not against it being raised to 15 points, but I think anymore is too much.

Glloyd:
that's true, those old clothing vendors haven't been updated in eons, and it might be a good place to move some fluff items to regardless of what ends up with loadouts.

PR #1447, opened 2024-04-22
Flashlight  Subtypes (Flares, Glowsticks,  Lanterns, Penlights, Flashdarks, Lamps) now have action buttons.

Yantarii:
image

Emergency Flares and Glowsticks now have action buttons to ignite/snap them.
Pressing them again will just inform the user in chat that it's already ignited/in use.

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -251,7 +251,7 @@
 	light_color = "#e58775"
 	icon_state = "flare"
 	item_state = "flare"
-	action_button_name = null //just pull it manually, neckbeard.
+	action_button_name = "Ignite Flare"
 	var/fuel = 0
 	var/on_damage = 7
 	var/produce_heat = 1500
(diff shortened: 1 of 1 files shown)


Yantarii:
Lanterns, Penlights, Flashdarks and Lamps now have their own specific action button text.

CPU-Blanc:
I don't really see why the glowstick & flares would need a dedicated action button, given they are single use?

ErrorsWindows:
you squeeze the pocket that has the glowstick in it with your hand to crackle it instead of having to take it out

Yantarii:
Going to adjust flares and glowsticks to have a button only when unlit to activate them.

I would prefer to keep subtypes of objects/devices consistent UI wise, I think single-use button prompts aren't going to cause much of an issue screen-estate wise.

Glloyd:
> Going to adjust flares and glowsticks to have a button only when unlit to activate them.
> 
> I would prefer to keep subtypes of objects/devices consistent UI wise, I think single-use button prompts aren't going to cause much of an issue screen-estate wise.

If you made this change I'd be fine merging this

PR #1463, opened 2024-05-07
Ports radio jammers for AI/borg signals, signalers, and headsets

ggermanicus:
70% auroracode, 25% baycode, 5% rewriting so I don't have to engage with tmpls

Jammers can be toggled on(all)/off, or set to block nothing, to block non-adjacent AI/borg actions on machinery, or to block that and radio transmission/reception. Victims will be alerted that something is blocking them from transmitting/remote operation, but not that they aren't hearing their radios anymore.

Uplink jammer:
- 7 tile range
- Lasts forever
- 32tc (for now)

Improvised jammer:
- 5 tile range
- 25 seconds of charge on a standard cell
- 2 signallers and a power cell to make


Glloyd:
The TC value will probably be adjusted, and I'm sure there's some edge cases where we'll need to add ai_can_interact checks to other objects, but as is this looks okay to me. 👍

PR #1465, opened 2024-05-11
Traitor sleepy ring

ErrorsWindows:
This thing's just been [sitting in the code since 2017](https://github.com/UristMcStation/UristMcStation/blob/baymerge-testing/code/modules/clothing/rings/rings.dm#L52). I think Baystation forgot about it. Apparently it's also in Aurora station? I don't know what they use it for either

13 cool kid points
When it's worn, it immediately injects its 15u of chloral hydrate into the wearer
Like the paralysis pen, it counts as an open container so it can be refilled
Do not propose ring to robots


CPU-Blanc (approved):
I personally can't see this really being used effectively as no-one on this server really uses jewellery, but hey, an option is an option

PR #1466, opened 2024-05-11
Yeosa'unathi - only 6 years late lmao

ErrorsWindows:
[Idea chain](https://discord.com/channels/145674110075273216/145674110075273216/1238661066595110944)
[second](https://discord.com/channels/145674110075273216/625806040595955713/625815161672040458)

These have just been sitting in the code since 2018, I guess it was supposed to be an ancient species vote from before my time? It's not like it's mule where they get 50% psionic chance, and it's literally a lizard_subspecies™️ so somebody can tell me if it's contentious
The only real powergaming concern I can see is probably the venom bite? Except it has a delay of like 6 whole ass seconds so you could probably roleplay being allergic to yeosa venom while they use the change attack verb

### DIFF
--- a/code/modules/species/station/lizard_subspecies.dm
+++ b/code/modules/species/station/lizard_subspecies.dm
@@ -17,7 +17,7 @@
 	their culture has diverged majorly from the Sinta, spending less time performing acts of violence and more time socializing. \
 	Their biology is heavily attuned to surviving Moghes' dangerous waters, including gills, fins, and a venomous bite."
 
-	spawn_flags = SPECIES_IS_WHITELISTED
+	spawn_flags = SPECIES_IS_WHITELISTED | SPECIES_CAN_JOIN
 
 /*	base_auras = list(
 		/obj/aura/regenerating/human/unathi/yeosa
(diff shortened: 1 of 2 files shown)


ErrorsWindows:
oh whoops i forgot to add SPECIES_CAN_JOIN so somebody can actually create a character as yeosa

only 3 and a half months late

ggermanicus (changes requested):
One major issue: captains can be Yeosa. Add /datum/species/unathi/yeosa to species_to_job_blacklist in maps/nerva/nerva_define.dm to fix this, if you're still interested. Otherwise this should work.

ErrorsWindows:
> One major issue: captains can be Yeosa. Add /datum/species/unathi/yeosa to species_to_job_blacklist in maps/nerva/nerva_define.dm to fix this, if you're still interested. Otherwise this should work.

only two and a half weeks late because I turned off notifications but ok

I would have thunk that it would already count because it's an unathi subspecies like mules and tritonians instead of its own thing like IPCs vs. body prosthetics but just to be sure it's done

PR #1473, opened 2024-06-23
Bug fixes

### LINKED ISSUE #1352: Condimaster does not make condiment bottles
### Description of issue

Title, a condimaster makes the same bottles as a chem master

_No response_

### Steps to reproduce

put beaker in condimaster
transfer reagents, press make condiment bottle,
get normal bottle and not a condiment bottle

should produce 
.../reagent_containers/food/condiment

instead produces
reagent_containers/glass/bottle

the most relevant bit of code i could find,

/obj/machinery/chem_master/condimaster/get_chem_info(datum/reagent/reagent)
    return ..(reagent, "Condiment Info", 0)

/obj/machinery/chem_master/condimaster/create_bottle(mob/user)
    var/obj/item/reagent_containers/food/condiment/P = new/obj/item/reagent_containers/food/condiment(loc)
    reagents.trans_to_obj(P, 50)

doesnt seem to me to have anything wrong.

Client Version: 514
Server Revision: 6276d967caef01480a526406638472aaccb68723 - 0a526406638472aaccb68723 - 2023-10-14
Game ID: cql-aCi4

### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/UristMcStation/UristMcStation/issues)

### LINKED ISSUE #1368: Cannot unban job bans from the job ban panel
### Description of issue

You cannot unban players anymore from the job ban panel accessed via the player panel. You must instead do it from the "Display Job bans" verb which lists ALL job bans held in the database, and do it individually for each role... Eugh

Clicking a role in the job ban panel should ban the ckey from that job. Clicking a banned role from the same panel should unban as it did before

### Steps to reproduce

1. Job ban a client from the PP Job ban panel
2. Observe not being able to unban them from said panel

I'll probably tackle this myself, but it's likely in the job ban panel html/href going wonk

Client Version: 514
Server Revision: [25a56a52bccbc1b501c1ca6db9a1d383bd50c0ee](https://github.com/UristMcStation/UristMcStation/commit/25a56a52bccbc1b501c1ca6db9a1d383bd50c0ee) - 01c1ca6db9a1d383bd50c0ee - 2023-10-26
Game ID: cqB-dCC1
Current map: ICS Nerva

### Issue bingo

### LINKED ISSUE #1381: PDA Crew Manifest shows null rank
### Description of issue

image

PDA Manifest shows both posistion and rank, instead of just having one of them.

Have one crew role visible.
Has two (one null)

### Steps to reproduce

1. Look at the crew manifest.

PDA Crew Manifest

Client Version:  514 
Server Revision:  [261f0cb64e444823da5166fb09b1ee9a9165146a ](https://github.com/UristMcStation/UristMcStation/commit/261f0cb64e444823da5166fb09b1ee9a9165146a) - da5166fb09b1ee9a9165146a - 2023-11-26 
Game ID: cra-dCS1 
Current map: ICS Nerva

### Issue bingo

### LINKED ISSUE #1411: cant graffiti
### Description of issue

normally when you use a sharp object like a screwdriver or a glass shard or a knife or something on a floor or wall it allows you to write a persistent message but I can't
(I have tried using write with blood earlier and I think that still does work, it's just chiseling that doesn't seem to work)

and I did try right clicking https://github.com/UristMcStation/UristMcStation/pull/671

_No response_

### Steps to reproduce

1. sharp object
2. can't cut floor

https://github.com/search?q=repo%3AUristMcStation%2FUristMcStation%20graffiti&type=code

Client Version: 515
Server Revision: [cc40971bd063df88431434211de6e64cfd493b33](https://github.com/UristMcStation/UristMcStation/commit/cc40971bd063df88431434211de6e64cfd493b33) - 431434211de6e64cfd493b33 - 2024-02-17
Game ID: csG-awYT
Current map: ICS Nerva

### Issue bingo

CPU-Blanc:
Tackles a few low-hanging bugs while I have a bit of spare time.

- Fixes #1352 - Caused by the subtype not setting the `condi` boolean
- Fixes #1381 - Deleted the "rank" header from the crew records nano program
- Fixes #1368 - This actually hurts my brain. Previously, clicking on a job to unban it would cause the code to **loop through every job in existence, check if they're banned, and unban them**. Rather than.. using the job passed to it in the href? This would not work for roles that aren't explicitly jobs, such as antag and special roles.
- Fixes unbanning categories in the unban panel - Related to above, it now populates a "unban" list, which it loops through and unbans. Early return added to clicking category headers, prioritising banning all if only some jobs in that category are previously banned.
- Removed duplicate job entries in the unban panel. It used to load all job titles, then output the job datum associated with it for each job title. A check has been added to make sure only the base job is listed, and alt-titles are ignored.
- Fixes #1411 - Caused by a regression during baymerge

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -812,7 +812,7 @@
 
 		//Unbanning job list
 		//all jobs in job list are banned already OR we didn't give a reason (implying they shouldn't be banned)
-		if(length(bannedlist))
+		if(LAZYLEN(bannedlist))
 			if(!config.ban_legacy_system)
 				to_chat(usr, "Unfortunately, database based unbanning cannot be done through this panel")
 				DB_ban_panel(M.ckey)
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -347,7 +347,8 @@
 		for(var/jobPos in SSjobs.titles_by_department(COM))
 			if(!jobPos)	continue
 			var/datum/job/job = SSjobs.get_by_title(jobPos)
-			if(!job) continue
+			if(!job || (jobPos in job.alt_titles))
+				continue
 
 			if(jobban_isbanned(M, job.title))
 				jobs += "<td width='20%'><a href='?src=\ref[src];jobban3=[job.title];jobban4=\ref[M]'>[SPAN_COLOR("red", replacetext(job.title, " ", "&nbsp"))]</a></td>"
@@ -367,7 +368,8 @@
 		for(var/jobPos in SSjobs.titles_by_department(SPT))
 			if(!jobPos)	continue
 			var/datum/job/job = SSjobs.get_by_title(jobPos)
-			if(!job) continue
+			if(!job || (jobPos in job.alt_titles))
+				continue
 
 			if(jobban_isbanned(M, job.title))
 				jobs += "<td width='20%'><a href='?src=\ref[src];jobban3=[job.title];jobban4=\ref[M]'>[SPAN_COLOR("red", replacetext(job.title, " ", "&nbsp"))]</a></td>"
@@ -388,7 +390,8 @@
 		for(var/jobPos in SSjobs.titles_by_department(SEC))
 			if(!jobPos)	continue
 			var/datum/job/job = SSjobs.get_by_title(jobPos)
-			if(!job) continue
+			if(!job || (jobPos in job.alt_titles))
+				continue
 
 			if(jobban_isbanned(M, job.title))
 				jobs += "<td width='20%'><a href='?src=\ref[src];jobban3=[job.title];jobban4=\ref[M]'>[SPAN_COLOR("red", replacetext(job.title, " ", "&nbsp"))]</a></td>"
@@ -409,7 +412,8 @@
 		for(var/jobPos in SSjobs.titles_by_department(ENG))
 			if(!jobPos)	continue
 			var/datum/job/job = SSjobs.get_by_title(jobPos)
-			if(!job) continue
+			if(!job || (jobPos in job.alt_titles))
+				continue
 
 			if(jobban_isbanned(M, job.title))
 				jobs += "<td width='20%'><a href='?src=\ref[src];jobban3=[job.title];jobban4=\ref[M]'>[SPAN_COLOR("red", replacetext(job.title, " ", "&nbsp"))]</a></td>"
@@ -430,7 +434,8 @@
 		for(var/jobPos in SSjobs.titles_by_department(MED))
 			if(!jobPos)	continue
 			var/datum/job/job = SSjobs.get_by_title(jobPos)
-			if(!job) continue
+			if(!job || (jobPos in job.alt_titles))
+				continue
 
 			if(jobban_isbanned(M, job.title))
 				jobs += "<td width='20%'><a href='?src=\ref[src];jobban3=[job.title];jobban4=\ref[M]'>[SPAN_COLOR("red", replacetext(job.title, " ", "&nbsp"))]</a></td>"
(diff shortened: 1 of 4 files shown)


code/modules/admin/topic.dm
```diff
@@ -794,16 +808,17 @@
 						return 1
 				if("Cancel")
 					return
+			return
 
 		//Unbanning job list
 		//all jobs in job list are banned already OR we didn't give a reason (implying they shouldn't be banned)
-		if(LAZYLEN(SSjobs.titles_to_datums)) //at least 1 banned job exists in job list so we have stuff to unban.
+		if(length(bannedlist))
```
scrdest, line 815:
```suggestion
		if(LAZYLEN(bannedlist))
```
LAZYLEN is currently macro'd out to length() anyway, but it would save on maintenance if Bay ever edits the definition. Plus, standardization is good.

code/modules/admin/topic.dm
```diff
@@ -347,7 +347,8 @@
 		for(var/jobPos in SSjobs.titles_by_department(COM))
 			if(!jobPos)	continue
 			var/datum/job/job = SSjobs.get_by_title(jobPos)
-			if(!job) continue
+			if(!job || (jobPos in job.alt_titles))
+				continue
```
scrdest, line 351: Instead of copypasting the edit for alt-titles everywhere, why don't we edit the jobs subsystem itself? 

I'm thinking we could add a new var on the subsystem to track CORE job titles for each department and ignore the alts. This would also provide a simple way of querying if something is in fact an alt-title - just query `SSjobs.base_titles_by_department()` (or whatever we call it) and if it's not there, it's either an alt/adminbus or something broke _horribly_.

We could also just straight up remove adding the alts to this list (here: https://github.com/Baystation12/Baystation12/blob/f1520a1a0ccf7ff9c1096bcda45f728130836950/code/controllers/subsystems/jobs.dm#L101), but this may have weird side-effects depending on where this method is used and how, so I wouldn't do that. 👀

code/modules/admin/topic.dm
```diff
@@ -734,9 +745,12 @@
 
 		//Create a list of unbanned jobs within job_list
 		var/list/notbannedlist = list()
+		var/list/bannedlist = list()
 		for(var/job in job_list)
 			if(!jobban_isbanned(M, job))
 				notbannedlist += job
+			else
+				bannedlist |= job
```
scrdest, line 753: This whole thing feels like it should just be tracked by the jobs subsystem honestly

scrdest (changes requested):
Happy with all the fixes, but think we can do a bit better with the jobbans

code/modules/admin/topic.dm
CPU-Blanc: I agree that making a new method within the job system to return only base jobs would be ideal. However, as this was just a quick bugfix PR, I was a little concerned about scope and future maintenance. I also wouldn't touch removing adding alts to that list, lord knows what bay in their infinite wisdom hook into that..

Could add that though if you'd like? Wouldn't take too much effort I'd imagine

code/modules/admin/topic.dm
CPU-Blanc: Honestly, this whole thing is a fucking mess lmao.
2k lines all in a single topic? Actually vile. But I feel like a re-work of this whole thing would likely fall out of scope of this PR. I'm not _too_ concerned as this only runs on-demand, when someone clicks ban/unban

code/modules/admin/topic.dm
scrdest: Hmm, honestly yeah, the odds of us touching that are low vs Bay messing with it. Developing downstream is hell 😿

scrdest (approved):
I guess we can todo these if we decide to free ourselves from Bayshackles

PR #1478, opened 2024-06-24
Fixes cyborg dropped stacks

### LINKED ISSUE #1457: Borg sheets break others' hands
### Description of issue

If a borg manages to get some of their sheets into someone else's hand, it superglues to their hand and renders it unusable. Haven't attempted reproducing.

Borgs either shouldn't be able to give sheets to others (but the functionality is there in sheet loaders) or sheets from borgs should successfully be removed from people's hands.

### Steps to reproduce

1. Borg selects steel sheet and sheet loader modules.
2. Borg uses sheet loader to load steel sheets, then RMB-drops them.
3. Person with hands picks them up.
4. Borg does borg things, changes out modules, unequipped both sheets and sheet loader in my case.
5. Sheets superglued to person's hands, only fixed by removal of hand.

engineering cyborg, sheet loader module, steel sheet module

Client Version: 514
Server Revision: e3aca7784eecadc12ec18f62cc24a9bda04f3401 - 2ec18f62cc24a9bda04f3401 - 2024-04-12
Game ID: cts-duDA
Current map: ICS Nerva

### Issue bingo

CPU-Blanc:
Closes #1457 

Caused by cyborg stacks being set to `canremove = FALSE`, meaning mobs cannot normally drop these. I explored allowing borgs to drop these, however, doing so allows for very broken behaviour.

For starters, the sheet loader doesn't _split_ on picking up the item, it simply references the same item. Meaning if you drop it, it's in _both_ the borg's inventory, as well as in the game world. 

Secondly, splitting these cyborg stacks spawns a _regular_ version of the material, meaning it now contains matter and bypasses the whole restriction on cyborg materials. So a regular mob who picks up one of these stacks, can split them almost infinitely (as it's bound to the borg still and uses charges, not amounts), and turns into regular metal. Oops.

Now if you drop one of these materials, it is simply unset from the gripper, freeing it back up

PR also removes a duplicate gripper from the engi borg

### DIFF
--- a/code/modules/mob/living/silicon/robot/modules/module_engineering.dm
+++ b/code/modules/mob/living/silicon/robot/modules/module_engineering.dm
@@ -44,7 +44,6 @@
 		/obj/item/taperoll/engineering,
 		/obj/item/taperoll/atmos,
 		/obj/item/gripper,
-		/obj/item/gripper,
 		/obj/item/gripper/no_use/loader,
 		/obj/item/device/lightreplacer,
 		/obj/item/device/paint_sprayer,
--- a/code/modules/mob/living/silicon/robot/drone/drone_items.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone_items.dm
@@ -196,7 +196,8 @@
 	if(!wrapped)
 		// Ensure fumbled items are accessible.
 		for(var/obj/item/thing in src.contents)
-			thing.dropInto(loc)
+			if(thing.canremove)
+				thing.dropInto(loc)
 		return
 
 	if(wrapped.loc != src)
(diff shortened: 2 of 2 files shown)


code/modules/mob/living/silicon/robot/modules/module_engineering.dm
```diff
@@ -44,7 +44,6 @@
 		/obj/item/taperoll/engineering,
 		/obj/item/taperoll/atmos,
 		/obj/item/gripper,
-		/obj/item/gripper,
```
Glloyd, line 47: how'd that even get there

Glloyd (approved):

PR #1480, opened 2024-07-05
[DRAFT] - Listening Outpost Away

Laniakeaaa:
Work in Progress.

A small listening outpost, capturing the signals of nearby vessels and stations.
1, maybe 2 potential spawns, compact, with not too much to use, has a small shuttle if you are lucky enough to spawn close enough to a Exoplanet.

Similar to DJ Stations back in ol' space, just need to see if there's a good way for it to somehow recieve radio signals.

StrongDMM-2024-07-05 19 52 00

### DIFF
--- a/maps/away/listeningpost/listening_areas.dm
+++ b/maps/away/listeningpost/listening_areas.dm
@@ -1,0 +1,1 @@
+// Add Areas here.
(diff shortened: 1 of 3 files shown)


Laniakeaaa:
Will see if it's worth redoing after Merge.

PR #1484, opened 2024-07-14
Fix landmarks not spawning correclty

### LINKED ISSUE #1392: Costume landmark doesn't spawn anything
### Description of issue

Base costume landmark doesn't seem to choose from any of the costume landmarks and spawn anything.
Non-base landmarks work such as costume/chameleon.

Base costume landmark should pick a random costume landmark and spawn it

### Steps to reproduce

1. Place /obj/effect/landmark/costume and start the server or go to the bottom deck maint costume storage by medbay storage
2. No costumes :(

/obj/effect/landmark/costume in code/game/objects/effects/landmarks.dm

Client Version: 514
Server Revision: 261f0cb64e444823da5166fb09b1ee9a9165146a - da5166fb09b1ee9a9165146a - 2023-11-26
Game ID: crm-aAcF
Current map: ICS Nerva

### Issue bingo

CPU-Blanc:
Fixes #1392 

I spent too long on this, end me. The issue was clear as day.

Caused by `delete_me` being `TRUE` on the base type, meaning it _always_ returned on ln 138, making the random part of the code unreachable...

Split the randomness of it into its own subtype `/costume/random`, because previously all types would've also had spawned random landmarks during their init call, which is not the intended behaviour. Yay inheritance!

### DIFF
--- a/code/game/objects/effects/landmarks.dm
+++ b/code/game/objects/effects/landmarks.dm
@@ -137,9 +137,12 @@
 	if (delete_me)
 		return INITIALIZE_HINT_QDEL
 
-	var/list/options = typesof(/obj/effect/landmark/costume)
-	var/PICK= options[rand(1,length(options))]
-	new PICK(src.loc)
+/obj/effect/landmark/costume/random/Initialize()
+	var/list/landmarks = subtypesof(/obj/effect/landmark/costume)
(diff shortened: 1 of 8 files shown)


Glloyd:
it's always something so seemingly simple.

PR #1487, opened 2024-07-14
Using a scope is now a dedicated hotkey - [CTRL + T]

Yantarii:
You can now unscope/scope without having to right click your weapon and go through the right click menu. This makes using scoped weapons not a test of how fast you can click Byond's awful menus.

I chose CTRL + V since it's close to the homerow and is easy to press quickly.

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -694,7 +694,7 @@ macro "hotkeymode"
 		name = "."
 		command = "move-down"
 	elem
-		name = "CTRL+V"
+		name = "CTRL+T"
 		command = "scope"
 
 macro "borgmacro"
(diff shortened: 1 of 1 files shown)


Yantarii:
Ideally we could have some sort of mappable keybind once Bay finishes that PR off for it.

CPU-Blanc (changes requested):
As someone who likes to use copy-paste hotkeys, this seems like a poor choice :P
Could we change it to something that isn't a global hotkey already?

ErrorsWindows:
normally i use the button in the object menu in the top right but this works too

Yantarii:
Adjusted to CTRL + T, as it isn't used anywhere

CPU-Blanc (approved):

PR #1492, opened 2024-07-16
[DRAFT] Resprites Old Urist Clothing Vendors

Laniakeaaa:
Some resprites I did for the clothing vendors, more or less the same design just different colors and decal on the front, a little noisy atm but I can adjust.
image


ErrorsWindows:
you should ms paint some other random vending machines into the picture so the aesthetic can be compared

Yantarii:
Original
image

New
348902469-d75f88bf-be4a-4397-8f38-67d2ac4ca425-1

Other
image
image

scrdest:
Maybe add a buckle outline to the belt? With the outline-only style, the belt looks like a donut or something.

Laniakeaaa:
closing for now, sprite dmi's are there if people want to use them for something else

PR #1495, opened 2024-07-16
Adjusts size & width of floating text + allows for ?? large text.

Yantarii:
- Increases the length of characters you can use in floating text, getting you closer to a larger full sentence of context when you are playing with floating text enabled. 

image

- Enables you to shout in profound confusion, as well as just normal loud voice sounds, by using ?? at the end of a sentence.

image

Context to a normal chat window - 
image

### DIFF
--- a/code/modules/mob/floating_message.dm
+++ b/code/modules/mob/floating_message.dm
@@ -20,7 +20,7 @@ var/global/list/floating_chat_colors = list()
 	if(small)
 		fontsize = 5
 	var/limit = 90
-	if(copytext(message, length(message) - 1) in list("!!","??","?!"))
+	if(copytext(message, length(message) - 1) in list("!!","?!"))
 		fontsize = 8
 		limit = 30
 		style += "font-weight: bold;"
--- a/code/modules/mob/floating_message.dm
+++ b/code/modules/mob/floating_message.dm
@@ -19,8 +19,8 @@ var/global/list/floating_chat_colors = list()
 	var/fontsize = 6
 	if(small)
 		fontsize = 5
-	var/limit = 50
-	if(copytext(message, length(message) - 1) == "!!")
+	var/limit = 130
+	if(copytext(message, length(message) - 1) in list("!!","??"))
 		fontsize = 8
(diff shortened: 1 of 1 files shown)


code/modules/mob/floating_message.dm
```diff
@@ -19,8 +19,8 @@ var/global/list/floating_chat_colors = list()
 	var/fontsize = 6
 	if(small)
 		fontsize = 5
-	var/limit = 50
-	if(copytext(message, length(message) - 1) == "!!")
+	var/limit = 130
```
CPU-Blanc, line 22: 130 feels a little too much for me, could end up cluttering the screen quite a bit. What about something in-between, in the 80-90ish area? How would that look?

code/modules/mob/floating_message.dm
```diff
@@ -19,8 +19,8 @@ var/global/list/floating_chat_colors = list()
...
+	if(copytext(message, length(message) - 1) in list("!!","??"))
```
CPU-Blanc, line 23: `??` seems like a bit of a odd choice for confused shouting. Would `?!`/`!?` make more sense?

Yantarii:
image

90 works for sure, I agree.

I've added ?! to the list, ?? I feel like is applicable in some cases - especially when I've been very surprised at something in game either all gunshots missing me or something else and it's quicker to do ?? instead of ?!

CPU-Blanc:
I disagree. `??` to me is being very quizzical, it doesn't implicate exclamation or volume. "You slipped him on.. what??" to me would simply elongate the "what", perhaps shift the pitch upwards. I wouldn't read that as someone shouting.

I understand this is just the difference between people, but I myself wouldn't want my text with `??` treated as if I'm shouting it, whereas `?!` shows explicit intent to yell. PR looks good though, so I'm not gonna block it on that basis alone

Glloyd:
I agree with Blanc about the ??, I think ?! Gets the point across better

Yantarii:
?? removed

CPU-Blanc (approved):
LGTM

PR #1498, opened 2024-07-16
Adds Ancient Toolbox & Jetfuel to Uplink, Adjusts Item Desc + Catergories.

Yantarii:
#### Adds Reagents & Plants Catergory

This will be used for buyable chemicals, toxins, plant matter, animal life, etc. Which will come later.

## Traitor Items:

**Ancient Toolbox -**

A more robust toolbox, available to buy at the badassery category, great for being a true greyshirt.

**Jet Fuel -**

Readded from conspiracy gamemode, jet fuel is a vial that may or may not contain thermite, or welding fuel instead. Good for quick break-ins on a budget.

## Minor Changes:

Sneakies Desc changed to reflect what they actually do.
Clown Declumsy Device moved to Medical w/ a warning as to what it will or will not do.


Glloyd:
you need to resolve the conflict here before I can merge it

Glloyd (changes requested):
should be good to merge if you resolve that conflict

PR #1474, opened 2024-06-23
[IDB IGNORE] [MDB IGNORE] Merge baymerge-testing into master

CPU-Blanc:
Seems like most major issues have been tackled, outside of a few outliers.
Should be good to merge back into master after the last issues are tackled, I'd imagine

Blockers:

- [x] #1369 
- [x] #1415 
- [x] #1457 
- [x] #1372

PR #1500, opened 2024-07-16
Assistants now have access again to Tool Storage.

Yantarii:
Assistants no longer have to B&E speedrun instantly.
They now have access to just tool storage, and anything with no access requirements.

It seems that there is an override for the assistant job which will remove any access you give it, so I've just commented it out, after an hour of debugging and waiting.

Fixes: https://github.com/UristMcStation/UristMcStation/issues/1415 - Initial Bug Report
Fixes: https://github.com/UristMcStation/UristMcStation/pull/1474  - Assistant Access Goal

### DIFF
--- a/maps/nerva/obj/nerva_ids.dm
+++ b/maps/nerva/obj/nerva_ids.dm
@@ -50,3 +50,9 @@
 
 /obj/item/card/id/medical/psychiatrist/nerva
 	job_access_type = /datum/job/psychiatrist
+
+/obj/item/card/id/civillian/nerva_assistant
+	name = "identification card"
+	desc = "A card issued to the ship's assistant"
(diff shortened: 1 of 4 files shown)


maps/nerva/obj/nerva_ids.dm
```diff
@@ -50,3 +50,9 @@
 
 /obj/item/card/id/medical/psychiatrist/nerva
 	job_access_type = /datum/job/psychiatrist
+
+/obj/item/card/id/civillian/nerva_assistant
```
Glloyd, line 54:
```suggestion
/obj/item/card/id/civilian/nerva_assistant
```

Glloyd:
you have a failing test, looks like you didn't change the actual object path

PR #706, opened 2019-06-02
Adds shimmering orbs.

ShibberinoShipperino:
🆑
Adds shimmering orbs, small floating orbs filled with bioluminescent growths.
🆑

Smalls rocks that could be used as a light source, most likely to be found on asteroids.
They have not been mapped on any away missions as of yet.
Made this sprite a while back and never got to use it.

Images: 
ShimmeringOrb
Capture

PR #1509, opened 2024-07-19
Restores Shimmering Orb asteroid flora + setups landmarks

Yantarii:
58758375-893ea380-8512-11e9-857e-4122f9762b93

- Readds Asteroid Flora from - https://github.com/UristMcStation/UristMcStation/pull/706
Lost to one of the baymerges

- Moves icon out of the way.

- Adds a landmark system we can use to define them spawning in if placed in areas.
I'll add support later if we get more flora to support different types.


code/modules/urist/structures/uristflora.dm
```diff
@@ -0,0 +1,24 @@
...
+
+// Shimmering Spawner -
+
+/obj/effect/spawner/structure/shimmeringorb				// Once we get more asteroid-specific flora, I'll adjust this to spawn a mixture, instead of just these.
+	var/obj/structure/flora/shimmering_orb/H
+
+/obj/effect/spawner/structure/shimmeringorb/Initialize()
+	. = ..()
+	if (rand(0, 3) == 0)
+		H = new/obj/structure/flora/shimmering_orb/(loc)
```
CPU-Blanc, line 24: Is there any reason we're referencing the spawned orb from the spawner? Surely this should be a one-time thing and the spawner should be deleted afterwards?
```suggestion
/obj/effect/spawner/structure/shimmeringorb/Initialize()
	. = ..()
	if(prob(25)
		new /obj/structure/flora/shimmering_orb(loc)
	return INITIALIZE_HINT_QDEL
```

CPU-Blanc (changes requested):

code/modules/urist/structures/uristflora.dm
Yantarii: ah good point, fixed

CPU-Blanc (approved):

PR #1522, opened 2024-07-21
Additional Example Map Additions:

Yantarii:
image

Suit Cycler, Suit Storage, Premade Mech, Captain's Spare.

image
A whole bunch of missing machinery - mainly botanical, kitchen, lathes, etc.
Atmos adjustments.
Additional material stacks including sheets of animal produce, etc.

image

All vendors in SS13, and the tools to hack and modify them.
Since most items are gotten through vendors, it's nice to have them all in one place.


maps/example/example-3.dmm
CPU-Blanc, line 1: Honestly not a fan of this. A large empty space is great for testing, and these vending machines feel very situational, rather than something that should always be there. They could simply be spawned in as and when required.

CPU-Blanc (changes requested):

maps/example/example-3.dmm
Yantarii: Removed all of the vendors in example-3 and everything else changed

CPU-Blanc (approved):

PR #1525, opened 2024-07-21
Lone Pioneer - Exoplanet Joinable Ruin

Yantarii:
A cabin setup by a lone survivor of a ship crash, well stocked and comfy.
Only one person can spawn as the pioneer, a good chunk of amenities and such to survive on, while living out in the harsh lands of whichever planet that spawn on.

StrongDMM-2024-07-23 22 19 37
Inside of the Turf Removers is a small garden and a small area with supplies to the south.

Available with more than 4+ players on station.

It's pretty much all mapped out and good to go, just have to test some things and get it to spawn alright and it's good.


Yantarii:
image

In game.

PR #1529, opened 2024-07-24
Sandwich Stacking Gains

Laniakeaaa:
You can fit more ingredients into a sandwich before needing to add a slice, gives it a more of a fun value, adding a slice lets you add even more.


ggermanicus (approved):
stack em tall brother

PR #1530, opened 2024-07-24
Janitorial Cart can now store Plungers and Firearms

Laniakeaaa:
Janitor Cart can now store plungers
Janitor Cart can store a firearm for advanced cleaning services.

If someone could help me figure out why NanoUI isn't updating when I put the plunger or weapon in the contents I would be very happy. 

Other than that, it works.

- [x] Fix Nano UI not updating for Gun + Plungers


ggermanicus (changes requested):
You should just need to go through janitorcart.tmpl and add links so you can remove the gun and plunger once you put them in.

Laniakeaaa:
> You should just need to go through janitorcart.tmpl and add links so you can remove the gun and plunger once you put them in.

fixed, thanks!

ggermanicus (approved):

code/game/objects/structures/janicart.dm
```diff
@@ -60,6 +61,22 @@
...
+
+	else if(istype(I, /obj/item/gun) && !mygun)
+		if(!user.unEquip(I, src))
+			return
+		mygun = I
+		update_icon()
+		updateUsrDialog()
+		to_chat(user, SPAN_NOTICE("You put [I] into [src]."))
+
 	else if(istype(I, /obj/item/device/lightreplacer) && !myreplacer)
```
scrdest, line 80: The copypasta is getting out of hand here. 

We could just have an assoc list at the top of the function here of {valid_type: overlay_icon_state}, check the type of I is in that list and do the rest generically (except for mop/signs which are special cases, but that can be done by checking for them first and ELSE-ing into the generic case) or something.

Annoyingly it's not that straightforward, because we have the 'slots', so I won't hard-block this, but I'm not super happy with pasta encroachment.

PR #1533, opened 2024-07-25
Some Xenophage Quick Fixes

Laniakeaaa:
- Resin Doors are now able to auto-close on a fairly quick time.
- Alien Weed Nodes now have 0 opacity and can be seen through
- Alien Weeds now have 0 opacity and can be seen through.
- Larva now get the vent-crawl proc to go run around the place.


ggermanicus (approved):
Unsure if xenos need auto closing doors on top of all their other supreme powers, but everything works, LGTM

PR #1534, opened 2024-07-25
Hextromycosalinate makes you trip balls again properly

Yantarii:
Brings back seeing colours when you take Hextro.
Nice and easy fix.


scrdest (approved):
The hero we need

CPU-Blanc (approved):

PR #1540, opened 2024-07-29
Ports 16 artifact effects from Polaris; fixes others

ggermanicus:
Heed my wisdom: do not port this much at once.

Adds the following effects, most of which fit their name:

- Animate Anomaly
- Cannibal Feeling
- Disarmament
- Electric Field
- Extinguisher
- Gaia
- Gas generation (O2, CO2, N2, Phoron, N2O)
- Gravity Wave
- Indiscriminate Spines
- Poltergeist
- Resurrect
- Tesla
- Vampire

Most if not all artifact effects will now also sparkle or do something else cosmetic when activated.

Modifies already present artifact effects:

- Cell drain and charge now also affect IPCs/FBPs/whoever else has cell organs
- Heal, hurt, and stun do more of what they do, because they were very weak and sad
- Hurt no longer affects synths...
- because robohurt and roboheal will.
- Sleepy will also affect synths mechanically. Messages indicating some effect existed without enforcement.
- Fixes punctuation/very minor spelling mistakes/len use in other effects.

I started this five months ago and can't remember anything. It works when I test it. I embrace undiscovered bugfix/runtime/review hell.


Glloyd:
This rocks, artifacts are relatively boring as is, so anything to make them more interesting is a good thing

code/modules/urist/datums/xenoarch_effects/cannibal.dm
scrdest, line 1: I'm not terribly comfortable with this one due to the self-antagging potential the messages have here. If this was gated behind adminbus or something, it might have been fine, but I don't think the selection logic is that sophisticated (IIRC it just does a typesof + pick).

code/modules/urist/datums/xenoarch_effects/disarm.dm
```diff
@@ -0,0 +1,29 @@
...
+/datum/artifact_effect/disarmament/proc/disarm(mob/living/L)
+	if(istype(L))
+		var/obj/item/Item = L.get_active_hand()
+		if(istype(Item))
+			to_chat(L, SPAN_WARNING("Something forces you to drop \the [Item]."))
+			L.drop_item()
+
+/datum/artifact_effect/disarmament/DoEffectTouch(mob/living/user)
+	var/weakness = GetAnomalySusceptibility(user)
+	if(prob(weakness * 100))
```
scrdest, line 14: Might be worth to sanitize it:
```suggestion
	var/weakness = clamp(GetAnomalySusceptibility(user), 0, 1)
	if(prob(weakness * 100))
```

code/modules/urist/datums/xenoarch_effects/poltergeist.dm
```diff
@@ -0,0 +1,42 @@
...
+
+/datum/artifact_effect/poltergeist/proc/throw_at_mob(mob/living/target, damage = 20)
+	var/list/valid_targets = list()
+	for (var/obj/O in oview(world.view, target))
+		if (!O.anchored && isturf(O.loc))
+			valid_targets |= O
+	if (length(valid_targets))
+		var/obj/obj_to_throw = pick(valid_targets)
+		obj_to_throw.visible_message("<span class='alien'>\The [obj_to_throw] levitates, before hurtling toward [target]!</span>")
+		obj_to_throw.throw_at(target, world.view, min(40, damage * GetAnomalySusceptibility(target)))
```
scrdest, line 42: A bit weird that the damage scales by the anomaly susceptibility of what we're throwing stuff *at* rather than the thing we throw. 

In principle, an unshielded guy standing right behind an anomaly-shielded guy could somehow cause the latter to take twice as much damage when acting as a human shield than when the shielded guy gets the item yeeted at the him directly.

code/modules/urist/datums/xenoarch_effects/poltergeist.dm
```diff
@@ -0,0 +1,42 @@
...
+		throw_at_mob(target, chargelevelmax)
+
+
+/datum/artifact_effect/poltergeist/proc/throw_at_mob(mob/living/target, damage = 20)
+	var/list/valid_targets = list()
+	for (var/obj/O in oview(world.view, target))
+		if (!O.anchored && isturf(O.loc))
+			valid_targets |= O
+	if (length(valid_targets))
+		var/obj/obj_to_throw = pick(valid_targets)
```
scrdest, line 40: Do we perhaps want to treat targets, e.g. closer or further, preferentially? That would be fairly easy to do since we've already got a PriorityQueue type defined.

code/modules/urist/datums/xenoarch_effects/resurrect.dm
scrdest, line 1: This seems *way* OP.

scrdest (changes requested):

code/modules/urist/datums/xenoarch_effects/cannibal.dm
ggermanicus: I think any dangerous effect has self-antagging potential but it's limited by having to drag it around for it to matter a lot - in other words it's extremely obvious to staff if someone is aggravating it. Cannibal thoughts won't ingrain 'I-need-to-drag-this-to-the-armory' thoughts, or even 'I-need-to-be-near-this' thoughts. Someone might die, but if it escalates beyond the room then any self-antagging is very obvious.

I'll remove it if you still feel it's best but wanted to give my two cents.

code/modules/urist/datums/xenoarch_effects/extinguisher.dm
```diff
@@ -0,0 +1,47 @@
...
+				lowertemp.temperature = max(min(lowertemp.temperature-2000, lowertemp.temperature / 2), 0)
+				lowertemp.react()
+				T.assume_air(lowertemp)
+			qdel(hotspot)
+
+		if (environment && environment.temperature > min_temperature) // Abstracted as steam or something
+			var/removed_heat = clamp(120 * WATER_LATENT_HEAT, 0, -environment.get_thermal_energy_change(min_temperature))
+			environment.add_thermal_energy(-removed_heat)
+			if (prob(5) && environment && environment.temperature > T100C)
+				T.visible_message(SPAN_WARNING("The water sizzles as it lands on \the [T]!"))
```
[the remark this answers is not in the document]
ggermanicus: Copypasta from water/touch_turf, mostly because without it it doesn't actually extinguish, and this artifact doesn't _actually_ generate water

code/modules/urist/datums/xenoarch_effects/electric_field.dm
```diff
@@ -0,0 +1,62 @@
...
+	for (var/mob/living/L in oview(effectrange, get_turf(holder)))
+		if (L == user)	// You're "grounded" when you contact the artifact.
+			continue
+		if (!L.stat)
+			nearby_mobs |= L
+	for (var/obj/machinery/light/light in range(effectrange, get_turf(holder)))
+		light.flicker()
+	for (var/mob/living/L in nearby_mobs)
+		if (L.isSynthetic())
+			to_chat(L, "<span class='danger'>ERROR: Electrical fault detected!</span>")
```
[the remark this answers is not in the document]
ggermanicus: Done other than alien spans which don't have a macro set up

code/modules/urist/datums/xenoarch_effects/poltergeist.dm
ggermanicus: Now adjusted to scale by w_class

code/modules/urist/datums/xenoarch_effects/resurrect.dm
ggermanicus: It's long odds to roll any specific anomaly effect, and I don't think it's a bad thing for mysterious artifacts to matter. That and every revive requires a whole bunch of someone else's health, it's not free by any means.

code/modules/urist/datums/xenoarch_effects/cannibal.dm
ErrorsWindows: it's only self antag if it's only yourself
which it's not myself, because the voices told me to examine your blood vessels

ErrorsWindows:
I agree glooyd they _should_ be overpowered so that science teamsters actively seek to loot artifacts without documentation like the true indiana jonses
An alien fungus growing on your engines that makes them go 10 times faster would be nice if only it didn't come with a side of zombie apocalypse
At least it's a lot more dynamic and scenario-building for stories than something like toxins or telescience that only seems to exist to blow up half the asteroid and then teleport the mangled miner bodies from the asteroid into the hospital

scrdest:
> I agree glooyd they _should_ be overpowered so that science teamsters actively seek to loot artifacts without documentation like the true indiana jonses An alien fungus growing on your engines that makes them go 10 times faster would be nice if only it didn't come with a side of zombie apocalypse At least it's a lot more dynamic and scenario-building for stories than something like toxins or telescience that only seems to exist to blow up half the asteroid and then teleport the mangled miner bodies from the asteroid into the hospital

I agree in principle, but bear in mind - Xenoarch _starts_ with an artifact.

I... don't actually think that having artifact spookiness take over the round-gimmick is a bad idea, actually, but we'd need to implement that very _deliberately_ or it will just be annoying.

I would propose a plan along the lines of:
1. Make a custom spawner for ship-side roundstart artifacts. By default nerf the possible effects to things that don't affect the round. We could do it by having a var on each effect to the effect of: `var/roundstart = TRUE/FALSE`
2. Keep the full list in the normal spawner for random finds
3. Add a new gamemode ("Incident" might be a good short name) that rewires the ship spawner from guaranteed safe-ish to guaranteed oh-crap. 

The new mode could compose with standard antag modes (e.g. Incident + Ling) and in fact I'd make it impossible to pick it without SOME other antag in case the crew doesn't interact with the artifact.

This way the players could decide whether they want to mess about with spess SCP or have a good old fashioned round of sitting in the Bar and chasing the Clown.

We could later make this mode more generic and wire up other kinds of randomly picked PvE event things, effectively generalizing Xenoarch Incident, Extended's increased event rates, and old modes like Meteor, Blob or Virus, but that's looking waaaay into the future.

If we're broadly happy with points 1 & 2, I can PR that sometime in early Sept.

ErrorsWindows:
of course I could have worded that better, only a month later now is my mind flooded with the repressed memory of all the times the /vg/ xenoarchaeologist players kind of just gunked it with the dumbshit [large artifacts](https://ss13.moe/wiki/index.php?title=Guide_to_Xenoarchaeology#Large_Artifacts_and_you)
that one where they just wordlessly make the air colder or suck it out
or the classic goldslime loose
or that funny forcefield they would put in departures so that nobody could get on to the escape shuttle
or how could I forget the cultify effect where they would drag it around the halls and make the walls and floors look like actual shit

Lest we forget that they can neither ruin the round or tell a good story for the party, but instead fall comfortably in between the two

ggermanicus:
Plan sounds good to me. Seconding incident coming w/ other antag modes as even hellportal is about as round-impacting as a blob, and not really a headliner event.

ErrorsWindows:
> in case the crew doesn't interact with the artifact

that's probably the second most important part, it might not even have any impact at all - there might not be a science player or there might not be a good one, just look as far as the non-science crew going off the ship and touching artifacts without gloves only for them to end up in traction in medbay for the next 30 minutes because of radiation

and I also now think that I may have overstated the overall impact on the party with the original statement, as opposed to its capacity for building an individual story

code/modules/urist/datums/xenoarch_effects/extinguisher.dm
```diff
@@ -0,0 +1,47 @@
...
+/datum/artifact_effect/extinguisher/New()
+	. = ..()
+	effect = pick(EFFECT_PULSE, EFFECT_TOUCH)
+
+
+/datum/artifact_effect/extinguisher/DoEffectTouch(mob/living/user)
+	if(user)
+		if(istype(user, /mob/living/carbon/human))
+			user.ExtinguishMob()
+			user.fire_stacks = clamp(user.fire_stacks - 1, -25, 25)
```
scrdest, line 17: Should just do a water_act() on the user atom too if we're acting like this actually produces moisture. It will already handle fire stuff as well.
```suggestion
	if(user)
	        user.water_act(2)
```

scrdest (changes requested):
I've done a review pass before you pushed a commit, so some comments might already be outdated

Glloyd:
grant made the changes that I requested to revive, so we can always tweak the chance for the artifact to disintegrate if need be. i think this is fine to merge as is 👍

PR #1543, opened 2024-07-30
Hypospray is instant again + storable in a Belt or Holster

Laniakeaaa:
CMOs everywhere screaming, crying, throwing up.
Hypospray now has a 0 timer again, instead of 7, making it a valuble item to keep again + for offensive use.

You can now store Hyposprays in your Belt slot, or a holster if you want to get real cowboy showdown with it, as a Doctor.

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -179,7 +179,7 @@
 		return
 
 	user.set_machine(src)
-	var/dat = "<B><HR>[FONT_LARGE(name)]</B><BR><HR>"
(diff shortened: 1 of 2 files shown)


ErrorsWindows:
if the hypospray does this what about the autoinjectors that are functionally similar

Yantarii:
> if the hypospray does this what about the autoinjectors that are functionally similar

The hypospray item uses hypospray/vial and is the only thing that will inherit this change, hypospray/autoinjectors will remain unaffected and the same behaviour as before.

ErrorsWindows:
oh wait this is the reloading of the vial not the injection

PR #1544, opened 2024-07-30
Autolathe Recipes - Geiger Counter, Ore Scanner

Laniakeaaa:
Autolathe can now print Geiger Counters and Orescanners.


Glloyd:
i feel like the geiger counter especially fits better as a protolathe recipe.

Yantarii:
Moved the Geiger to the Protolathe, let me know if these values work

PR #1551, opened 2024-08-05
Destroyed Colony - Alien AI + Neurotoxin Adjustment

Laniakeaaa:
Alien Simple Mobs now use AI_Holders 

- Warriors and Drones will use Melee AI Holders.
- Sentinels now use a Ranged AI Holder, and will attempt to move back to continue shooting if too close to an enemy, to compensate their machine-gun spitting ability has been reduced.
- Queen and Empress now use ranged AI holders, and have slightly increased spit attacks delay, and will attack with claws up close.
- Sets up Alien specific aggro_sounds, Queens and Empress have their own specific sounds, while sentinels and melee use the same.
- Lowers the damage of alien spit by 5, so it isn't always an instant bone-breaker shot.

- Claws now have a cutting sound, meaning the aliens will now cut instead of bludgeon sound you.

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/alien.dm
+++ b/code/modules/mob/living/simple_animal/hostile/alien.dm
@@ -15,6 +15,7 @@
 	health = 100
 	harm_intent_damage = 5
 	natural_weapon = /obj/item/natural_weapon/claws
+	bleed_colour = COLOR_LIME
 	attacktext = "slashed"
 	a_intent = I_HURT
 	attack_sound = 'sound/weapons/bladeslice.ogg'
@@ -30,7 +31,7 @@
 	see_in_dark = 10
 	move_to_delay = 1
 	pry_time = 6 SECONDS
-	ai_holder = /datum/ai_holder/simple_animal/melee/alien
+	ai_holder = /datum/ai_holder/simple_animal/urist_humanoid/alien/hunter
 	natural_armor = list(
 		melee = ARMOR_MELEE_KNIVES
 		)
@@ -43,6 +44,7 @@
 	health = 60
 	natural_weapon = /obj/item/natural_weapon/claws
 	move_to_delay = 2
+	ai_holder = /datum/ai_holder/simple_animal/urist_humanoid/alien/drone
 
 /mob/living/simple_animal/hostile/alien/sentinel
 	name = "alien sentinel"
@@ -55,7 +57,7 @@
 	projectiletype = /obj/item/projectile/neurotox
 	projectilesound = 'sound/weapons/pierce.ogg'
 	move_to_delay = 2
-	ai_holder = /datum/ai_holder/simple_animal/ranged/kiting/aliensentinel
+	ai_holder = /datum/ai_holder/simple_animal/urist_humanoid/alien/sentinel
 	base_attack_cooldown = 20
 
 /mob/living/simple_animal/hostile/alien/queen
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_animal/hostile/alien.dm
```diff
@@ -78,16 +84,37 @@
...
 	natural_armor = list(
 		melee = ARMOR_MELEE_KNIVES,
 		laser	= ARMOR_LASER_HANDGUNS,
 		energy	= ARMOR_ENERGY_SMALL
 		)
 
 /obj/item/projectile/neurotox
-	damage = 30
+	damage = 25
 	icon_state = "toxin"
```
scrdest, line 97: Oddly, we have a fork here. Player xenos use /obj/item/projectile/energy/neurotoxin (which deals a much more modest 5 TOX + 5 WEAKEN) instead. 

I can only assume this is a separate type to let simple_animal xenos deal damage to simple_animals or something in lieu of paralysis?

Even if we want to have a separate type, I'd suggest we change damage_type to DAMAGE_TOXIN to make more goddamn sense. I BELIEVE this would also stop it breaking bones, but I may be getting lost in the spaghetti.

scrdest (changes requested):

code/modules/mob/living/simple_animal/hostile/alien.dm
```diff
@@ -78,16 +84,37 @@
...
 	..(gibbed, "lets out a waning guttural screech, green blood bubbling from its maw...", show_dead_message)
 	playsound(src, 'sound/voice/hiss6.ogg', 100, 1)
+
+// AI Holders
+
+/datum/ai_holder/simple_animal/melee/alien
+	speak_chance = 0
+	aggro_sound = 'sound/voice/hiss2.ogg'
+
+/datum/ai_holder/simple_animal/ranged/kiting/aliensentinel
```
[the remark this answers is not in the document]
Yantarii: It would be a lot nicer if the default ai holder or ranged just had these options, I agree

code/modules/mob/living/simple_animal/hostile/alien.dm
Glloyd: Yeah, player xenos intentionally operate under different rules, partly as a holdover from the SCOM days, partially because player xenos have been so fucked up by Bay over the years. DAMAGE_TOXIN would function a little fucky with how toxin damage works in Baymed, but a switch to DAMAGE_BURN would probably work better here and would help with the limb breaking issue.

scrdest: Worth calling out that BURN has its own special rules, because God has abandoned this codebase long ago - most notably, _it creates hotspots_. So an alien splitting toxin could set the room on fire.

Yantarii:
Updated and change it up a bit, going to playtest and see what needs changing, I've made it the spit attacks do burn-damage for the moment, I imagine that using Toxin damage will lead to organ failure fairly quickly which will probably get frustrating for players, but I'm not well-versed in Baymed, I'll see how it goes

code/modules/mob/living/simple_animal/hostile/alien.dm
```diff
@@ -78,16 +91,110 @@
...
 		laser	= ARMOR_LASER_HANDGUNS,
-		energy	= ARMOR_ENERGY_SMALL
+		energy	= ARMOR_ENERGY_SMALL,
+		ballistic = ARMOR_BALLISTIC_MINOR
 		)
 
 /obj/item/projectile/neurotox
-	damage = 30
+	damage = 25
+	damage_type = DAMAGE_BURN
```
scrdest, line 105: I guess we can avoid the neurotox-firestarting issue by overriding the proc that causes it. There is a risk that this will drift from the Bay implementation though.
```
/obj/item/projectile/neurotox/on_impact(atom/A)
	impact_effect()		// generate impact effect
        // cut out the firestarting bits from Bay that were here
        return
```

Laniakeaaa:
Happy with how the melee is at the moment, but range-wise not so much. I'll look into adjusting more, might be a good thing for GOAI test-case or something else too if we want smarter xenos, though I'd have to get my head around how it works.

Glloyd:
I'm fine with this, as long as the firestarting spit is resolved.

PR #1552, opened 2024-08-05
Raft Construction Update - (SPAN + Desc)

Yantarii:
- Rafts now use SPAN_NOTICE keeping in line with current Bay Changes.
- Rafts now update their desc with the next stage needed to advance the craft, to help make it easier to figure out what you need.


scrdest (approved):

Yantarii:
Accidently pushed some UI to this, reverted, should be good again.

PR #1556, opened 2024-08-06
Standarizes almost all RIG Control Modules to support Backpacks in Suit Storage Slot

Laniakeaaa:
Standardizes all RIGs to allow for Backpacks.

Certain RIG Suits were missing the ability to carry a backpack in suit storage, mainly CE + CMO, while other RIG Suits already supported this. Made it so pretty much every RIG suit can place a Backpack in Suit Storage. 

RIG's aren't as common here but it's another feature to make it worthwhile to use one, probably some balance considerations as well for this.

### DIFF
--- a/code/modules/clothing/spacesuits/rig/suits/combat.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/combat.dm
@@ -54,7 +54,8 @@
 		/obj/item/storage/firstaid,
 		/obj/item/reagent_containers/hypospray,
 		/obj/item/roller_bed,
-		/obj/item/device/suit_cooling_unit
+		/obj/item/device/suit_cooling_unit,
+		/obj/item/storage/backpack
 	)
 
 /obj/item/clothing/shoes/magboots/rig/combat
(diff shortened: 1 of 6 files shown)


ggermanicus:
Unstaling this but I think this leaves RIGs with little to no downsides. Someone with more recent gameplay experience can probably weigh in

ErrorsWindows:
Nominally on bay code the hardsuit is balanced by the fact that it's rarer (only like 4 on the torch I think), it's hard to access (they're all command staff toys), it requires a higher EVA skill to use, and they're usually only able to be worn by humans and IPCs - and because this place isn't baystation, only one of those is really a downside that comes into regular play since nobody is subject to the weird whitelisting of everything

There's also actually wearing it like how you walk all slow and stuff and also if you get hurt it would take a while to perform the hardsuit removal surgery (which is based on the EVA skill rather than the surgery or anatomy skill, making it a bit slower at experienced as compared to surgery master, but like I said, not baystation)

I think if you really wanted to pester hardsuit enjoyers all the modules would come separately and you have to put them in before you use it for the first time but that would require a lot more work than these literal one line changes

CPU-Blanc:
I agree with Grant on this one. This makes RIGs _very_ strong in a way I don't think I'm comfortable with. Trying to stop people from running around in RIGs/hardsuits is already enough of a problem as-is.
Allowing a backpack slot means they'll just _constantly_ be in it, making them essentially invulnerable to antags fucking with the atmosphere/fires etc, or at the very least makes them a lot harder to deal with.

Imo RIGs should be worn with reason, not something you just troop about in just because.

I'd like to see other people's opinions

Laniakeaaa:
Thinking about it further, changed my opinion and agree.

PR #1557, opened 2024-08-06
Diving Helmet Tweaks

Laniakeaaa:
- Diving Helmets now hides your identity and your head/mask features.
- Diving Helmets now follow the same flash-protection parity as space helmets, and provide great flash protection when worn.
- Diving Helmets can now be worn by Humans, Unathi, Machines and Skrell (Must be uncomfortable)

### DIFF
--- a/code/modules/urist/mob/uristanimals.dm
+++ b/code/modules/urist/mob/uristanimals.dm
@@ -69,7 +69,7 @@ Please keep it tidy, by which I mean put comments describing the item before the
 
 //ARANEUS. Squee?
 
-/mob/living/simple_animal/hostile/retaliate/spider/araneus
+/mob/living/simple_animal/hostile/retaliate/bat/araneus
 	desc = "A fierce companion for any person of power, this spider has been carefully trained by NanoTrasen specialists. Its beady, staring eyes send shivers down your spine."
 	faction = "spiders"
 	harm_intent_damage = 3
--- a/code/modules/urist/mob/uristanimals.dm
+++ b/code/modules/urist/mob/uristanimals.dm
@@ -69,7 +69,7 @@ Please keep it tidy, by which I mean put comments describing the item before the
 
 //ARANEUS. Squee?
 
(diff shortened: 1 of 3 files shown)


code/modules/urist/mob/uristanimals.dm
```diff
@@ -69,7 +69,7 @@ Please keep it tidy, by which I mean put comments describing the item before the
 
 //ARANEUS. Squee?
 
-/mob/living/simple_animal/hostile/retaliate/bat/araneus
+/mob/living/simple_animal/hostile/retaliate/spider/araneus
```
ggermanicus, line 72: while araneus does not need to be a bat to avoid spawnkilling the HoS, you may want to make sure it still has the bat-traits it was inheriting like attacktext, response_harm/disarm, faction, maybe see_in_dark. all spiders are currently just hostile/giant_spider also, so araneus has nothing to inherit from /spider/ here and remains a pretend spider

Glloyd: yeah, Araneus was always a weird bat subtype because of being a retaliate creature, this is intentional and doesn't need to be fixed unless there's some issues being caused by inheriting weirdness from being a bat that have been missed in the last 10 years or so since Araneus was added (very possible)

Yantarii: I thought it was accidental change from baymerge, makes sense though I'll change it back

ggermanicus (approved):

CPU-Blanc (commented):
Not sure I agree with both making it available to skrell & unathi as well - those helmets are _tight_. Where do the tendrils and ears go?

Major flash blocking also seems odd for an old school diving helmets. I could imagine EVA helmets having special coatings to provide protection from stars (much like the mirrored visors on IRL space suits), but I can't imagine these retro helmets offering the same level of protection as that. Perhaps knock it down a level?

CPU-Blanc (approved):

PR #1560, opened 2024-08-23
[WIP] GhettoBooze 

Laniakeaaa:
Brew delicious and healthy drinks in the comfort of maintenance, using what you scavenge, adding more options for makeshift chemistry, using ethanol instead.

Currently:

**Welder Bomb**
Welding Fuel, Whiskey and Water.

**Rattle**
Motor Oil, Space Cola and Welding Fuel

More to come - I'd like to make purifying it an option, to make it safer at the end. I'll work on this more soon.

### DIFF
--- a/code/modules/reagents/Chemistry-Recipes.dm
+++ b/code/modules/reagents/Chemistry-Recipes.dm
@@ -3255,3 +3255,17 @@
 	catalysts = list(
 		/datum/reagent/enzyme = 1
 	)
+
+/datum/chemical_reaction/welderbomb
+	name = "Welder Bomb"
+	result = /datum/reagent/ethanol/welderbomb
+	required_reagents = list(/datum/reagent/ethanol/whiskey = 2, /datum/reagent/water = 1, /datum/reagent/fuel = 2)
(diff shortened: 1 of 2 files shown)


code/modules/reagents/Chemistry-Recipes.dm
```diff
@@ -3255,3 +3255,17 @@
 	catalysts = list(
 		/datum/reagent/enzyme = 1
 	)
+
+/datum/chemical_reaction/welderbomb
+	name = "Welder Bomb"
+	result = /datum/reagent/ethanol/welderbomb
+	required_reagents = list(/datum/reagent/ethanol/whiskey = 2, /datum/reagent/water = 1, /datum/reagent/fuel = 2)
```
scrdest, line 3262: I don't quite get the purpose of this recipe, design-wise. 

If someone has access to whiskey, why would they bother cooking up a toxic recipe instead? Yes, it stretches out the supply, but not significantly so considering the toxicity.

The Rattle recipe is far more significant, because it provides an actual resource _source_.

PR #1561, opened 2024-08-24
Away Map Maintenance - Bluespace River

Laniakeaaa:
Supermatter Cascade walls don't seem to really be defined as anything anymore in current Baymerge.

**Main Fixes:**

**turf/unsimulated/mask -> turf/simulated/mineral/random:**

Outskirt turf now uses simulated turf, meaning your lungs and the entire area won't get nuked if someone wants to do a little mining as a treat. Lets people get more minerals and artifacts too.

**Bluespace Liquid:**

Readds bluespace liquid melting just about anything it touches, supermatter cascade walls seem to be mostly removed in current Baymerge sadly, so the functionality seemed to be lost.

**Flora:**

Adds some more trees and such, gives it a little bit more life.

**Tunnels:**

Makes tunnels less straight, slightly more organic look.

StrongDMM-2024-08-24 01 38 27
StrongDMM-2024-08-24 01 38 34

### DIFF
--- a/maps/away/blueriver/blueriver-1.dmm
+++ b/maps/away/blueriver/blueriver-1.dmm
@@ -134,6 +134,10 @@
 	},
 /turf/simulated/floor/away/blueriver/alienfloor,
 /area/bluespaceriver/underground)
+"H" = (
+/obj/structure/scrap,
+/turf/simulated/floor/asteroid,
+/area/bluespaceriver/underground)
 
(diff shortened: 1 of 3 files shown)


ErrorsWindows:
oh so that's why the horizon ship looks so weird it's supposed to be a flying saucer

scrdest (approved):

PR #1580, opened 2024-11-02
ports Saints and Sins from Polaris + nerf nt store

ggermanicus:
i have good priorities

- Saints and Sins: spess deckbuilder MtG-style game. Procure box and deck from the games vending machine, and additional builder and booster packs from the NT store computer in the cargo lobby!
- The NT store computer in cargo has been overpriced for actual years, since no one ever nerfed the cost when Bay dramatically slashed spaceman incomes in like... 2015... Prices are now cut in half so an assistant stands a chance of owning a beach ball one day. Even if it's half or more of their bank account.


code/modules/urist/items/saintsandsins.dm
```diff
@@ -0,0 +1,681 @@
...
+	booster_type = "Fatebringer Kit"
+	pack_type = CARDPACK_FATEBRINGER
+
+/*
+// Builder pack cards can appear in other packs to fluff them up. Pack-specific are less likely to do so, if ever.
+*/
+/singleton/hierarchy/sas_card
+	var/desc
+	var/ability
+	var/pack_probability
```
scrdest, line 124: Is there any reason this is a Hierarchy type? AFAICT it doesn't use the Hierarchy API anywhere.

At first glance, it looks like it was PLANNED to be a full-on minigame, but was turned into just 'read the desc and do what it says manually'.

ggermanicus: Gave datums a shot and it runtimed+broke, can't think of a better way.

PR #1584, opened 2024-11-08
adjust AO layers and rename terminal layer

### LINKED ISSUE #1581: (some) Pipes/wires in engine waste loop are unclickable by default
### Description of issue

Certain pipes in Engine Waste Handling (north of the SM) and maint outside aren't clickable or are incompletely clickable and can only be interacted with via alt-click. As far as your mouse is concerned, it's just plating there!

Pipes that are visible to the player should be clickable to the player without alt-click or wizardry

### Steps to reproduce

1. mouse over previously described pipe/wire in engine waste handling/maint next to it
2. be unable to click on it

These pipes are all on Deck 2/main engine deck:
Both 4way manifolds on the left side (X:62 Y:102&103) are partially but not completely clickable
Everything including cable in scrubber pipe route beginning at X:64 Y:104 (partially clickable on the left side) and remaining unclickable until X:74 Y:105, EXCEPT X:068 Y:106 and X:66 Y:106 (partially clickable on the bottom side), disregarding the catwalk pipe
Power cable on X:068 Y:107

Weird things I did while trying to find the issue:

- Set the pipe layer higher, even though the layer is the same as it is on the rest of the map (works but illegal)
- Deleted wall at X:067 Y:107 and exposed the maint hallway to space. This fixed X:067 Y:106 and no other tile, even after repairing it. Seems to fix directly adjacent tiles in general.
- Put down a bunch of floodlights. This did not work.
- Deleted wall at X:072 Y:105 and exposed the hallway to Deck 3 instead of space. This fixed the air supply pipe and wire directly north of it, but not the scrubber pipe? Broke again when replacing the wall.
- Removed the tile under the pipes (fixed the tile, until I replaced it)
- Bled on the tiles (did nothing. lol)
- Set everything on fire (did nothing. lol)
- Put more cables down (nothing)
- Alt-click removed everything on the tile but the plating and put it back (nothing)
- Alt-click removed everything on the tile, including the plating, and put it back (nothing)

Client Version: 514
Server Revision: [1ca76f9a6083b3c260acdb1154c55706ed42f0ce](https://github.com/UristMcStation/UristMcStation/commit/1ca76f9a6083b3c260acdb1154c55706ed42f0ce) - baymerge-testing - 2024-11-02
Game ID: cwv-a0cc
Current map: ICS Nerva

### Issue bingo

ggermanicus:
fixes #1581 - you can click on all the pipes now!
also renames one layer used by one object

### DIFF
--- a/code/__defines/__renderer.dm
+++ b/code/__defines/__renderer.dm
@@ -63,28 +63,29 @@
 	#define TURF_SHADOW_LAYER           2.02
 	//ABOVE TURF
 	#define DECAL_LAYER                 2.03
-	#define RUNE_LAYER                  2.04
-	#define ABOVE_TILE_LAYER            2.05
-	#define EXPOSED_PIPE_LAYER          2.06
-	#define EXPOSED_WIRE_LAYER          2.07
-	#define EXPOSED_WIRE_TERMINAL_LAYER 2.08
-	#define CATWALK_LAYER               2.09
-	#define ABOVE_CATWALK_LAYER         2.10
-	#define BLOOD_LAYER                 2.11
-	#define MOUSETRAP_LAYER             2.12
-	#define PLANT_LAYER                 2.13
-	#define AO_LAYER                    2.14
+	#define AO_LAYER                    2.04
+	#define ABOVE_AO_LAYER              2.05
+	#define RUNE_LAYER                  2.06
+	#define ABOVE_TILE_LAYER            2.07
+	#define EXPOSED_PIPE_LAYER          2.08
+	#define EXPOSED_WIRE_LAYER          2.09
+	#define ABOVE_EXPOSED_WIRE_LAYER    2.10
+	#define CATWALK_LAYER               2.11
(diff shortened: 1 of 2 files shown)


scrdest (commented):
Okay in principle, but this will almost certainly be a maintenance nightmare for upstream merges (and/or will just get reverted by the merge).

Abusing conditional compilation macros (I think we have a 'is Uristcode' macro; #IFDEF against it to use the edited layers, #ELSE use Bay layers) might make this work better - at least, on merge we will see the edited values and have an easier time untangling it.

ggermanicus:
This is a port of a bay bugfix (I have no idea why cherrypicking didn't include other authors this time) so I think it will survive merge.

scrdest:
> This is a port of a bay bugfix (I have no idea why cherrypicking didn't include other authors this time [or maybe it never did and I forgot how to do it right]) so I think it will survive merge.

Ah, alright then, might be worth linking the cherrypicked PR/commit in the desc to make it clearer in the future

scrdest (approved):

PR #1594, opened 2024-12-07
Artifact: Hivebots!

ggermanicus:
This adds yet another artifact effect which turns the artifact into a hivebot fabricator. This fabricator:

- changes turfs
- spawns up to 5 hivebots
- spawns up to 5 worse fabricators that spawn up to 2 hivebots each

As of right now it's stronger than hellportal, much stronger if RNG hates you, and weakest early.
~There's a risk of the crew giving up on the deck and the artifact having so many turfs to convert that things start to stutter. This can be fixed by limiting the amount of turfs in range instead of letting it steadily grow forever but I'm hoping a wiser coder has a better idea so it can eventually consume the whole deck in peace.~ The for() responsible for this has been sacked.


Glloyd:
i gave my approval to this elsewhere ages ago, so 👍

PR #1610, opened 2025-03-22
uplink contortionist jumpsuit

ErrorsWindows:
title
allows you to ventcrawl through the pipes like a monkey
20 good boy points
uses an unused alternate robotics jumpsuit sprite, so it should be a bit visually distinctive
allows you to carry gloves, glasses, mask, ID, shoes, and pocket items while ventcrawling; but not anything else (e.g.: belt, backpack, helmet, hardsuit, in-hand)
does not protect you from the atmosphere or temperature inside the pipe - sorry skrell

https://github.com/vgstation-coders/vgstation13/pull/5332

i'd personally put it just above the chameleon projector but below the emag, very strategic item with how you have to leave all your stuff behind at the entry point, you're not just gonna whip it out and go through a crawl through the vents to stretch the legs

### DIFF
--- a/code/modules/urist/items/clothes/uristclothes.dm
+++ b/code/modules/urist/items/clothes/uristclothes.dm
@@ -1813,3 +1813,16 @@ Update 26/07/2014 - All generic clothing goes under obj/item/clothing/under/uris
 		bio = ARMOR_BIO_SMALL,
 		rad = ARMOR_RAD_MINOR
 		)
+
+/obj/item/clothing/under/contortionist
+	name = "contortionist's jumpsuit"
+	desc = "A light jumpsuit useful for squeezing through narrow vents."
+	icon_state = "robotics2"
+	item_state = "robotics2"
+
+/obj/item/clothing/under/contortionist/proc/check_clothing(mob/user as mob)
+	//Allowed to wear: glasses, shoes, gloves, pockets, mask, and jumpsuit (obviously)
+	var/list/slot_must_be_empty = list(slot_back,slot_handcuffed,slot_legcuffed,slot_l_hand,slot_r_hand,slot_belt,slot_head,slot_wear_suit)
+	for(var/slot_id in slot_must_be_empty)
+		user << "<span class='warning'>You can't fit inside while wearing those bulky clothes.</span>"
+		return 0
--- a/code/modules/urist/items/clothes/uristclothes.dm
+++ b/code/modules/urist/items/clothes/uristclothes.dm
@@ -1820,9 +1820,9 @@ Update 26/07/2014 - All generic clothing goes under obj/item/clothing/under/uris
 	icon_state = "robotics2"
 	item_state = "robotics2"
 
-/obj/item/clothing/under/contortionist/proc/check_clothing(mob/user as mob)
+/obj/item/clothing/under/contortionist/proc/ventcrawl_check_clothing(mob/user as mob)
 	//Allowed to wear: glasses, shoes, gloves, pockets, mask, and jumpsuit (obviously)
 	var/list/slot_must_be_empty = list(slot_back,slot_handcuffed,slot_legcuffed,slot_l_hand,slot_r_hand,slot_belt,slot_head,slot_wear_suit)
 	for(var/slot_id in slot_must_be_empty)
 		to_chat(user, SPAN_WARNING("You can't fit inside while wearing those bulky clothes."))
-		return 0
+		return
--- a/code/modules/urist/items/clothes/uristclothes.dm
+++ b/code/modules/urist/items/clothes/uristclothes.dm
@@ -1824,5 +1824,5 @@ Update 26/07/2014 - All generic clothing goes under obj/item/clothing/under/uris
 	//Allowed to wear: glasses, shoes, gloves, pockets, mask, and jumpsuit (obviously)
 	var/list/slot_must_be_empty = list(slot_back,slot_handcuffed,slot_legcuffed,slot_l_hand,slot_r_hand,slot_belt,slot_head,slot_wear_suit)
 	for(var/slot_id in slot_must_be_empty)
-		user << "<span class='warning'>You can't fit inside while wearing those bulky clothes.</span>"
+		to_chat(user, SPAN_WARNING("You can't fit inside while wearing those bulky clothes."))
 		return 0
--- a/code/modules/ventcrawl/ventcrawl.dm
+++ b/code/modules/ventcrawl/ventcrawl.dm
@@ -3,6 +3,15 @@ var/global/list/ventcrawl_machinery = list(
 	/obj/machinery/atmospherics/unary/vent_pump
 	)
 
+/mob/living/carbon/human/AltClickOn(var/atom/A)
+	if(is_type_in_list(A,ventcrawl_machinery) && w_uniform && istype(w_uniform,/obj/item/clothing/under/contortionist))
+		var/obj/item/clothing/under/contortionist/C = w_uniform
+		if(C.check_clothing(src))
+			src.handle_ventcrawl(A,1)
+			return
+	else
+		..(A)
+
 // Vent crawling whitelisted items, whoo
 /mob/living/var/list/can_enter_vent_with = list(
 	/obj/item/implant,
@@ -65,6 +74,8 @@ var/global/list/ventcrawl_machinery = list(
 		return carried_item.w_class <= ITEM_SIZE_NORMAL
 	return ..()
 
+
+
 /mob/living/proc/ventcrawl_carry()
 	for(var/atom/A in contents)
 		if(!is_allowed_vent_crawl_item(A))
@@ -80,6 +91,7 @@ var/global/list/ventcrawl_machinery = list(
 		handle_ventcrawl(A)
 		return 1
 	return ..()
+
 
 /mob/proc/start_ventcrawl()
 	var/atom/pipe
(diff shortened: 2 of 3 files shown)


ggermanicus:
Trying to think of use cases for this item. So far I can think of stealing the captain's ID roundstart and... crawling into perma and _then_ spawning all your items to break somebody out?

ErrorsWindows:
please clarify my failed check

The original denizens of /vg/ did a lot of thinking about it as well (in linked addition)
Your second example is probably the best one, A tier for making a stealthy entry but it's kinda mid for everything else because of the pockets, you just don't have as many options past that initial entry point compared to all the things an emag/door hacking tool can do on top of simply gaining access - you wanna steal the armory guns? ok how are you gonna get back out now
If anything it's probably a better tool for making a quick escape as you give the AI a migraine by avoiding all its cameras... It'll probably work better here since you can pester people with roleplay things like hacking people's computers

As an assassination tool, you would have to rely on items that can fit in a pocket (paralytic pen, concealed syringe gun, a holdout pistol that's probably stopped by the armor value of a second officer holding the stack of papers to his chest, I should probably add some other pocket items in a different pull request like that TG energy dagger or something) or whatever implants/augments you can bring along (which in fairness are not really a consideration on /vg/, implants are kind of lame and augments don't exist), a robust player will be able to pick off any HRPers that stand idly next to a vent with a chloral syringe or an ebow or the armblade augment or something, or just an esword like a normal person, but it's definitely not all powerful. Plus it's one of those things where the defense is just laughably easy, if you know someone's using a contortionist jumpsuit to attack people just supercharge the distro lmao watch as they crawl to medbay with popped lungs

People also made complaints about "stealin muh spare" yet seem to forget that regular non-antag crew members with toolbelts regularly prove that stealing the captain's spare is still a rather trivial exercise even when you remove traitor toys from the equation

ggermanicus:
Clarifying as much as I can about line 1827:
Code quality checks do not allow additional uses of <<. `user <<` is effectively a use of `to_chat`. To remove the code quality issue, replace line 1827 with something that fits, like a `to_chat(user` command.

Uses of `<span class='warning'>` and similar are functional but antiquated, clarity when reading code is improved by using `SPAN_WARNING` instead.

If you searched in Visual Studio Code you can see like ten thousand uses of `to_chat(user` as well as SPAN_WARNING and other types that you could easily replicate here, but the actual, full line replacement will just look like this:
`to_chat(user, SPAN_WARNING("You can't fit inside while wearing those bulky clothes."))`

scrdest:
> Clarifying as much as I can about line 1827: Code quality checks do not allow additional uses of <<. `user <<` is effectively a use of `to_chat`. To remove the code quality issue, replace line 1827 with something that fits, like a `to_chat(user` command.
> 
> Uses of `<span class='warning'>` and similar are functional but antiquated, clarity when reading code is improved by using `SPAN_WARNING` instead.
> 
> If you searched in Visual Studio Code you can see like ten thousand uses of `to_chat(user` as well as SPAN_WARNING and other types that you could easily replicate here, but the actual, full line replacement will just look like this: `to_chat(user, SPAN_WARNING("You can't fit inside while wearing those bulky clothes."))`

I'm honestly kind of confused why you bothered to clarify all that, this is just standard code style at this point

code/modules/ventcrawl/ventcrawl.dm
```diff
@@ -100,7 +112,7 @@ var/global/list/ventcrawl_machinery = list(
 /mob/living/carbon/alien/ventcrawl_carry()
 	return 1
 
-/mob/living/proc/handle_ventcrawl(atom/clicked_on)
+/mob/living/proc/handle_ventcrawl(atom/clicked_on, var/allowed_carry=0)
```
scrdest, line 115: What's the point of this parameter? It doesn't seem to affect anything.

code/modules/urist/items/clothes/uristclothes.dm
```diff
@@ -1813,3 +1813,16 @@ Update 26/07/2014 - All generic clothing goes under obj/item/clothing/under/uris
...
+	desc = "A light jumpsuit useful for squeezing through narrow vents."
+	icon_state = "robotics2"
+	item_state = "robotics2"
+
+/obj/item/clothing/under/contortionist/proc/check_clothing(mob/user as mob)
+	//Allowed to wear: glasses, shoes, gloves, pockets, mask, and jumpsuit (obviously)
+	var/list/slot_must_be_empty = list(slot_back,slot_handcuffed,slot_legcuffed,slot_l_hand,slot_r_hand,slot_belt,slot_head,slot_wear_suit)
+	for(var/slot_id in slot_must_be_empty)
+		to_chat(user, SPAN_WARNING("You can't fit inside while wearing those bulky clothes."))
+		return 0
```
scrdest, line 1828: I don't see how this will ever do anything but return 0 instantly every time. This is not checking anything about the slots AFAICT, it just blanket-returns 0.

scrdest: Separate issue altogether: this unnecessarily mixes logic and UI together. 

Check functions should be pure logic, if they return false/0/null? then we can decide how to communicate that to the user in... whatever we're doing (ventcrawling, in this case). 

Otherwise we if we want to check before committing to doing stuff, or similarly when the check inevitably gets repurposed for something else, we'll get random irrelevant chat message spam.

code/modules/urist/items/clothes/uristclothes.dm
```diff
@@ -1813,3 +1813,16 @@ Update 26/07/2014 - All generic clothing goes under obj/item/clothing/under/uris
...
 		rad = ARMOR_RAD_MINOR
 		)
+
+/obj/item/clothing/under/contortionist
+	name = "contortionist's jumpsuit"
+	desc = "A light jumpsuit useful for squeezing through narrow vents."
+	icon_state = "robotics2"
+	item_state = "robotics2"
+
+/obj/item/clothing/under/contortionist/proc/check_clothing(mob/user as mob)
```
scrdest, line 1823: This proc name is not terribly informative. At a glance, I wouldn't realize this is even a custom proc defined only for this subclass. 

ventcrawl_check_clothing()? ventcrawling_allowed()?

code/modules/ventcrawl/ventcrawl.dm
```diff
@@ -3,6 +3,15 @@ var/global/list/ventcrawl_machinery = list(
 	/obj/machinery/atmospherics/unary/vent_pump
 	)
 
+/mob/living/carbon/human/AltClickOn(var/atom/A)
+	if(is_type_in_list(A,ventcrawl_machinery) && w_uniform && istype(w_uniform,/obj/item/clothing/under/contortionist))
```
scrdest, line 7: grumble: I _hate_ single-type typecheck carveouts, though with the garbage legacy architecture there's not much we can do to work around this without making this a whole huge PR. Something something ECS something.

scrdest: Also, I would flip the order of checks here, is_type_in_list at the end. 

The other checks are much cheaper, and if any fails the rest won't run (short-circuiting).

CPU-Blanc: Quickly popping in with my usual `istype` _already_ checks for nulls. You don't need the added `&& w_uniform`

code/modules/ventcrawl/ventcrawl.dm
```diff
@@ -3,6 +3,15 @@ var/global/list/ventcrawl_machinery = list(
...
+		var/obj/item/clothing/under/contortionist/C = w_uniform
+		if(C.check_clothing(src))
+			src.handle_ventcrawl(A,1)
```
Glloyd, line 10: This will fail if it makes it to handle_ventcrawl, because the mob doesn't have the ventcrawl verb and won't pass the can_ventcrawl() check.

code/modules/ventcrawl/ventcrawl.dm
```diff
@@ -65,6 +74,8 @@ var/global/list/ventcrawl_machinery = list(
 		return carried_item.w_class <= ITEM_SIZE_NORMAL
 	return ..()
 
+
```
CPU-Blanc, line 77: whitespace?

code/modules/ventcrawl/ventcrawl.dm
```diff
@@ -81,6 +92,7 @@ var/global/list/ventcrawl_machinery = list(
 		return 1
 	return ..()
 
+
```
CPU-Blanc, line 95: and here

CPU-Blanc (changes requested):
scr and grant have already discussed much of what I'd have issue with in this PR, aside from gameplay/balance wise if we'd actually want such an item

Last thing I've not seen mentioned is some random whitespace additions. Can we remove these? It's just causing unnecessary diffs

Glloyd:
Scr beat me to it, I noticed it when I glanced at this last night, this won't work as written, for a few reasons. I like the idea, but you'll need to move the logic to the human can_ventcrawl and ventcrawl_carry procs for this to work as intended.

Glloyd:
Did we all just decide to comment on this at the same time?

scrdest:
> Did we all just decide to comment on this at the same time?

Uristcoder brain is implemented as a singleton to keep memory consumption low

Laniakeaaa:
I would change the name from contortionist jumpsuit to something more discreet as well, since it's a stealth item in comparison to something highly visible like a gun. 

My own two-cents balance-wise, I think this is fairly balanced since vent-crawling already makes fairly sounds while you move anyway, though one can argue this also defeats the point of it as a stealth item for assasination. Additionally if you did crawl infront of sec in the brig, they could just give you a prisoner uniform if they figured out that was what was causing it. 

The clothing/storage limitations pretty much stop easy armory looting without forking out money for a storage implant which I think is nice to make these implants more useful. We could also have vents store the fingerprints/fibers of someone if they did vent-crawl giving more considerations for using it.

ggermanicus:
> I'm honestly kind of confused why you bothered to clarify all that, this is just standard code style at this point

average_familiarity

ErrorsWindows:
you know i've come to the realization that i would much prefer to play video games rather than make them

PR #1613, opened 2025-03-25
EMPs lock robolimbs up instead of liquefying them; microbattery slightly more resistant to EMP

ggermanicus:
When EMPed all prosthetic/robolimbs will 'break' for up to 30 seconds. Timer won't stack.
Microbattery EMP damage adjusted so a full ion pistol shouldn't break it and leave you dead with no sensors in maint (just very close).

Adds another usage of spawn because I couldn't think of a better way to un-break the limbs.
This doesn't play great with bodyscanners if you happen to shove an EMP victim in there in the stun duration (shows 'Mechanical and' instead of 'Mechanical and Broken'), but goes back to normal when the EMP effects are over.

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -156,23 +156,25 @@
 		return
 
 	var/burn_damage = 0
+	var/stun_time = 0
 	switch (severity)
 		if (EMP_ACT_HEAVY)
+			stun_time = 1 MINUTE
 			burn_damage = 30
 		if (EMP_ACT_LIGHT)
-			burn_damage = 15
-
-	var/mult = 1 + !!(BP_IS_ASSISTED(src)) // This macro returns (large) bitflags.
-	burn_damage *= mult/species.get_burn_mod(owner) //ignore burn mod for EMP damage
+			stun_time = 30 SECONDS
+			burn_damage = 30
 
 	var/power = (3 - severity)/5 //stupid reverse severity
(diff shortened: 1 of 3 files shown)


Glloyd:
You'll need to update the code tests to include that spawn usage

code/modules/organs/external/_external.dm
```diff
@@ -156,23 +156,25 @@
...
 		if (EMP_ACT_HEAVY)
+			stun_time = 30 SECONDS
 			burn_damage = 30
 		if (EMP_ACT_LIGHT)
-			burn_damage = 15
-
-	var/mult = 1 + !!(BP_IS_ASSISTED(src)) // This macro returns (large) bitflags.
-	burn_damage *= mult/species.get_burn_mod(owner) //ignore burn mod for EMP damage
+			stun_time = 15 SECONDS
+			burn_damage = 30
```
Glloyd, line 166: Should this not remain as 15 for EMP_ACT_LIGHT? burn_damage still affects implanted objects. Actually, because line 175 was removed, burn_damage does nothing now, so the var should either be removed, or these checks should be modified to keep the effect for implanted objects.

Glloyd (changes requested):

ggermanicus:
Closing this for now, I will reopen it and do changes if https://github.com/UristMcStation/UristMcStation/pull/1624 is deemed insufficient

PR #1618, opened 2025-03-30
Holdout Revolvers, Single Pistol Bullets & Screwdrivers can be hidden in shoes.

Laniakeaaa:
**Features:**

- Holdout Revolvers - Can be hidden inside shoes.
- Pistol Bullets (Empty/Loaded) - Can be hidden inside shoes (singular)
- Screwdrivers - Can be hidden inside shoes.  (Bay Change - late 2023ish, we'll be getting it when we merge either way.)

With the Holdout Revolver, my intention is to make it more concealable at the cost of less magazine capacity & loudness, vs the Holdout Pistol which can be silenced & have more magazine capacity at the cost of not being as easily hidable if surpressed, or if pockets are checked.

### DIFF
--- a/code/game/objects/items/weapons/tools/screwdriver.dm
+++ b/code/game/objects/items/weapons/tools/screwdriver.dm
@@ -16,6 +16,7 @@
 	attack_verb = list("stabbed")
 	lock_picking_level = 5
 	sharp = TRUE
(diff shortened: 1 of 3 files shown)


Glloyd (approved):
putting a revolver in your shoe is funky, but at least its something that differentiates it from the holdout pistol

PR #1624, opened 2025-04-05
ports bay changes to ions

ggermanicus:
See https://github.com/Baystation12/Baystation12/pull/34346

This is infinitely more merciful than what I was going to do to the poor robots and I would rather try what they're doing first


Glloyd:
Works for me. We'll be getting this anyways whenever I have the time to get back to working on the merge, so we can see how it goes.

PR #1625, opened 2025-04-05
Lar Maria - Away Map Update

Laniakeaaa:
### General:

- Changed a good chunk of the default grey-floor tiles so Lar Maria looks slightly less grey.
- Scientist and Security Guards have their own dorm rooms color coded.
- Dorms have bedsheets now so no-one sleeps in the cold...
- Decal Pass on most of the map.
- Barricades added to certain areas on the map.

### Loot:

- Cafeteria has some food items that medbay/science can extract interesting chemicals from.
- Sci-Wings have some rare slime cores, phoron and chemical vial. 
- A potentially booby-trapped locker somewhere.

### Enemies:

- Enemy Placement adjusted and changed up somewhat for guards and scientists.
- Additional prisoners, virologists and guards around both floors of the Away Map, including a few more beanbag guards.

### Balancing:

- The random vial is the same as the one on the crashed pod exoplanet, meaning it will have one of the following reagents:  Reagent/Random, Rezadone, Three Eye.

### Screenshots:

StrongDMM-2025-04-05 21 55 30
StrongDMM-2025-04-05 21 55 50

### DIFF
--- a/maps/away/lar_maria/lar_maria-1.dmm
+++ b/maps/away/lar_maria/lar_maria-1.dmm
@@ -202,17 +202,18 @@
 	d2 = 8;
 	icon_state = "2-8"
 	},
-/turf/simulated/floor/tiled,
+/mob/living/simple_animal/hostile/lar_maria/guard/ranged,
+/turf/simulated/floor/tiled/monotile,
 /area/lar_maria/vir_hallway)
 "ay" = (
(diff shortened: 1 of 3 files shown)


Glloyd:
This will almost certainly need to be done again for the next baymerge, but no objections to it right now 👍

PR #1639, opened 2025-04-21
Syndicate Minibomb - Uplink Item

Laniakeaaa:
### Available for both Traitors & Mercenary.

Saw that we ported these from /tg/ at some point and wanted to re-add the classic back to the arsenal. These are a decent alternative to fragmentation grenades if you just want to explode someone, or cause a breach and explode someone, etc. They do a decent amount of damage to the area. Feedback on price, etc is always welcome.

- Traitor Cost:

28TC

- Mercenary Cost:

38TC

#### **Damage Potential:**
mini


Glloyd:
No objections here, this is the kind of thing that likely got lost in some previous merge

PR #1640, opened 2025-04-23
Kills known skill checks

ggermanicus:
do_skilled checks replaced with do_afters
flat skill checks removed, information gained is now free
probability-based skill checks arbitrarily replaced with a different probability


Glloyd:
absolute legend. thanks grant!

PR #1641, opened 2025-04-23
more merge fixes

ggermanicus:
-After deep soul searching I have decided that new players who are 30% likely to be griefers should not get insta-antagged on spawn. This fixes the multiple high/low selections for antag prefs
-Smartfridges work now. ~~I also removed other things that were for unit tests, just to see if it breaks the unit tests :)~~ The tests broke but the fridges work now
-detergent pods added to deck 2, false walls now have wall trim to hide better, research got their chemistry temperature thing back


ggermanicus:
wow! the unit tests broke!

PR #1643, opened 2025-04-24
Fixes and improves ship combat boarding, and adds fancy combat lighting (and other fixes)

Glloyd:
Fixes combat boarding just not working, and also adds new boarding methods. The old style is there once you knock the shields down, but there is also the potential for counterboarding. Hostile teleporters now have a per-component chance to leave a portal for early counterboarding. This plays into some other ideas I have for ship combat related things.

When entering combat, the nerva will also go into combat lighting mode in non critical areas. Lighting values are still in flux, so give me feedback on here or on discord.

Also fixes a bug with our template loaders not actually working because I repathed them incorrectly during the effectpocalypse.

Also fixes the pirate station being the wrong colour.


code/modules/urist/modules/shipbattles/ship_overmaps.dm
```diff
@@ -47,23 +47,43 @@
...
+				var/area/light_area = get_area(light)
+				if(!light_area.is_critical)
+					if(set_lighting)
+						if(LIGHTMODE_COMBAT in light.lightbulb.lighting_modes)
+							light.set_mode(LIGHTMODE_COMBAT)
+							light.update_power_channel(ENVIRON)
+					else
+						if(light.current_mode == LIGHTMODE_COMBAT)
+							light.set_mode(null)
+							light.update_power_channel(initial(light.power_channel))
```
scrdest: This is a potentially pretty huge operation, might be worth doing async.
```suggestion
/obj/overmap/visitable/ship/combat/proc/set_combat_lighting(set_lighting) //atmospheric combat lighting
        set waitfor = FALSE
        sleep(0)
        
	for(var/obj/machinery/light/light in SSmachines.machinery)
		if(light.z in map_z)
			if(light.lightbulb && light.current_mode != LIGHTMODE_EMERGENCY)
				var/area/light_area = get_area(light)
				if(!light_area.is_critical)
					if(set_lighting)
						if(LIGHTMODE_COMBAT in light.lightbulb.lighting_modes)
							light.set_mode(LIGHTMODE_COMBAT)
							light.update_power_channel(ENVIRON)
					else
						if(light.current_mode == LIGHTMODE_COMBAT)
							light.set_mode(null)
							light.update_power_channel(initial(light.power_channel))
```

Glloyd: good suggestion, thanks

PR #1645, opened 2025-04-24
Floor Lights are brighter again

Laniakeaaa:
Brighter Value looked nice + made example map less frustrating to showcase stuff.

### DIFF
--- a/code/game/machinery/floor_light.dm
+++ b/code/game/machinery/floor_light.dm
@@ -17,7 +17,7 @@ var/global/list/floor_light_cache = list()
 
 	var/damaged
 	var/default_light_power = 0.75
-	var/default_light_range = 5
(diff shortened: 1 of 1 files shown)


code/game/machinery/floor_light.dm
```diff
@@ -17,7 +17,7 @@ var/global/list/floor_light_cache = list()
 
 	var/damaged
 	var/default_light_power = 0.75
-	var/default_light_range = 3
+	var/default_light_range = 5
```
Glloyd: why 5 and not 4? Seems closer to the pre-merge value

Laniakeaaa: didnt check pre-merge version, was compairing on the example map when i was testing for other prs, i've lowered it tho

Issue #1335, opened 2023-10-14
Xenophage


Yantarii:
### Description of issue

- Queen can't pry doors.
- Drone's can't pry anything (Which is probably intended but should be communicated somewhat)
- Larvae don't have dark sight
- Xenos can't smash barricades
- Xenos can't destroy objects (tables, racks, etc, etc.)
- You can't see through weeds (they block sight)
- Would be nice if aliens could plant resin lights
- Weeds block light from emitting
- Alien Doors don't close behind you and have a bit of a slowdown.
- Membranes are super weak
- Xenos can pick up and wear clothing.
- Xeno Weeds seem to randomly retract, or go back and remove their own spread?
- Xenos suffer bleed out, and have to rest for quite a while to heal it.

godspeed anyone who touches this

### Difference between expected and actual behaviour

functional xenophages (one day)

### Steps to reproduce

1. Play Xenos on Colony.
2. Suffer.

### Specific information for locating

Xenophage code.

### Client version, server revision, & game ID

Current Merge (FOV Branch)
Part of Baymerge.

### Issue bingo

- [X] Issue could be reproduced at least once
- [X] Issue could be reproduced by different players
- [X] Issue could be reproduced in multiple rounds
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] [Couldn't find an existing issue about this](https://github.com/UristMcStation/UristMcStation/issues)

CPU-Blanc:
This is more of a feedback/feature request than a bug :P

Glloyd:
this is kind of a bug, xenophages shouldn't be able to wear clothes. xenophages also have some other wild issues:

xenophages can use mechs
xenophages can use energy swords
weeds block ai from targeting

CPU-Blanc:
Opps. Did this at 4am and didn't notice the bug nested in the middle. My bad

Yantarii:
Yeah, it's a little mix of the two really, It's probably good for us to just keep all the unholy shit of xenophage code in here

ggermanicus:
Additional April 2025 merge bugs/rewritten bug complaints:
- Xenos can pick up items, which causes problems in combat.
- Xenos are easily grabbed and wrestled into submission by humans
- Ability targeting is antiquated to the point of being unusable in any practical sense
- Odds of clicking yourself in combat is high due to sprite size
- Xeno thinkpans fit in MMI
[...]

PR #1646, opened 2025-04-27
Ghosts can no longer become xenos

ggermanicus:
You can still spawn them if you really want for an event.
Come back when someone maintains it


Glloyd:
One day I'll clean up xenos, but that day is not today. Can you add the new issues with Xenos to #1335?

PR #1650, opened 2025-04-29
Silenced Holdout Pistol & Knight-45 - Small Overhaul

Laniakeaaa:
## **Silenced Weapons adjustment:**
Samples Videos w/ audio provided below.
I might adjust the casings again, as they are quite loud, still. 

### **Holdout Pistol**

- Holdout Pistol changes firesound when surpressed, making it quieter.
- Holdout Pistol has a quieter unload and reloaded for stealths-sake.
- Holdout has a quieter casing drop sound (it's still quite loud)

Ideally, changing this to be more quiet when it's firing will make it actually viable to use it for assassination, instead of just a pistol with some jank.

### **Knight-45**

- Knight 45 has a quieter unload and reload sound.
- Knight 45 is now correctly silenced w/ silencer boolean.
- Updates the w_class to new standard. w_class = 2, to w_class = ITEM_SIZE_MEDIUM

### New Version:
[Changed.webm](https://github.com/user-attachments/assets/4258b3c9-5246-4b53-8fae-b9adc1006400)

### Original:
[Original.webm](https://github.com/user-attachments/assets/d62e9473-9b1f-4ee6-8b7b-8251807fb10b)


code/modules/projectiles/guns/projectile/pistol.dm
```diff
@@ -183,6 +183,9 @@
 	origin_tech = list(TECH_COMBAT = 2, TECH_MATERIAL = 2, TECH_ESOTERIC = 2)
 	magazine_type = /obj/item/ammo_magazine/pistol/small
 	allowed_magazines = /obj/item/ammo_magazine/pistol/small
+	auto_eject_sound = 'sound/urist/guns/casing_quiet.ogg'
+	mag_insert_sound = 'sound/urist/guns/pistol_magin_quiet.ogg'
+	mag_remove_sound = 'sound/urist/guns/pistol_magout_quiet.ogg'
```
scrdest, line 188: This is a non-Urist file, so I would advocate for wrapping this in an `ifdef INCLUDE_URIST_CODE` block (the macro is already supported).

This will make it easier to detect as our change on a merge and (if we applied this consistently) would eventually allow us to compile pure-Bay on Uristcode to check if an issue is theirs or ours.

Laniakeaaa: > This is a non-Urist file, so I would advocate for wrapping this in an `ifdef INCLUDE_URIST_CODE` block (the macro is already supported).
> 
> This will make it easier to detect as our change on a merge and (if we applied this consistently) would eventually allow us to compile pure-Bay on Uristcode to check if an issue is theirs or ours.

I'll take a look at doing this so we can avoid some issues in the future, gotcha

PR #1651, opened 2025-05-01
Refactors + Optimizes Planet Code for huge performance gains + footstep sounds for planet turfs + other fixes

Glloyd:
This saves fully 10-20 seconds on init times on Nerva, and probably has some gains on glloydstation too, although I haven't tested to see how significant it is there. 

I was wondering why init time was taking so long while testing something unrelated, and tracked down one of the major causes - the absolute unoptimized pile of dogshit that is planet code, parts of which haven't been touched in **11 years.** Some of that code is even older, from at least 2013. Suffice to say, it fucking sucks. It could probably use a full rewrite, but this still brings significant improvements with zero changes in functionality.

Main points:
Initialize() for planet turfs has been totally reworked - no longer do we have an insane chain of if(prob) checks, things are slightly more organized and sane, and we also don't run down that whole chain if we're not even generating anything. Now, that's handled by the generate_planet_things() proc, which handles most of what Initialize() used to handle, and returns a lateInitialize hint, allowing a ton of turfs to init way quicker than before because they're not checking a million things on init. Importantly, for the turfs that do spawn things, all the actual spawning of things has been moved to LateInitialize - all of the things to be spawned are added to a list on init, which is then spawned in during LateInitialize using create_objects_in_loc(), which is a much smoother way of handling spawning potentially multiple atoms at once. Possibly more importantly, instead of being dealt with in Initialize(), expensive icon ops are queued to be handled by the icon update subsystem on planet turfs, and on the jungle plants, which have also had some of their init stuff reworked.

This also purges a bunch of old, unused, or otherwise dogshit code that is not needed, and in doing so, actually adds proper footstep sounds for planet turfs (or readds? I think we had that at some point for the jungle at least but who the fuck knows with this mess)

Death to me 11 years ago

This PR also fixes a couple unrelated issues that popped up as I was testing this:
Skyboxes were just broken - bad merge.
Pirate station and voxship appearances were slightly obvious.
I also removed medium pirate ships from spawning pending some changes to spawning code.

### DIFF
--- a/code/modules/urist/modules/jungle/planet_plants.dm
+++ b/code/modules/urist/modules/jungle/planet_plants.dm
@@ -162,10 +162,6 @@
 		to_chat(user, "<span class='notice'> You pick a fruit off [src].</span>")
 
 		var/obj/item/reagent_containers/food/snacks/grown/jungle_fruit/J = new (src.loc)
-//		J.potency = plant_strength
-//		J.icon_state = fruit_icon_states[fruit_type]
-//		J.reagents.add_reagent(reagent_effects[fruit_type], 1+round((plant_strength / 20), 1))
-//		J.bitesize = 1+round(J.reagents.total_volume / 2, 1)
 		J.attack_hand(user)
 
 		CutOverlays(fruit_overlay)
--- a/code/modules/urist/modules/jungle/planet_plants.dm
+++ b/code/modules/urist/modules/jungle/planet_plants.dm
@@ -14,19 +14,22 @@
 	var/stump = 0
 	atom_flags = ATOM_FLAG_CLIMBABLE
 
-/obj/structure/bush/New()
-
-	icon_state = "bushnew[rand(1,4)]"
-
-	if(prob(20))
-		name = "thick foliage"
-		opacity = 1
-		desc = "Very thick scrub that blocks your vision. It'll take something sharp and a lot of determination to clear away"
-		icon_state = "thickbush[rand(1,2)]"
+/obj/structure/bush/Initialize()
+	. = ..()
 
 	if(indestructable)
 		icon_state = "thickbush[rand(1,2)]"
-	..()
+
+	else
+		if(prob(20))
+			name = "thick foliage"
+			opacity = 1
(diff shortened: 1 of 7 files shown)


Glloyd:
at war with the labels bot

code/modules/urist/modules/jungle/planet_plants.dm
```diff
@@ -201,16 +196,9 @@ var/global/jungle_plants_init = 0
 	icon_state = "reedbush_1"
 	anchored = TRUE
 
-/obj/structure/flora/reeds/New()
```
Glloyd, line 204: what the fuck even was this bullshit

code/modules/urist/modules/jungle/planet_turfs.dm
```diff
@@ -11,96 +11,117 @@
...
+			if(prob(large_trees_chance))
+				spawn_list.Add(large_tree_type)
+		if(trap_spawn_chance)
+			if(prob(0.1))
+				spawn_list.Add(/obj/structure/pit/punji6/hidden/dull)
+
+	if(length(animal_spawn_list))
+		if(prob(animal_spawn_chance)) //even 1% is a fuck ton. A 100x100 area is 10000 tiles. This means that at 1% chance, 100 animals will spawn in that area.
+			var/A = pick(animal_spawn_list) //A conservative estimate for possible spawning tiles in the jungle would be 22500, which is very conservative. This means that roughly 225 animals would spawn.
+			spawn_list.Add(A) //so set your numbers fucking low. god i hate this shit.
```
scrdest, line 118: Rather than statically spawning them per-tile, might be nicer to use the uspawner objects I've added a while back.

You can set the spawn radius on them or set them to target the whole area, and you can set the max spawn limit and adjust spawn tickrates if desired. If using endless, this would emulate animals migrating in.

There's currently no tracking for how many of these are living so that we could spawn based on active mobs, but some Observer magic would almost certainly easily handle that now that I know it exists.

Glloyd: That's a great idea, but something that I would probably address in a future PR. This is moreso just a hotfix to address the worst of the init slowdown.

PR #1414, opened 2024-03-13
feature: GOAI Urist port v2

scrdest:
Yep. This again. And a cool 10k+ SLOCs to go through!

Featuring an entirely re-done architecture - the heart of the AI is now a Utility AI, blatantly lifting the core design from Infinite Axis Utility System by Dave Mark & Co., with improved and optimized GOAP Planner AI as an optional sidecar for dealing with complex environmental problem-solving (e.g. GODDAMN DOORS REEE).

### Fast and smart

A major consequence of that is that this iteration of GOAI is at least as smart as the previous (with a strong argument that it's much more so), but also _very, very **fast**_. And that's _before_ AI LODs or any other future optimizations.

### Data-driven

The redesign also showcases my progressing conversion to the church of Data-Driven Design. As a result, the AI system is really two things: an **AI Engine** in DM code (because I am a masochist) and an **AI Specification Language** in the form of a bunch of JSON schemas.

The AI Engine provides a pile of abstractions and building blocks (curves, Considerations, ContextFetchers...) that can be used in the JSONs to build actual AIs and SmartObjects. 

Thanks to that, the AI Specs can be edited without an in-depth knowledge of DM's weird quirks. You don't have to be a full-on coder to make GOAI AIs.

They can also be hot-swapped at runtime without recompiling the whole game as long as you're not extending the AI Engine itself. Should make iteration speed muuuuuuuch faster.

### Modular and extentable

Unlike _Certain Other SS13 AI Implementations_, GOAI provides only a cheap, bare minimum shared framework for any gameobject that opts into having GOAI enabled at all.

Everything else - all behaviors and checks - only runs if an AI Agent uses them for something and they run in whatever sequences they are acquired. The only hardcoded state-machines are either the AI Brain's core decision/execution loop or the scripts powering the actions (if you choose to write them that way). The AI will transition fluidly between combat, idling, chores, and everything in between based on whatever's the next logical priority.

Actions can be added or removed on the fly - in fact, they already are. 

To avoid duplicating logic across 20 different mob types, GOAI now includes Sims-style SmartObjects, which are AI Specs attached to gameobjects. Agents can harvest new abilities from those objects - for example, a held gun may provide a Shoot Action, and a nearby Door may provide an OpenDoor Action.

Even the GOAP integration is more flexible than traditional implementations - GOAI plan execution is smart enough to handle third-party interference cancelling out the effects of actions earlier in the plan or short-circuiting the requirements to make the ultimate goal available.

PR #1652, opened 2025-05-01
updates a bunch of Urist code to 515 and 516 standards

Glloyd:
This is mostly just changing proc references to TYPE_PROC_REF, GLOBAL_PROC_REF, and PROC_REF

This also now includes updates for some Urist UIs to be more functional with 516.

### DIFF
--- a/code/modules/mob/floating_message.dm
+++ b/code/modules/mob/floating_message.dm
@@ -38,7 +38,7 @@ var/global/list/floating_chat_colors = list()
 	if(small)
 		fontsize = 6
 
-	if(copytext_char(message, length_char(message) - 1) == "!!")
+	if(copytext_char(message, length_char(message) - 1) in list("!!","?!"))
 		fontsize = 8
 		limit = 60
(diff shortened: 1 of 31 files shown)


Glloyd:
Some of this will need to be applied to #1414, namely the proc ref changes.

code/modules/mob/floating_message.dm
```diff
@@ -38,7 +38,7 @@ var/global/list/floating_chat_colors = list()
 	if(small)
 		fontsize = 6
 
-	if(copytext_char(message, length_char(message) - 1) == "!!")
+	if(copytext_char(message, length_char(message) - 1) in list("!!","?!"))
```
scrdest, line 41: Might be worth extracting this to a global thingy, otherwise this does an alloc every time it procs.

Glloyd: yeah, if you want to take a stab at that. I was just restoring our pre-merge implementation here.

PR #1655, opened 2025-05-16
Maymerge

Glloyd:
Baymerge to May 16th, 2025.

Of note are fixes for some 516 client issues regarding goonchat.
There are also some other minor features, including space carp actually migrating around on the overmap.
Some other bug fixes too, including some I already did here and pushed upstream, and my fix for the sryinge preview icons @Laniakeaaa


Glloyd:
this will conflict with #1654 but i'll cross that bridge when i get there.

PR #1669, opened 2025-08-05
Updates Random Loot Spawners - (Contraband & Ammunition)

Laniakeaaa:
Updates multiple loot tables of random spawners.

**Contraband Loot Spawners can also spawn the following:**

Fake Moustache - (1%)
Ancient Toolbox - (1%)
Rig Module - Energy Net - (1%)
Jet Fuel Vial - (2%)
Disguised Vape Syringe Gun - (1%)
Pistol Silencer - (2%)

**Ammunition Spawners can also spawn the following:**

Singular Haywire Shell - (1%)
9mm Holdout Speedloader - (2%)
Chem Dart - (1%)
9mm Broomstick Pistol Clip - (2%)
7.62mm Stripper Clip - (1%)
Hunting Rifle Stripper Clip - (1%)
10mm Speedloader - (2%)

Gives more contraband items that are semi harmless that crew can discover, or traitors can save TC by finding. 
Antagonists will benefit from potentially finding ammo they need by scouring maintenance at the start to save TC. Since crew cannot access most of these weapons the ammo won't be as useful to them.

### DIFF
--- a/code/game/objects/random/random.dm
+++ b/code/game/objects/random/random.dm
@@ -269,7 +269,8 @@
 				/obj/item/storage/toolbox/ancienttoolbox = 1,
 				/obj/item/rig_module/fabricator/energy_net = 1,
 				/obj/item/storage/box/syndie_kit/jetfuel = 2,
-				/obj/item/gun/launcher/syringe/disguised = 1 // rare enough
+				/obj/item/gun/launcher/syringe/disguised = 1, // rare enough
+				/obj/item/silencer = 2
 				)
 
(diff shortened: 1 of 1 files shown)


ErrorsWindows:
A part of me would recommend boxes of cartridges rather than loaded magazines akin to how /vg/station does it, where you can print boxes of revolver ammunition from the autolathe/ammolathe but you have to get the speedloaders from the uplink, but I don't think it would really count here as you can just get all the magazines you want from an autolathe anyways

PR #1670, opened 2025-08-06
fix byond caching/install

CPU-Blanc:
Downloading Byond when there is no cache is causing 429 - Too Many Requests, as each test tries to download the installer simultaneously. Adds a new Job to download and cache Byond first, so that subsequent jobs can just use the cache instead and only result in a single download

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,7 +54,26 @@ jobs:
           wget https://raw.githubusercontent.com/DiscordHooks/github-actions-discord-webhook/master/send.sh
           chmod +x send.sh
           ./send.sh failure $WEBHOOK_URL
+
+  InstallByond:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+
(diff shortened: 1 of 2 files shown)


scrdest (approved):

scrdest:
CI failing currently

CPU-Blanc:
Yeah, we're still getting 429/403'd. Might need to wait a little for the runner's IP to be unbanned

CPU-Blanc:
I have a feeling we're getting a CF javascript challenge that the runner is failing..

We could do a jank workaround if we're happy to, and re-work this into a scheduled job with a workflow dispatch, then manually upload the zip (or mirror the binary to a different site), cache it, then delete it again. That way the scheduled job will keep the cache alive if we run it say.. every week?
Would only work until the version changes, but there's not much else we can do, as it seems they've locked the site down a bit more, most likely due to the DDoS attacks

CPU-Blanc:
Setting the user-agent seems to have done the trick on resolving the 403's. If you're happy with these changes @scrdest , we can merge this and rebase outstanding PRs to get testing running again on those

scrdest:
@CPU-Blanc - sounds good to me! :shipit:

PR #1684, opened 2025-08-17
[WIP] Reworks Jungle Planetoid Plants

Laniakeaaa:
### Features:

- Updates Jungle Plants to be similar to modern Baycode, makes them a bit less janky.
- WIP atm, I'll update progress later.


code/modules/urist/modules/jungle/planet_plants.dm
```diff
@@ -2,7 +2,24 @@
...
+//	density = TRUE
+	anchored = TRUE
+	atom_flags = ATOM_FLAG_CLIMBABLE
+
+
+/obj/structure/flora/bush/Initialize()
+	. = ..()
+	var/state = rand(3)
+	if (state)
+		icon_state = "[initial(icon_state)]_[state]"
```
scrdest, line 19: rand(3) can return 0, 1, 2 or 3, but the zero case would not trip this if-clause. Is this intentional? I cannot look at the icon file ATM (where were you when scr's mobo was kill?)

Laniakeaaa:
probably worth redoing after https://github.com/UristMcStation/UristMcStation/pull/1732 goes through. A lot of the jungle planet stuff needs updating to modern bay

PR #1702, opened 2025-09-16
Urist McStation: MOTD Update

Laniakeaaa:
- Rearranges a bunch of the current text, moves links higher up, , fixes bay13 to bay12, etc.
- Adds our unique differences to normal baycode
- Adds information about appealing bans. (mainly job)
- Adds some new player information about neural laces & respawning, as it's different to normal Baycode servers. New player experience is important 2 me

feedback on text always welcome
it's a little longer, but using  [center] looks nice but adds a line and shit, should be alright

### New MOTD:
image

### Old MOTD:
image

### DIFF
--- a/config/motd.txt
+++ b/config/motd.txt
@@ -1,15 +1,19 @@
-<h1>Welcome to Urist McStation! Before you begin...</h1>
+<center><h1>Welcome to Urist McStation!</h1></center>
 
-<i>This server is running a BayStation13 Git build with a custom map and features.</i>
-<i>This is a medium roleplay server.  Roleplay is encouraged, and please don't be egregiously out of character.</i>
-<h2>Server Rules:</h2>
-Click the "Rules" tab at the top of this window and read them!
-If you aren't aware of our rules, you could be warned by an admin or jobbanned! Poor conduct as a member of the security force/a head/a silicon will not be tolerated. These roles will be treated more harshly should rules be broken.<br>
-Keep IC in IC chat and OOC in OOC chat.  Please don't discuss ongoing rounds in OOC or you will be muted.
-No killing, except in self defense, or as an antagonist (and show restraint even then).  Sitting around watching isn't fun.<br>
-If you have a problem or a question, <b>adminhelp it!</b> You can do so at any time by pressing F1.<br>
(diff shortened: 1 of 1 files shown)


config/motd.txt
```diff
@@ -1,15 +1,19 @@
...
-If you have a problem or a question, <b>adminhelp it!</b> You can do so at any time by pressing F1.<br>
-We exist for a reason, and would like to keep the server fun. We can't predict everything, and questions are likely to crop up during play, especially with regard to antagonists. Asking won't hurt anyone.<br>
-<b>We recently updated our ruleset, click the rules tab at the top of your screen to read through them. In the case of an incident, it will be assumed that you have already read them.</b><br>
-<b>Golden Rule:</b> Do unto others as you would have done unto you.  Try to keep things enjoyable for everyone.
+Check our server rules by clicking on the <b>Rules</b> Tab in the top right, it will be assumed that you have already read them in the case of an incident. If you aren't aware of our rules, you could be warned by an admin or jobbanned! Game & Job Bans can be appealed on our <a href="https://discord.gg/0oRsdvoS1DDd5Rv0">Discord</a> in the <i>#ban-appeals</i> channel. <br>
+Poor conduct as Command Staff, Security or Silicons roles will not be tolerated, and will be treated more harshly should rules be broken.
```
ggermanicus: I think in this case 'silicon' is already plural.

config/motd.txt
```diff
@@ -1,15 +1,19 @@
...
+Do not discuss on-going events in OOC chat, or you will be muted.<br>
+If you are not an antagonist, do not kill unless in self defense within sufficent escalation.
+As an antagonist, avoid murderboning the crew (unless they instigate) as it isn't fun to sit and watch!<br>
```
ggermanicus: An actual admin should weigh in, but I personally don't think the wording is strong enough here. If I read this instead of the rules page I would interpret this more as 'murderboning is impolite' rather than 'murderboning is going to get me banned'. 'Unless they instigate' might also be unnecessary.
Per rules: 'Murderboning is not allowed, not even as an antag.'

Laniakeaaa: I'll adjust it to be more stern, looking back this definitely isn't harsh enough haha

Laniakeaaa:
changes made

ggermanicus (approved):

PR #1704, opened 2025-09-17
Setup nerva_config, cleans nerva_define, adds nerva_lobby.dm, Nerva overmap details added

Laniakeaaa:
- Setup nerva_config in /configs/, allows for quick changes to the map without code-diving, etc such as economy, fluff, overmap + gen
- Adds nerva_lobby.dm - Controls our lobby tracks & lobby images, we now swap between rainbow skybox and non-rainbow Nerva, and has support for more lobby images.
- Job Blacklists moved to nerva_jobs.dm, same as baycode struct
- Adds usable emails - freemail, nerva.ics, nerva.nt
- Adds NT R&D Sector and Pirate Listening Post to admin faxes, for fun aboose
- Removes bluespace interlude. Haven't ever seen it fire, especially with low pop, disabled but easily re-enabled if we want to.
- moved base_floor_type to nerva_areas, similar to how it's setup in Bay Code.
- logo set to ["\[nervalogo]\] instead of the raw png for paperwork stuff.

ICS Nerva has a description that can be scanned by other vessels on the Overmap now:
_"A independently commissioned salvage freighter class ship, broadcasting the designation \"ICS Nerva\"."_

Trajan has a description that can be scanned by other vessels on the Overmap now:
_"A TL-432 long range exploration shuttle, broading the callsign \"Nerva-1 Trajan\""_

Hadrian has a description that can be scanned by other vessels on the Overmap now:
_"A Nanotrasen NT-SV research shuttle, broadcasting the callsign \"Nerva-3 NT-Hadrian\""_


config/nerva_config.txt
```diff
@@ -0,0 +1,54 @@
...
+## Messages to use for different round-end conditions. %ETA% is replaced with a time.
+#SHUTTLE_DOCKED_MESSAGE Attention all hands: Jump preparation complete. The bluespace drive is now spooling up, secure all stations for departure. Time to jump: approximately %ETD%.
+#SHUTTLE_LEAVING_DOCK Attention all hands: Jump initiated, exiting bluespace in %ETA%.
+#SHUTTLE_CALLED_MESSAGE Attention all hands: Jump sequence initiated. Transit procedures are now in effect. Jump in %ETA%.
+#SHUTTLE_RECALL_MESSAGE Attention all hands: Jump sequence aborted, return to normal operating conditions.
+#EMERGENCY_SHUTTLE_DOCKED_MESSAGE
+#EMERGENCY_SHUTTLE_LEAVING_DOCK
+#EMERGENCY_SHUTTLE_RECALL_MESSAGE
+
+## Money related config options.
```
ggermanicus, line 46: This file seems to work, but literally everything in it is commented out. This has resulted in the Nerva's starting money being the default 75000, since the starting money here is commented out and you removed the value in nerva_define.dm. If a value here is still defined in nerva_define.dm, the value in nerva_define.dm will override the value here.

You should at minimum uncomment L47-54, and if you intended to uncomment more, make sure it isn't going to be overridden by values in other files.

maps/nerva/nerva_define.dm
```diff
@@ -39,43 +26,34 @@
...
 	away_site_budget = 5
 	min_offmap_players = 0
 
-	date_offset = 560
+	id_hud_icons = 'maps/nerva/icons/assignment_hud.dmi'
+	welcome_sound = 'sound/AI/welcome.ogg'
 
-	base_floor_type = /turf/simulated/floor/reinforced/airless
-	// for base_floor_area, look in nerva_define.dm
+	use_bluespace_interlude = FALSE // nah.
```
ggermanicus, line 54: re: bluespace interlude
You've never seen it fire, but it does work. As-is, it's a 'peril of the warp' sort of deal for those who teleport to something 5+ overmap tiles away... which is practically never on the Nerva.

I personally think there's no harm in it staying in, and although it's a bit Torch-centric at the moment, it might be interesting for future development. If you really want it gone, that might fit better in a separate PR.

ggermanicus (changes requested):

PR #1712, opened 2025-09-25
Ports Client Screen Recoil from Upstream.

Laniakeaaa:
It's an open PR but it seems more or less done, this is just a port of the screen recoil system.
screen_shake did not look nice, so I'm glad we have a new system now.

This has dependancies on another PR, which I'll PR soon which will have individually set the values for guns. Currently all guns use the default parent value.

https://github.com/user-attachments/assets/5dc94b94-b4cf-48a6-8951-6d550f969b1e

This is the basic default value per the Upstream PR, we can adjust the strength, duration and easing to make this a lot worst for big powerful guns.

### DIFF
--- a/code/_helpers/maths.dm
+++ b/code/_helpers/maths.dm
@@ -62,6 +62,11 @@
 /// The cotangent of degrees
 #define Cot(degrees) (1 / tan(degrees))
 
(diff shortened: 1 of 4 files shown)


ggermanicus:
Gonna reserve judgement until I see the manual values on guns. Screenshake is a bit jarring but the default value seems really gentle, shooting will definitely be easier currently, for good or ill

PR #1714, opened 2025-10-12
Welding Tank Packs are less OSHA compliant

Laniakeaaa:
- Welding Tool Backpacks do significantly more damage to the nearby surroundings
- These don't spawn on the Nerva anyways, and the current explosion they make is pitiful at best
- Makes them peak industrial accident.

**Old:**
image

**New:**
image

### DIFF
--- a/code/game/objects/items/weapons/weldbackpack.dm
+++ b/code/game/objects/items/weapons/weldbackpack.dm
@@ -33,7 +33,7 @@
 					SPAN_DANGER("You hold \the [T] up to \the [src], causing an explosion!")
 				)
 				log_and_message_admins("triggered a fueltank explosion.", user)
-				explosion(get_turf(src), 9, EX_ACT_HEAVY)
+				explosion(get_turf(src), 7, EX_ACT_HEAVY)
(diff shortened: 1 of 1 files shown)


ErrorsWindows:
i do think some of them spawn in the welding supplies locker in engineering (or at least i vaguely remember them being in there), but when's the last time *you* took one from that locker as an engineer? or did you just grab the big welder with the big red canister on the bottom like a normal person
saying that i'm a bit apprehensive at the hull breach potential but i'm also not sure how you'd get the authentic "welding tank fire" effect like with a proper welding tank explosion so it'll probably be fine

Laniakeaaa:
I think giving them the fuel leak ability would be quite powerful for what they are, since you'd get a very portable superheated room melter that isn't as visible as pulling a actual welding tank. The welding packs would definitely be lethal on the bottom floor for breaches but I think it's already in the 1/1000 someone would actually want to blow themselves up with this, where as fuel tanks tend to explode a fair bit here and cause a lot of damage

ggermanicus (approved):

PR #1716, opened 2025-10-19
Adds Firearm Attachment Uplink Catergory, URIST_CODE new name for silencers, etc.

Laniakeaaa:
- Adds a Firearm Attachment catergory to the Traitor Uplink, currently just housing the Universal Suppressor, but I'd like to add more items eventually to this, especially in non-universal attachments for specific gear.

- Sets up a URIST_CODE for silencers, making them be renamed and changing the desc from silencers to suppressors.

PR #1700, opened 2025-09-15
Silenced Holdout Pistol & Knight-45 - Small Overhaul

Laniakeaaa:
## **Silenced Weapons adjustment:**
- Reopens #1650 

Samples Videos w/ audio provided below.
I might adjust the casings again, as they are quite loud, still. 

**TODO:**

- [ ] ifdef IncludeUristCode implementation - (scrdest)

### **Holdout Pistol**

- Holdout Pistol changes firesound when surpressed, making it quieter.
- Holdout Pistol has a quieter unload and reloaded for stealths-sake.
- Holdout has a quieter casing drop sound (it's still quite loud)

Ideally, changing this to be more quiet when it's firing will make it actually viable to use it for assassination, instead of just a pistol with some jank.

### **Knight-45**

- Knight 45 has a quieter unload and reload sound.
- Knight 45 is now correctly silenced w/ silencer boolean.
- Updates the w_class to new standard. w_class = 2, to w_class = ITEM_SIZE_MEDIUM

### New Version:
[Changed.webm](https://github.com/user-attachments/assets/4258b3c9-5246-4b53-8fae-b9adc1006400)

### Original:
[Original.webm](https://github.com/user-attachments/assets/d62e9473-9b1f-4ee6-8b7b-8251807fb10b)

PR #1717, opened 2025-10-19
Ports Bay's Supressor PR

Laniakeaaa:
**Original Baystation PR Features:**            CL Credit:  Karl Johansson & l3lb0t
- Barrels can now be threaded, allowing supressors to be attached to any specified weapon with a threaded barrel.
- Supressors can be offset by a specific amount for sprite's sake.
- Clicking on the pistol no longer takes the supressor off, instead of taking the magazine out.
- Adds a right click verb for removing the suppressor.
- Ports new Holdout Pistol Sprites
image

**Urist-Specific Changes:**
- This also ports the original supressor sounds, and keeps our version as a urist def over it.
- Also fixes a bug in the current PR pending where the empty version of the pistol does not show up.
- Currently the only weapon that can be supressed is the holdout pistol, but this will change.

**TODO:**

- [ ]  Remove caliber restriction, so all weapons fit a suppressor unless someone really wants to sprite like 10 unique ones.


Laniakeaaa:
Adds:

https://github.com/UristMcStation/UristMcStation/pull/1700
https://github.com/UristMcStation/UristMcStation/pull/1716

code/modules/projectiles/guns/projectile/pistol.dm
```diff
@@ -171,6 +171,10 @@
 
 /obj/item/gun/projectile/pistol/holdout
 	name = "holdout pistol"
+	#ifdef INCLUDE_URIST_CODE
+	desc = "The Lumoco Arms P3 Whisper. A small conceable pistol with a threaded barrel. Uses 9mm rounds."
```
ggermanicus, line 175: Typo conceable -> concealable

code/modules/projectiles/guns/projectile/pistol.dm
```diff
@@ -183,71 +187,44 @@
...
 
 /obj/item/gun/projectile/pistol/flash
 	name = "holdout signal pistol"
 	magazine_type = /obj/item/ammo_magazine/pistol/flash
 
 /obj/item/silencer
+
+	#ifdef INCLUDE_URIST_CODE
+	name = "universal suppressor"
+	desc = "A universal suppressor that can be attached to any weapon with a threaded barrel."
```
ggermanicus, line 213: I think until more than one gun can use this suppressor it's better to have this desc clearly say it only works with the holdout pistol.

code/datums/uplink/attachments.dm
```diff
@@ -0,0 +1,10 @@
+// Firearm and Melee Weaponry Attachments - Urist
+
+/datum/uplink_item/item/attachment
+	category = /datum/uplink_category/attachment
+
+/datum/uplink_item/item/attachment/universal_suppressor
+	name = "Universal Suppressor - Attachment"
+	desc = "A universal suppressor, designed to attach to the threaded barrel of a firearm to give a significant reduction in sound profile."
```
ggermanicus, line 8: Also suggesting to make it clear it only works on the holdout pistol for now. Given that you can buy the suppressor and its pistol for 20tc together, this might also be unnecessary to have at the moment.

ggermanicus (changes requested):

PR #1671, opened 2025-08-06
Reworks Jungles, Arid, Snow Trees & Adds Cacti & Old Tree

Laniakeaaa:
trees

### Features:

Reworks all Planet Trees:

- Planet/Jungle  & Variants
- Planet/Arid  & Variants
- Plantet/Temperate & Variants

- Adds Cacti that were previously unused.
- Adds Oldtree from Jungle Sprites.

- Trees now are updated to be at the same feature level as current Bay Trees, such as Pine Trees, etc. You should be able to cut them down with various objects again, and currently will recieve normal wooden planks.

- Moves multiple flora files into the proper obj location for them to have similar parity to Bay Sprite folder struct. 

- Changes Forest2 Tree to Bay's Pine Tree version.

- Moves Asteroid Flora and updates .dmi naming.

PR #1732, opened 2025-10-26
Rework: Simplifies urist trees and carpentry system

### LINKED ISSUE #1661: Cutting down Trees no longer generates logs.
### Description of issue

Trees that are cut with the Axe/Hatchet do not drop any materials such as logs on the ground, sometimes you will get a single wooden plank with a wooden axe. This prevents people from making rafts on the jungle planet without bringing wood with them. 

Also the attack cooldowns may be broken for tree cutting as you can speed cut them down in seconds.

Trees should be cut down in logs, which require a saw to process, and / or wood processor.

Acutal Behaviour: Trees disappear, you can speed cut in a few seconds.

### Steps to reproduce

1. Find a jungle tree.
2. Use a axe/hatchet.
3. Click as fast as you can.
4. No logs or wood.

Jungle Planetoid is likely culprit, probably some material changes on Bay's behalf, we should probably update it at some point to use the Bay exo gen.

Client Version: 516
Server Revision: [cdb81763c38742f9478625e2bac50125f04f6007](https://github.com/UristMcStation/UristMcStation/commit/cdb81763c38742f9478625e2bac50125f04f6007) - 478625e2bac50125f04f6007 - 2025-05-17
Game ID: 8070ada9
Current map: ICS Nerva

### Issue bingo

- [ ] Issue happened in a recent (less than 7 days ago) round
- [ ] [Couldn't find an existing issue about this](https://github.com/UristMcStation/UristMcStation/issues)

ggermanicus:
- Urist trees, like Bay trees, magically turn into planks when you chop them down.
- Trees that fall in the forest do make a sound.
- The saw is gone. Hitting logs with a hatchet/axe/similar will delicately form a log into planks.
- The carpentry planer and unrefined wood are also gone. planer.dm renamed to woodprocessor.dm.
- 'Initial' versions of urist tree paths have become small trees, to prevent devs spawning invisible trees on accident.
- /planet/ in tree paths removed, as it serves no purpose now and makes spawning clunky. No real trees were removed.
- Tree sprite selection mimics Bay's to prevent tree stumps becoming errors and sprites never being selected
- Adds hint to initial raft step to add more planks
- Modernizes some spans
- Attacking trees now accounts for your attack speed. Fixes https://github.com/UristMcStation/UristMcStation/issues/1661

The axe>logs>saw>planer system is a lot of steps for some planks. IMO planks aren't so awesome as to merit a 4-step process.


ggermanicus:
now implements restructuring and oldtree + cactus from https://github.com/UristMcStation/UristMcStation/pull/1671

